PR #16, opened 2013-02-20
Enable Geolocation in InAppBrowser

ts1:
This allows to run Google Maps in InAppBrowser. (my project needs it)
Actually I just copied a method from CordovaChromeClient.java.


odanquah:
Am trying to use this in my solution but it is not working. How do i use the InAppBrowser.java? If add what you added to my InAppBrowser, i can not compile my project. Can you help?

PR #116, opened 2014-08-22
[CB-7371][Android] Fixed: app crash after phone language change

martincgg:
The main reason of this behavior is the config.xml, origin="*" tag, specifies access to all kind of resources on the network and makes the whiteList object on the WhiteList class a null object, indicating that when origin="", there's no whiteList, all is allowed.

When it tries to resume the application it checks:
if (url.startsWith("file://") || url.startsWith("javascript:") || whitelist.isUrlWhiteListed(url))
This condition passes and triggers the load for 'about:blank'
The condition passes because, whitelist.isUrlWhiteListed(url) statement returns true.
When the origin="*" is processed by the WhiteList class, it sets a null to the whiteList object, which it should contain the access to resources allowed, however when whitelist.isUrlWhiteListed(url) it returns a true if the whiteList object is null, therefore the condition passes and continues to the loadurl method on webClient, stopping the application, because of the 'about:blank' url.

Adding another condition along with that statement for when whiteList = null, but url= about:blank, in that way it won't be able to load the url.

### DIFF
--- a/framework/src/org/apache/cordova/CordovaWebView.java
+++ b/framework/src/org/apache/cordova/CordovaWebView.java
@@ -427,7 +427,7 @@ public class CordovaWebView extends WebView {
         if (LOG.isLoggable(LOG.DEBUG) && !url.startsWith("javascript:")) {
             LOG.d(TAG, ">>> loadUrlNow()");
         }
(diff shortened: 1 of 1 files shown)


Cadey:
Hi guys, I just grabbed cordova 5.0.0 and the issues of the app crashing is still present - unfortunately im not in a position to put any debugging info at the moment but wanted to let you know its still happening with some configs.

PR #143, opened 2015-01-05
CB-8201 Add support for auth dialogs into Cordova Android

sgrebnov:
https://issues.apache.org/jira/browse/CB-8201

Makes it possible to provide custom pluggabble logic to process auth challenge (for example by installing corresponding plugin, sample implementation: https://github.com/msopentech/cordova-plugin-auth-dialog).

*Motivation*
If you load some resource which requires authentication using android default browser default login dialog appears, but if you try to load the same resource using cordova app then no dialog is shown for user and request is failed w/ 401 Unauthorized. So Currently apps have to provide own JS logic to show web based auth dialog. 

It will be great if Cordova Android offers special functionality (in core or via special interface plugin developers can extend) to automatically show native auth dialog. This is extremely useful in corpnet scenarios where all resources requires user to be authenticated.

### DIFF
--- a/framework/src/org/apache/cordova/CordovaWebViewClient.java
+++ b/framework/src/org/apache/cordova/CordovaWebViewClient.java
@@ -115,15 +115,22 @@ public class CordovaWebViewClient extends WebViewClient {
     @Override
     public void onReceivedHttpAuthRequest(WebView view, HttpAuthHandler handler, String host, String realm) {
 
-        // Get the authentication token
+        // Get the authentication token (if specified)
         AuthenticationToken token = this.getAuthenticationToken(host, realm);
         if (token != null) {
             handler.proceed(token.getUserName(), token.getPassword());
+            return;
         }
-        else {
-            // Handle 401 like we'd normally do!
-            super.onReceivedHttpAuthRequest(view, handler, host, realm);
+
+        // Check if there is some plugin which can resolve this auth challenge
+        PluginManager pluginManager = this.appView.pluginManager;
+        if (pluginManager != null && pluginManager.onReceivedHttpAuthRequest(this.appView, new CordovaHttpAuthHandler(handler), host, realm)) {
+            this.appView.loadUrlTimeout++;
+            return;
         }
+
+        // By default handle 401 like we'd normally do!
+        super.onReceivedHttpAuthRequest(view, handler, host, realm);
     }
 
     /**
--- a/framework/src/org/apache/cordova/PluginManager.java
+++ b/framework/src/org/apache/cordova/PluginManager.java
@@ -243,6 +243,27 @@ public class PluginManager {
     }
 
     /**
+     * Called when the system received an HTTP authentication request. Plugins can use
+     * the supplied HttpAuthHandler to process this auth challenge.
+     *
+     * @param view              The WebView that is initiating the callback
+     * @param handler           The HttpAuthHandler used to set the WebView's response
+     * @param host              The host requiring authentication
+     * @param realm             The realm for which authentication is required
+     * 
+     * @return                  Returns True if there is a plugin which will resolve this auth challenge, otherwise False
+     * 
+     */
+    public boolean onReceivedHttpAuthRequest(CordovaWebView view, ICordovaHttpAuthHandler handler, String host, String realm) {
+        for (CordovaPlugin plugin : this.pluginMap.values()) {
+            if (plugin != null && plugin.onReceivedHttpAuthRequest(view, handler, host, realm)) {
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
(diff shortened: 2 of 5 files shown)


framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -147,6 +147,17 @@ public void setAuthenticationToken(AuthenticationToken authenticationToken, Stri
     }
 
     /**
+     * Specifies the HTTP auth handler for handling auth challenge requests.
+     *
+     * @param handler The HTTP auth handler.
+     */
+    public void setHttpAuthRequestHandler(HttpAuthRequestHandler handler) {
```
agrieve: If there's no direct reason to add this to CordovaActivity, I'd prefer it not added. CordovaActivity has a lot of symbols in it already, and they a little bit just add noise.

framework/src/org/apache/cordova/CordovaWebViewClient.java
```diff
@@ -89,6 +97,15 @@ public void setWebView(CordovaWebView view) {
         this.appView = view;
         helper = new CordovaUriHelper(cordova, view);
     }
+    
+    /**
+     * Specifies the HTTP auth handler for handling auth challenge requests.
+     *
+     * @param handler HTTP auth handler.
+     */
+    public void setHttpAuthRequestHandler(HttpAuthRequestHandler handler) {
```
agrieve: Why use a setter here instead of following the plugin pattern of allowing all plugins to handle the callback?

framework/src/org/apache/cordova/HttpAuthRequestHandler.java
```diff
@@ -0,0 +1,30 @@
...
+       software distributed under the License is distributed on an
+       "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+       KIND, either express or implied.  See the License for the
+       specific language governing permissions and limitations
+       under the License.
+*/
+package org.apache.cordova;
+
+import android.webkit.HttpAuthHandler;
+import android.webkit.WebView;
```
agrieve: For android-4.0.x, we're trying to do away with referring to android.webkit.WebView explicitly. It'd be better to pass in the CordovaWebView, and to create an abstraction around HttpAuthHandler that just wraps the webkit one.

framework/src/org/apache/cordova/CordovaWebViewClient.java
sgrebnov: Andrew could you please clarify the callback idea in more details? - Does it mean walk thru all available plugins and see which one implements some specific interface (like HttpAuthRequestHandler) and if so call corresponding method?

framework/src/org/apache/cordova/HttpAuthRequestHandler.java
sgrebnov: Thank you, will be updated!

framework/src/org/apache/cordova/CordovaWebViewClient.java
agrieve: Right now, almost all plugin hooks are implemented by adding a method to CordovaPlugin and PluginManager (which loops over all plugins and calls the corresponding CordovaPlugin method).

Calling plugins based on what interfaces they implement is a neat idea, although it'd be a new concept to the project. I'm not actually sure if there are benefits to it or not! Likely we could use that info to improve performance a bit, but perf isn't an issue atm afaik.

sgrebnov: Understand and thx!

sgrebnov:
Addressed all CR notes, @agrieve could you please take a look

sgrebnov:
Rebased on top of master branch

framework/src/org/apache/cordova/PluginManager.java
```diff
@@ -243,6 +243,27 @@ public void onPause(boolean multitasking) {
...
+     *
+     * @param view              The WebView that is initiating the callback
+     * @param handler           The HttpAuthHandler used to set the WebView's response
+     * @param host              The host requiring authentication
+     * @param realm             The realm for which authentication is required
+     * 
+     * @return                  Return True if there is a plugin which will resolve this auth challenge, otherwise False
+     * 
+     */
+    public Boolean onReceivedHttpAuthRequest(CordovaWebView view, ICordovaHttpAuthHandler handler, String host, String realm) {
```
agrieve: Unless you want this to be tri-state, then "Boolean" should be "boolean". I don't see the need for tristate, so go with "boolean"

agrieve:
Just the one nit-pick, then looks good to me! Feel free to just squash and commit it yourself :)

sgrebnov:
Done. @agrieve thx for review!

PR #149, opened 2015-01-18
CB-8317: Make it work to load about:blank and to dispatch exit message

shingotoda:
JIRA: https://issues.apache.org/jira/browse/CB-8317

Please take a look at my patch and give feedback. I'm not really confident on the modification. As alternative workaround, I thought isCurrentlyLoading should be set to true just after loadUrlNow() if an app is running on KitKat.

### DIFF
--- a/framework/src/org/apache/cordova/CordovaWebViewClient.java
+++ b/framework/src/org/apache/cordova/CordovaWebViewClient.java
@@ -193,14 +193,9 @@ public class CordovaWebViewClient extends WebViewClient {
     @Override
     public void onPageFinished(WebView view, String url) {
         super.onPageFinished(view, url);
-        // Ignore excessive calls.
-        //     On KitKat, loading about:blank does not invoke WebViewClient#onPageStarted
-        //     So do not check isCurrentLoading if app is running on KitKat
-        //         https://code.google.com/p/android/issues/detail?id=65701
-        if (android.os.Build.VERSION.SDK_INT != android.os.Build.VERSION_CODES.KITKAT){
-            if (!isCurrentlyLoading) {
-                return;
-            }
+        // Ignore excessive calls, if url is not about:blank (CB-8317).
+        if (!isCurrentlyLoading && !url.startsWith("about:")) {
+            return;
         }
         isCurrentlyLoading = false;
         LOG.d(TAG, "onPageFinished(" + url + ")");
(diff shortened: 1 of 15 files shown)


framework/src/org/apache/cordova/CordovaWebViewClient.java
```diff
@@ -171,8 +171,13 @@ public void onPageStarted(WebView view, String url, Bitmap favicon) {
     public void onPageFinished(WebView view, String url) {
         super.onPageFinished(view, url);
         // Ignore excessive calls.
-        if (!isCurrentlyLoading) {
-            return;
+        //     On KitKat, loading about:blank does not invoke WebViewClient#onPageStarted
+        //     So do not check isCurrentLoading if app is running on KitKat
+        //         https://code.google.com/p/android/issues/detail?id=65701
+        if (android.os.Build.VERSION.SDK_INT != android.os.Build.VERSION_CODES.KITKAT){
```
agrieve: Have you confirmed that it's working on Lollipop? If not, we should probably just remove the isCurrentlyLoading guard.

shingotoda: I just tested it on Lollipop and confirmed onPageStarted was NOT called for about:blank loading so my patch does not work on Lollipop. The same bug is still there in Android > KITKAT.

agrieve: Hmm, saw your comment about maybe setting it in loadUrlNow(). Wondering if maybe the best thing to do is just ignore the check when url.startsWith("about:")? WDYT? e.g.:

    if (!isCurrentlyLoading && !url.startsWith("about:")) {
        return;
    }

shingotoda: Ah, I agree with you on that. It looks the best thing. That worked fine as well.

shingotoda: So do I need to resend pull request again with revised patch?

agrieve: You don't need to re-request a PR, but you do need to make the change an upload it to github (either as a separate commit, or by force-pushing an ammended commit). Then, add a comment to let me know you've done so and I'll pull it in (once you do it, it should show up on this interface).

shingotoda: Thank you. I've uploaded it. Is it OK?

PR #151, opened 2015-01-27
CB-5059 Add a CookieManager abstraction for pluggable webviews.

dpogue:
Crosswalk and GeckoView implementations of CordovaWebView can provide their own ICordovaCookieManager implementation for plugins to use.

/cc @infil00p & @agrieve 
If you think this is the correct way to go about this, I'll make an implementation for the Crosswalk engine and update the FileTransfer plugin.

### DIFF
--- a/framework/src/org/apache/cordova/ICordovaCookieManager.java
+++ b/framework/src/org/apache/cordova/ICordovaCookieManager.java
@@ -1,0 +1,33 @@
+/*
+       Licensed to the Apache Software Foundation (ASF) under one
+       or more contributor license agreements.  See the NOTICE file
+       distributed with this work for additional information
+       regarding copyright ownership.  The ASF licenses this file
+       to you under the Apache License, Version 2.0 (the
+       "License"); you may not use this file except in compliance
+       with the License.  You may obtain a copy of the License at
+
+         http://www.apache.org/licenses/LICENSE-2.0
+
+       Unless required by applicable law or agreed to in writing,
+       software distributed under the License is distributed on an
+       "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+       KIND, either express or implied.  See the License for the
+       specific language governing permissions and limitations
+       under the License.
+*/
+
+package org.apache.cordova;
+
+public interface ICordovaCookieManager {
+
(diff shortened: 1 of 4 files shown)


framework/src/org/apache/cordova/ICordovaCookieManager.java
```diff
@@ -0,0 +1,45 @@
...
+       software distributed under the License is distributed on an
+       "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+       KIND, either express or implied.  See the License for the
+       specific language governing permissions and limitations
+       under the License.
+*/
+
+package org.apache.cordova;
+
+public interface ICordovaCookieManager {
```
agrieve: Looks great, but I think it would be wise to shrink the interface at the start. I'm thinking just:

- setCookiesEnabled()
- setThirdPartyCookiesEnabled()
- setCookie()
- getCookie()
- clearCookies()
- flush()

dpogue: My plan was to make it mostly API-compatible with existing CookieManagers so that it would be easier to migrate existing plugins rather than trying to figure out a different API.

One other issue is that the XWalkCookieManager doesn't appear to have setThirdPartyCookiesEnabled(): https://github.com/crosswalk-project/crosswalk/blob/master/runtime/android/core_internal/src/org/xwalk/core/internal/XWalkCookieManager.java

If those aren't concerns (I guess XWalkCordovaCookieManager could just ignore setThirdPartyCookiesEnabled?), then I'll shrink down the interface.

infil00p: The problem with using Android APIs and adopting them as our own is that we have to support stupid crap that isn't universal.  I'd rather add an API later that we needed than have some extra cruft that we have to test.  We should work on having the bare minimum needed for FileTransfer to work, and if a third-party API uses this, they can then submit a bug or send a pull request to add something to the API.  Adding to an API is an API change, but it's not a breaking one.

dpogue: Sounds good. FileTransfer was just using getCookie() as I recall, but I'll go with Andrew's list (minus setThirdPartyCookiesEnabled?)

PR #155, opened 2015-02-17
CB-8500 Change gradle signing property names to be the same as Ant

nikhilkh:


nikhilkh:
@agrieve @clelland  Can you please take a look at this change?

jsoref:
Please spell "signing" with two "n"s (the summary at least has this wrong, I suspect that the bug does too...)

nikhilkh:
@jsoref Fixed - good catch!

PR #160, opened 2015-02-22
CB-7827: Allow user to specify android activity name

cjpearson:
This change adds a parameter to create to allow the user to specify an activity name. If none is specified it will still default to MainActivity. See http://markmail.org/message/pj5ig4b54nk4l7dr for a discussion of this issue.


bin/create
```diff
@@ -23,14 +23,15 @@ var create = require('./lib/create');
 var args  = require('./lib/simpleargs').getArgs(process.argv);
 
 if (args['--help'] || args._.length === 0) {
-    console.log('Usage: ' + path.relative(process.cwd(), path.join(__dirname, 'create')) + ' <path_to_new_project> <package_name> <project_name> [<template_path>] [--link]');
+    console.log('Usage: ' + path.relative(process.cwd(), path.join(__dirname, 'create')) + ' <path_to_new_project> <package_name> <project_name> <activity_name> [<template_path>] [--link]');
```
agrieve: Shifting the `<template_path>` arg will break scripts that currently use it. Maybe use `--activity-name=` instead?

cjpearson: I wasn't sure of the best way to make both template_path and activity_name optional. I explored using --activity-name <activity_name> using nopt, which cordova-ios uses for arguments.

cjpearson:
@agrieve I've modified it to use --activity-name with nopt. I tried to follow the style of the iOS create script.

agrieve:
Merged. dcff8794adb5a8e772a

PR #161, opened 2015-02-22
Update gitignore

oliviertassinari:
Prevent www folder, automatically generated, to be committed.


agrieve:
This is generated when using the cordova npm tool, but it's still a valid workflow to use the `bin/create` script directly. I don't think we'd want to add this in that case.

PR #162, opened 2015-02-26
CB-3360 Set custom User-Agent

sinofool:
Add two configurations to set user agent of webview.
OverrideUserAgent to override UA to the setting value.
AppendUserAgent to append the setting value to default UA


framework/src/org/apache/cordova/engine/SystemWebViewEngine.java
```diff
@@ -199,7 +199,19 @@ private void initWebViewSettings() {
         
         // Fix for CB-1405
         // Google issue 4641
-        settings.getUserAgentString();
+        String defaultUserAgent = settings.getUserAgentString();
+
+        // Fix for CB-3360
+        String overrideUserAgent = getCordovaWebView().getPreferences().getString("OverrideUserAgent", null);
```
agrieve: I'd like to avoid circular dependency if possible on CordovaWebView.

Rather than depending on CordovaWebView here, could you save the instance of CordovaPreferences passed to the constructor and use that instead? Would also need to add CordovaPreferences as a 2nd arg to the other constructor as well.

sinofool: @agrieve Yes, thanks for the great suggestion, I will change it this weekend.

sinofool:
The Travis CL build failed due to new android-21 setup.

staugaard:
Yeah I also need this. :+1:

delijati:
Need this feature too. The old way in `onCreate` is not working anymore.

```
super.appView.getSettings().setUserAgentString("Your own agent string");
```

agrieve:
Merged. would be great if you could do another PR to update the docs:
https://github.com/apache/cordova-docs/blob/master/docs/en/edge/guide/platforms/android/config.md

agrieve:
(and thanks!)

sinofool:
I will update the docs later. For both iOS and Android.

shazron:
@sinofool There is already an issue for the iOS Doc CB-8960

delijati:
This solution is not sufficient for my needs i need to set the User-Agent from JS.

roysG:
How should we use it?

kleeb:
I guess that JS bridge is missing?

PR #163, opened 2015-03-16
Part of fix for CT-7992

jchorn:
New Cordova users, especially those building applications from building
blocks, sometimes end up with more than one <script> tag for cordova.js.
This currently results in bizarre behavior.  This change makes such
mistakes harmless, thus resulting in a more user-friendly experience.

### DIFF
--- a/framework/assets/www/cordova.js
+++ b/framework/assets/www/cordova.js
@@ -19,6 +19,9 @@
  under the License.
 */
 ;(function() {
+if (window.cordova) {
+    return; // protect against accidental multiple inclusion
+}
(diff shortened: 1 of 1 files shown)


purplecabbage:
If only it were that easy ...
This file is generated so we cannot just modify it. Changes need to be implemented in the cordova-js repo.

jchorn:
Thanks for the pointer!  I had a feeling all these platforms were sharing code somewhere.  I will try again with the right component.

purplecabbage:
This should handle it. Will test some more on windows and probably merge tomorrow. 
https://github.com/apache/cordova-js/pull/104

PR #164, opened 2015-03-17
CB-8484: Adds support creating signed archives for Android

nikhilkh:


bin/templates/cordova/lib/build.js
```diff
@@ -409,7 +442,29 @@ function parseOpts(options, resolvedTarget) {
             console.warn('Build option \'' + options[i] + '\' not recognized (ignoring).');
         }
     }
+    if (packageArgs.keystore && packageArgs.alias) {
+        var keystore = path.relative(ROOT, path.resolve(packageArgs.keystore));
+        ret.packageInfo = new PackageInfo(keystore, packageArgs.alias, packageArgs.keystorePassword,
+            packageArgs.password, packageArgs.keystoreType);
+    } else if (fs.existsSync(path.join(CORDOVAROOT, BUILD_CONFIG_FILE))) {
```
agrieve: We've made it this far without platforms having CLI-specific code, so I think we should not check at this file location. Instead, we could have "cordova prepare" create -signing.properties files, or have "cordova run" pass along all the signing args as CLI arguments. Could even pass it along as a single argument that points to the JSON file, but I don't think we should look for it here directly.

bin/templates/cordova/lib/build.js
```diff
@@ -422,11 +477,15 @@ function parseOpts(options, resolvedTarget) {
...
-    return builder.prepEnv()
+    return builder.prepEnv(opts)
     .then(function() {
-        return builder.clean(opts.extraArgs);
+        return builder.clean(opts);
     }).then(function() {
         shell.rm('-rf', path.join(ROOT, 'out'));
+    }).then(function() {
+        ['debug', 'release'].forEach(function(config) {
+            removeIfExists(path.join(ROOT, config + SIGNING_PROPERTIES));
```
agrieve: For non-CLI workflow, the user will likely create these files manually and be upset if clean deletes them. Probably best to just never delete them.

bin/templates/cordova/lib/build.js
```diff
@@ -530,8 +589,60 @@ module.exports.findBestApkForArchitecture = function(buildResults, arch) {
     throw new Error('Could not find apk architecture: ' + arch + ' build-type: ' + buildResults.buildType);
 };
 
+function removeIfExists(file) {
```
agrieve: nit: `shell.rm('-f', file)` will do this.

bin/templates/cordova/lib/build.js
```diff
@@ -542,5 +653,12 @@ module.exports.help = function() {
     console.log('    \'--versionCode=#\': Override versionCode for this build. Useful for uploading multiple APKs. Requires --gradle.');
     console.log('    \'--minSdkVersion=#\': Override minSdkVersion for this build. Useful for uploading multiple APKs. Requires --gradle.');
     console.log('    \'--gradleArg=<gradle command line arg>\': Extra args to pass to the gradle command. Use one flag per arg. Ex. --gradleArg=-PcdvBuildMultipleApks=true');
+    console.log('');
+    console.log('Signed APK flags:');
```
agrieve: Would be good to mention that these are overrides for what's already in `debug-signing.properties` / `release-signing.properties`

bin/templates/cordova/lib/build.js
nikhilkh: Good point. To delete these files the user has to explicitly run the `clean` script  from platforms\android\cordova. This made me think about other scenarios as well for CLI workflow:
1. cordova build <nopackageInfo>
2. cordova build <packageInfo>
3. cordova build <nopackageInfo>

The third invocation will use the pacakgeInfo from the previous invocation as I do not delete debug-signing.properties & release-signing.properies.

I can make this a bit more sophisticated to account for non-CLI workflows - detect if the file is auto-generated (It has the canned comment ("Do not modify this file -- YOUR CHANGES WILL BE ERASED")). If it is then I go ahead and delete it not only on `clean` but also when starting a build with no package info specified.

bin/templates/cordova/lib/build.js
nikhilkh: Resolved in next iteration

bin/templates/cordova/lib/build.js
nikhilkh: Resolved in next commit

bin/templates/cordova/lib/build.js
agrieve: hmm, maybe for this reason you should pass the parameters on the command line as gradle properties so that there is no file to clean up. Note that it's unlikely the ANT workflow is important anymore, as gradle is the new default and seems to work much better.

bin/templates/cordova/lib/build.js
nikhilkh: I did not realize that this was the design all along - of not looking into parent directories. I like your idea of passing the path to the JSON file as a single argument. I'm less inclined to do others:
1. 'codova prepare' creating -singing.properties file: There will require a lot of platform specific logic being in cordova-lib and less than ideal.
2. Pass multiple command line parameters: Escaping complex arguments correctly on command line might be tricky and messy.

From what I can understand, this results in two changes:
1. cordova-lib to pass path to build.json. Perhaps find a way to version this change - so that older versions of the platforms do not see this unexpected arg.
2. Android build script consuming this build.json and reading the `android` section of it.

bin/templates/cordova/lib/build.js
nikhilkh: I was really hoping to get the scenario of CLI + IDE (Android studio) interop working. If they both consume the same .properties file they can interoperate. The user could create a build.json file with the correct values and would be able to achieve this. I'm getting inclined to delete the file _only_ when it is autogenerated.

nikhilkh:
@agrieve I have finished addressing your code review feedback. Can you please take another look?

bin/templates/cordova/lib/build.js
```diff
@@ -372,6 +409,7 @@ function parseOpts(options, resolvedTarget) {
             if (multiValueArgs[flagName] && !flagValue) {
                 flagValue = options[i + 1];
                 ++i;
+                console.warn('A value is expected for the flag: ' + flagName);
```
agrieve: I think this warning will show when not using an = for the flag value (generally for flags with values, you're allowed to do: --foo=bar *or* --foo bar

bin/templates/cordova/lib/build.js
```diff
@@ -402,6 +440,16 @@ function parseOpts(options, resolvedTarget) {
...
                     break;
+                case 'keystore':
+                case 'alias':
+                case 'keytorePassword':
+                case 'password':
+                case 'keystoreType':
+                    packageArgs[flagName] = flagValue;
+                    break;
+                case 'buildConfig':
+                    buildConfig = flagValue.replace(/"/g, '');
```
agrieve: Shouldn't need the `.replace()` here... Did you find it necessary for some reason?

nikhilkh: If the path has a space (Windows) - the flag value will need to have quotes.

bin/templates/cordova/lib/build.js
```diff
@@ -530,8 +609,54 @@ module.exports.findBestApkForArchitecture = function(buildResults, arch) {
...
+        };
+    }
+}
+
+PackageInfo.prototype = {
+    toProperties: function() {
+        var self = this;
+        var result = '';
+        Object.keys(self).forEach(function(key) {
+            if (self[key]) {
```
agrieve: Probably should not require the value to be non-empty. Perfectly valid to have `key.alias.password=` in the file (although not recommended :P)

bin/templates/cordova/lib/build.js
```diff
@@ -530,8 +609,54 @@ module.exports.findBestApkForArchitecture = function(buildResults, arch) {
     throw new Error('Could not find apk architecture: ' + arch + ' build-type: ' + buildResults.buildType);
 };
 
+function PackageInfo(keystore, alias, keystorePassword, password, storeType) {
+    keystore = keystore.replace(/\\/g, '\\\\');
```
agrieve: This escaping will need to be done for all of the values I think. Would be more appropriate to do it within the toProperties function

bin/templates/cordova/lib/build.js
```diff
@@ -409,7 +457,31 @@ function parseOpts(options, resolvedTarget) {
...
+    } else if (fs.existsSync(buildConfig)) {
+        console.log('Reading build config file: '+ buildConfig);
+        var config = JSON.parse(fs.readFileSync(buildConfig, 'utf8'));
+        if (config.android && config.android[ret.buildType]) {
+            var androidInfo = config.android[ret.buildType];
+            if (androidInfo.keystore && androidInfo.alias) {
+                var configDir = path.dirname(buildConfig);
+                var keystorePath = path.relative(ROOT, path.join(configDir, androidInfo.keystore));
+                ret.packageInfo = new PackageInfo(keystorePath, androidInfo.alias,
+                    androidInfo.keystorePassword, androidInfo.password, androidInfo.storeType);
```
agrieve: A bit inconsistent to have "keystorePassword" and "storeType" (should be storePassword, or keystoreType). I think the [gradle terms](http://stackoverflow.com/questions/18328730/how-to-create-a-release-signed-apk-file-using-gradle) for these make sense, so might as well use them?

bin/templates/cordova/lib/build.js
```diff
@@ -409,7 +457,31 @@ function parseOpts(options, resolvedTarget) {
             console.warn('Build option \'' + options[i] + '\' not recognized (ignoring).');
         }
     }
+    if (packageArgs.keystore && packageArgs.alias) {
+        var keystore = path.relative(ROOT, path.resolve(packageArgs.keystore));
+        ret.packageInfo = new PackageInfo(keystore, packageArgs.alias, packageArgs.keystorePassword,
+            packageArgs.password, packageArgs.keystoreType);
+    } else if (fs.existsSync(buildConfig)) {
```
agrieve: Should just check: `if (buildConfig)`, and show an error if it's set and the file doesn't exist.

bin/templates/cordova/lib/build.js
```diff
@@ -297,6 +320,14 @@ var builders = {
                 var distributionUrl = 'distributionUrl=http\\://services.gradle.org/distributions/gradle-2.2.1-all.zip';
                 var gradleWrapperPropertiesPath = path.join(projectPath, 'gradle', 'wrapper', 'gradle-wrapper.properties');
                 shell.sed('-i', distributionUrlRegex, distributionUrl, gradleWrapperPropertiesPath);
+            }).then(function() {
```
agrieve: nit: don't need this to be in a separate promise.

nikhilkh: Fair enough - removed all superfluous promises in next iteration.

bin/templates/cordova/lib/build.js
```diff
@@ -409,7 +457,31 @@ function parseOpts(options, resolvedTarget) {
             console.warn('Build option \'' + options[i] + '\' not recognized (ignoring).');
         }
     }
+    if (packageArgs.keystore && packageArgs.alias) {
```
agrieve: Wonding if this should actually not be exclusive with the else if clause. E.g. allow specifying buildConfig, and then also allow setting the password via --password

bin/templates/cordova/lib/build.js
agrieve: I like it!

bin/templates/cordova/lib/build.js
nikhilkh: Good point! Resolved in next iteration.

bin/templates/cordova/lib/build.js
nikhilkh: Good point. Resolved in next iteration.

bin/templates/cordova/lib/build.js
nikhilkh: Yup. That's a better place for it. Moved it next iteration.

bin/templates/cordova/lib/build.js
nikhilkh: I see - I get it now.

nikhilkh:
@agrieve I have resolved all the code review comments. Let me know if this is ready - I can rebase this to make it ready for merge.

bin/templates/cordova/lib/build.js
agrieve: Quotes for these reasons are generally interpreted by the shell (cmd.exe), and when we see it here, they should be gone already (and flagValue will have a space in it).

bin/templates/cordova/lib/build.js
```diff
@@ -422,11 +500,18 @@ function parseOpts(options, resolvedTarget) {
...
     .then(function() {
-        return builder.clean(opts.extraArgs);
+        return builder.clean(opts);
     }).then(function() {
         shell.rm('-rf', path.join(ROOT, 'out'));
+
+        ['debug', 'release'].forEach(function(config) {
+            var propertiesFilePath = path.join(ROOT, config + SIGNING_PROPERTIES);
+            if(isAutoGenerated(propertiesFilePath)){
+                shell.rm('-f', propertiesFilePath);
```
agrieve: prepEnv (which is called in this case) takes care of deleting these if there is no supplied signing args. I don't think there's a need to delete them unconditionally since we don't bother deleting other build-time-supplied-files (like build.xml, build.gradle)

bin/templates/cordova/lib/build.js
```diff
@@ -360,9 +391,15 @@ function parseOpts(options, resolvedTarget) {
     var multiValueArgs = {
       'versionCode': true,
       'minSdkVersion': true,
-      'gradleArg': true
+      'gradleArg': true,
+      'keystore' : true,
+      'alias' : true,
+      'password' : true,
+      'keytorePassword' : true,
```
agrieve: Missed renaming  this in the switch stmt, and in the help text at the bottom.

agrieve:
Just a couple more comments then we are good to go! Thanks for working through this with me. Feel free to rebase & squash commits.

bin/templates/cordova/lib/build.js
nikhilkh: That's true only when the argument is specified separately e.g. build --buildConfig "foo bar\build.json" - the shell will remove the quotes. But not for a scenario like --buildConfig="foo bar\build.json"

bin/templates/cordova/lib/build.js
nikhilkh: Good catch! Fixed in next iteration!

bin/templates/cordova/lib/build.js
nikhilkh: My previous statement is not true. I think I found my issue where I was explicitly adding quotes in cordova-lib code which was escaped by superspawn. I do not need this and will remove it. Thanks!

bin/templates/cordova/lib/build.js
nikhilkh: My main rationale was that these .properties file have sensitive information including passwords and I wanted developers to have a way to clean this if needed. Invoking the 'clean' script will run this and delete these files.

agrieve: I don't think clean is really used (the command isn't even wired up to CLI afaik, although it really should be). If passwords are the concern, then probably we should be deleting the files after building.

nikhilkh: Yes, currently its not wired up to the CLI. We have received feedback that it should be wired. I agree deleting after every build is the most 'secure' - though it breaks the android studio IDE interop story. My thinking is a developer will specify the build.json file with the required parameters and iterate using a combination of Android studio and CLI.
Some of this is hard to predict, I'm of the opinion that let's get it out there and drive some usage and listen to feedback.
Let me know if you feel strongly about removing the deletion of these files from 'clean'. Having this be there 'clean' allows the developer a way to remove these files and not rely on internals.

agrieve: I don't feel strongly. Let's ship it! :)

nikhilkh:
@agrieve Fixed last set of issues & rebased! This should be ready for a merge.

agrieve:
Sorry - just realized one more thing. You'll need to add the new flags to lib/run.js as well or else they won't work with the "cordova run" command.

agrieve:
(esp. for debug signing keys, it's important that they are used not just when packaging for release)

nikhilkh:
@agrieve  Good catch - totally missed this one. Fixed now.

agrieve:
Merged!

nikhilkh:
Thanks, @agrieve ! :clap:

PR #168, opened 2015-03-26
CB-8753 Maintain splash screen aspect ratio

AlexTalis:
This pull request implements https://issues.apache.org/jira/browse/CB-8753. Apologies for multiple commits. I was trying to edit the commit message to include CB-8753 in the summary and ended up with three commits instead of one. I'm new to Git and Github.

There are some things about existing code that I did not completely understand. I want to list them here to help in the review process. If you know what I didn't understand it may help you find problems with my code changes.

- Is there a reason why most fields in SplashScreenInternal are static? It seems like they could just be private instance fields.

- What is the purpose of adding preferences to activity's intent extras in CordovaPreferences.copyIntoIntentExtras(Activity)? Not all preferences are copied and I did not add my new preferences either. Should I?

- Why does SplashScreenInternal keep getting preferences from CordovaPreferences object instead of just storing them in instance fields during initialization? For example, drawableId is retrieved like this throughout the class:

    int drawableId = preferences.getInteger("SplashDrawableId", 0);

Why not just have a drawableId instance variable and store the value in it once during initialization?

- Is it OK that the new preferences are only supported on Android?

### DIFF
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -34,6 +34,7 @@ import android.app.AlertDialog;
 import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
+import android.content.res.Configuration;
 import android.graphics.Color;
 import android.media.AudioManager;
 import android.os.Bundle;
@@ -131,6 +132,11 @@ public class CordovaActivity extends Activity implements CordovaInterface {
     protected Whitelist externalWhitelist;
     protected String launchUrl;
     protected ArrayList<PluginEntry> pluginEntries;
+    
+    /**
+     * Last orientation of device. Used to detect orientation changes.
+     */
+    private int orientation;
 
     /**
     * Sets the authentication token.
@@ -219,6 +225,9 @@ public class CordovaActivity extends Activity implements CordovaInterface {
         }
 
         super.onCreate(savedInstanceState);
+        
+        // Save startup orientation. Used to determine when orientation changes in "onConfigurationChanged".
+        orientation = getResources().getConfiguration().orientation;
 
         if(savedInstanceState != null)
         {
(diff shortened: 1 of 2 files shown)


agrieve:
Great questions.

The statics are meant to prevent multiple splash screens from showing if there is more than one CordovaWebView within a project.

`CordovaPreferences.copyIntoIntentExtras(Activity)` - it's legacy.

keep getting preferences from CordovaPreferences - because embedders can change these preferences during app start-up, and we want those changes to take effect.

New Prefs android only - yes. that's fine. Splash screen already has mostly platform-specific prefs :(

Overall comments:
- Randomly, an `onConfigurationChange` hook was just recently added to master, so you should just use that instead.
- We aren't planning on doing another release of 3.x branch, so you should make these changes to cordova-plugin-splashscreen, which is what will work with cordova-android@4.0.x (and with master branch of cordova-android)
- I don't see why anyone would want reloadOnOrientationChange=false, so maybe don't make that a preference?
- Might be better to get rid of the LinearLayout, and instead switch to an ImageView in both cases

AlexTalis:
I'm working on a pull request for cordova-plugin-splashscreen master. The only way I found to set up for the plugin development was to copy SplashScreen.java from the repo into another project, make the changes, then copy it back to cordova-plugin-splashscreen repo, then push. That's because SplashScreen.java in cordova-plugin-splashscreen is not in the right package folder structure, so it can't be linked into a workspace directly. This is a little painful. Do you have better way of doing this?

agrieve:
Install the plugin with the --link flag, and then you should be able to edit it in the projects and the changes will be reflected in the repo (it creates symlinks)

AlexTalis:
It doesn't work for me on Windows, probably because my repo structure is totally different from what you use. I'm not going to bug you with more questions - it's not that big of a deal. I can manually create the link.

AlexTalis:
Closing this pull request in favor of this one: https://github.com/apache/cordova-plugin-splashscreen/pull/43

PR #170, opened 2015-03-29
solves CB-8768 issue where onActivityResult gets called before onResume after MainActivity gets killed

thehuijb:
if one of the plugins launches startActivityForResult and the Android OS decides to kill our MainActivity. what happens when the launched activity is closed.
the previous activity (our MainActivity is relaunched and onActivityResult is called, our mainActivity however has not yet had the time to load all onload=true plugins so therefore the savedInstanceState callBackService can not be called.

### DIFF
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -20,6 +20,8 @@ package org.apache.cordova;
 
 import java.util.ArrayList;
 import java.util.Locale;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
 
 import org.json.JSONException;
 import org.json.JSONObject;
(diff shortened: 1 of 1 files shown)


framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -314,10 +318,52 @@ public void startActivityForResult(Intent intent, int requestCode, Bundle option
...
+        if (this.pluginsReady) {
+            cordovaInterface.onActivityResult(requestCode, resultCode, intent);
+            this.findCallbackTries = 0;
+        } else {
+            /**
+             * If the Android OS kills this activity when a plugin launches an a new activity 
+             * the onActivityResult event fires before the onResume event 
+             * so we have to wait for the plugins to be loaded again before we can hand the result to the correct plugin.
+             */
+            final ScheduledThreadPoolExecutor exec = new ScheduledThreadPoolExecutor(1);
```
agrieve: Using a separate thread to accomplish what you want here is a bit overcomplicated and as is has the unwanted side-effect that the callback will be called on a non-UI thread. We also don't want to wait for a resume event to fire this.

I'd suggest instead:
- Call `CordovaInterfaceImpl.onActivityResult()`, as before (probably just revert all changes to CordovaActivity).
- Have `onActivityResult` just store the result if no initialization has happened yet.
- Change `CordovaInterfaceImpl.setPluginManager(pluginManager)` to `onCordovaInit(pluginManager)`
- Have `onCordovaInit` dispatch any pending result to the pluginManager (plugins services should be registered by now).

thehuijb: thanks for having a look and commenting.

PR #171, opened 2015-03-31
solves CB-8768 issue where onActivityResult gets called before onResume after MainActivity gets killed

thehuijb:
situation: one of the plugins launches startActivityForResult and the Android OS decides to kill our MainActivity.
once the launched activity is fulfilled it comes back to our MainActivity, which has to be recreated first.
unfortunately Android calls onActivityResult before our Activity has fully loaded our installed plugins.

this commit fixes that issue. thanks to suggestions by Andrew Grieve

### DIFF
--- a/framework/src/org/apache/cordova/CordovaInterfaceImpl.java
+++ b/framework/src/org/apache/cordova/CordovaInterfaceImpl.java
@@ -13,6 +13,9 @@ import java.util.concurrent.Executors;
  */
 public class CordovaInterfaceImpl implements CordovaInterface {
     private static final String TAG = "CordovaInterfaceImpl";
+    
+    private ActivityResultHolder savedResult;
+
     protected Activity activity;
     protected ExecutorService threadPool;
     protected PluginManager pluginManager;
@@ -30,8 +33,12 @@ public class CordovaInterfaceImpl implements CordovaInterface {
         this.threadPool = threadPool;
     }
 
-    public void setPluginManager(PluginManager pluginManager) {
+    @Override
+    public void onCordovaInit(PluginManager pluginManager) {
         this.pluginManager = pluginManager;
+        if (savedResult != null) {
+            onActivityResult(savedResult.getRequestCode(), savedResult.getResultCode(), savedResult.getIntent());
+        }
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


framework/src/org/apache/cordova/CordovaInterfaceImpl.java
```diff
@@ -119,4 +130,29 @@ public void onSaveInstanceState(Bundle outState) {
...
+        private int resultCode;
+        private Intent intent;
+
+        public ActivityResultHolder(int requestCode, int resultCode, Intent intent) {
+            this.requestCode = requestCode;
+            this.resultCode = resultCode;
+            this.intent = intent;
+        }
+
+        public int getRequestCode() {
```
agrieve: nit: no point in creating getters for a private class

framework/src/org/apache/cordova/CordovaInterfaceImpl.java
```diff
@@ -119,4 +130,29 @@ public void onSaveInstanceState(Bundle outState) {
     public void restoreInstanceState(Bundle savedInstanceState) {
         initCallbackService = savedInstanceState.getString("callbackService");
     }
+    
+    private class ActivityResultHolder {
```
agrieve: nit: should be "private static class"

framework/src/org/apache/cordova/CordovaWebViewImpl.java
```diff
@@ -500,6 +500,7 @@ public void onPageFinishedLoading(String url) {
 
             clearLoadTimeoutTimer();
 
+            cordova.onCordovaInit(pluginManager);
```
agrieve: I'm not sure it makes sense to wait until onPageFinished to dispatch onCordovaInit. Why not dispatch at the end of the .init() function?

If the concern is that you want to be able to communicate with the JS, I don't think onPageFinished guarantees that. Instead, you should have your plugin wait for an `exec()` to come in from the JS in order to communicate with it on start-up.

thehuijb: Applied your suggestions,
please review once more.

thanks in advance

agrieve:
Awesome! Merged and made one minor change: Removed onCordovaInit from CordovaInterface (but left it in CordovaInterfaceImpl). If it were to be called from CordovaWebViewImpl, then it would belong in the interface, but not otherwise.

thehuijb:
great, thanks for that, makes sense.

PR #173, opened 2015-04-08
fix CB-8684 - support onStart/onStop lifecycle events in plugins for Android

tony--:
A couple notes about this PR.
1. I feel pretty sure that CordovaActivity.shouldHandleStartOnAppViewInit is needed in order to reliably get an onStart event to interested plugins, but not sure about CordovaWebViewImpl.shouldCallOnStartDuringInit.  Don't want to add cruft, but would like these events to be reliable.
2. I tried to add a reasonable test for this new feature, but I feel like there might be a more elegant approach?  I'm open to feedback and willing to modify the approach.  Also (per @agrieve suggestion), I set onload=true, but it seemed to be working without it (which seems a little weird, but I didn't want to delay this PR any further - will step through it later).


test/androidTest/src/org/apache/cordova/test/CordovaPluginTest.java
```diff
@@ -0,0 +1,126 @@
...
+                synchronized(this)
+                {
+                    this.notify();
+                }
+            }
+        };
+        synchronized( blockingRunnable ) {
+            testActivity.runOnUiThread(blockingRunnable);
+            try {
+                blockingRunnable.wait();
```
agrieve: The normal way (I think) is to use `getInstrumentation().waitForIdleSync();` You should switch to this or add a comment saying why the custom wait is necessary.

test/androidTest/src/org/apache/cordova/test/CordovaPluginTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+    protected void setUp() throws Exception {
+        super.setUp();
+        testActivity = getActivity();
+        cordovaWebView = testActivity.getCordovaWebView();
+        testPlugin = (LifeCyclePlugin)cordovaWebView.getPluginManager().getPlugin("LifeCycle");
+    }
+
+    private void invokeBlockingCallToLifeCycleEvent(final String lifeCycleEventName) {
+        final CordovaPluginTest me = this;
```
agrieve: nit: no need to alias the containing scope. Just use `getInstrumention()` rather than `me.getInstrumentation()`. If you really want to be explicit, you can use `CordovaPluginTest.this.getInstrumentation()`

test/src/org/apache/cordova/test/LifeCyclePluginTestActivity.java
```diff
@@ -0,0 +1,44 @@
...
+       under the License.
+*/
+package org.apache.cordova.test;
+
+import android.os.Bundle;
+
+import org.apache.cordova.ConfigXmlParser;
+import org.apache.cordova.CordovaWebView;
+
+public class LifeCyclePluginTestActivity extends BaseTestCordovaActivity {
```
agrieve: Doesn't look like creating this custom activity is necessary. Just use `MainTestActivity`

test/androidTest/src/org/apache/cordova/test/CordovaPluginTest.java
```diff
@@ -0,0 +1,126 @@
...
+            try {
+                blockingRunnable.wait();
+            } catch (InterruptedException e) {
+                fail("An InterruptedException occurred in invokeBlockingCallToLifeCycleEvent while waiting for " + lifeCycleEventName);
+            }
+        }
+
+    }
+
+    public void testPluginLifeCycle() throws IOException {
```
agrieve: Nice!

test/androidTest/src/org/apache/cordova/test/CordovaPluginTest.java
```diff
@@ -0,0 +1,126 @@
...
+
+    public static String TAG = "CordovaPluginTest";
+    protected LifeCyclePluginTestActivity testActivity;
+    protected LifeCyclePlugin testPlugin;
+    protected CordovaWebView cordovaWebView;
+
+    @SuppressWarnings("deprecation")
+    public CordovaPluginTest() {
+
+        super("org.apache.cordova.test", LifeCyclePluginTestActivity.class);
```
agrieve: This ctor is deprecated. Would make more sense to have this class extend `BaseCordovaIntegrationTest` I think.

test/androidTest/src/org/apache/cordova/test/CordovaPluginTest.java
```diff
@@ -0,0 +1,126 @@
+
+package org.apache.cordova.test;
+
+/*
```
agrieve: Put the copyright above the package declaration.

framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -95,6 +95,9 @@ Licensed to the Apache Software Foundation (ASF) under one
     protected ArrayList<PluginEntry> pluginEntries;
     protected CordovaInterfaceImpl cordovaInterface;
 
+    //flag for case when appView is null during onStart
+    private boolean shouldHandleStartOnAppViewInit = false;
```
agrieve: The webview (and thus plugins) are generally created from onCreate(), which means they will be initialized before the first call to onStart. If you're unit tests passes without this flag (and without the one in CordovaWebViewImpl), then I think you should remove them. It's not obvious (to me anyways) that this behaviour would even be the desired behaviour. If there's a reason it needs to be this way, then let me know! But if so, then we should upated onResume as well so that we're consistent.

test/src/org/apache/cordova/test/LifeCyclePluginTestActivity.java
tony--: Originally was trying to isolate the plugin loading so I wouldn't have to modify config.xml.  But now that I modified config.xml... I'll switch to MainTestActivity.

test/androidTest/src/org/apache/cordova/test/CordovaPluginTest.java
tony--: Ok.

test/androidTest/src/org/apache/cordova/test/CordovaPluginTest.java
tony--: I wasn't sure how to do it.  I looked over the docs and didn't see anything.  I'll read the doc for getInstrumentation().waitForIdleSync(); and switch.

test/androidTest/src/org/apache/cordova/test/CordovaPluginTest.java
tony--: Whoops!  I used one of the other tests as a template - the package is above the copyright in all of them.  I'll move it regardless.

framework/src/org/apache/cordova/CordovaActivity.java
tony--: My tests pass without it.  I'll remove it.  I'll also remove the similar flag from CordovaWebViewImpl.

tony--:
I think I made changes that address all of your comments.  Thanks for the feedback.

tony--:
Whoops, forgot to remove LifeCyclePluginTestActivity from the manifest.  Thanks for catching it!

svranch:
unsubscribe

 

Sandi Laufenberg 

Cell  425.633.7888

sandi@ladysword.biz  

Portfolio:   www.LadySword.com <http://www.ladysword.com/>

rejhgadellaa:
Hi, quick question: is this change just for plugins or can I hook into the onStart and onStop events in javascript as well? Because with Nougat apps need to be able to know if they stopped (not visible) or might be in split view but not focused (paused). In the latter case the app should probably still refresh its UI and such.

PR #174, opened 2015-04-15
fix package name validation

boynet:
you can see accepted applications in google play that contain number in the first part of package name, those apps cant be build with  because of this validation bug\inconsistent with google play. for example this package: http://bit.ly/1aVO6Re

### DIFF
--- a/bin/lib/create.js
+++ b/bin/lib/create.js
@@ -150,7 +150,7 @@ function validatePackageName(package_name) {
     //Make the package conform to Java package types
     //http://developer.android.com/guide/topics/manifest/manifest-element.html#package
     //Enforce underscore limitation
-    if (!/^[a-zA-Z][a-zA-Z0-9_]+(\.[a-zA-Z][a-zA-Z0-9_]*)+$/.test(package_name)) {
+    if (!/^[a-zA-Z0-9][a-zA-Z0-9_]+(\.[a-zA-Z][a-zA-Z0-9_]*)+$/.test(package_name)) {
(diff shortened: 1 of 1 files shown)


agrieve:
I do believe that the first letter of the name cannot be a number, which is what the current regexp checks for (non-first characters can be numbers)

boynet:
this package name failed: "be111.net" with the error of "Package name must look like: com.company.Name" although is accepted by google so maybe something else in the regex is wrong but the fix I made fix it for me you can see simmilar package name here: https://play.google.com/store/apps/details?id=be106.net

purplecabbage:
The platform agnostic version of this check is here:
https://github.com/purplecabbage/valid-identifier/blob/master/valid-identifier.js

The regex in use should probably be the same as there.
var regX = /([a-zA-Z_$][a-zA-Z\d_$]*\.)*[a-zA-Z_$][a-zA-Z\d_$]*/;
or refactored to use valid-identifier.

validateIdentifier("be111.net") passes!

agrieve:
@boynet - your fiddle doesn't use the regex that's checked in.
http://jsfiddle.net/hakxc2uv/

shows that the regexp as-is works fine for your case (I just copied and pasted it from the "Files changed" section of this PR)

boynet:
@agrieve omg you are right, thanks.
i saw it here https://github.com/crosswalk-project/crosswalk-cordova-android/blob/master/bin/lib/create.js#L161 and I was sure they just fork you version sorry

PR #175, opened 2015-04-23
CB-8902: use immersive mode when available and "fullscreen" is set to 'true'

kicktheken:
"Fullscreen" preference set to 'true' in the config.xml should hide the navigation bar on android devices that support immersive mode.

### DIFF
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -31,6 +31,7 @@ import android.content.Intent;
 import android.content.res.Configuration;
 import android.graphics.Color;
 import android.media.AudioManager;
+import android.os.Build;
 import android.os.Bundle;
 import android.util.Log;
 import android.view.Menu;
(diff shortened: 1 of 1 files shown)


framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -130,6 +147,18 @@ public void onCreate(Bundle savedInstanceState) {
             cordovaInterface.restoreInstanceState(savedInstanceState);
         }
     }
+
+    private void setSystemUiVisibilityMode(Window window) {
+        final int uiOptions =
+                    View.SYSTEM_UI_FLAG_LAYOUT_STABLE
```
agrieve: Are these LAYOUT flags really necessary? (if so, add a comment explaining why)

framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -115,8 +116,24 @@ public void onCreate(Bundle savedInstanceState) {
             getWindow().setFlags(WindowManager.LayoutParams.FLAG_FULLSCREEN,
                     WindowManager.LayoutParams.FLAG_FULLSCREEN);
         } else if (preferences.getBoolean("Fullscreen", false)) {
-            getWindow().setFlags(WindowManager.LayoutParams.FLAG_FULLSCREEN,
+
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.KITKAT)
+            {
+                setSystemUiVisibilityMode(getWindow());
+                getWindow().getDecorView().setOnSystemUiVisibilityChangeListener(new View.OnSystemUiVisibilityChangeListener() {
```
agrieve: I don't think this listener should be necessary. According to:
https://developer.android.com/training/system-ui/immersive.html

setting IMMERSIVE_STICKY should take care of re-entering fullscreen.

framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -115,8 +116,24 @@ public void onCreate(Bundle savedInstanceState) {
             getWindow().setFlags(WindowManager.LayoutParams.FLAG_FULLSCREEN,
```
agrieve: SetFullscreen is deprecated, but we shouldn't make it behave differently from "Fullscreen". How about changing this to:

    if (SetFullscreen) {
      Log();
      preferences.set("Fullscreen", true);
    }
    if (Fullscreen) {
      ...
    }

framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -130,6 +147,18 @@ public void onCreate(Bundle savedInstanceState) {
             cordovaInterface.restoreInstanceState(savedInstanceState);
         }
     }
+
+    private void setSystemUiVisibilityMode(Window window) {
```
agrieve: If you're going to pass in the window, then you should make this method static.
If you're going to call it "set...Mode", then it should accept the mode as a parameter.
Really though, hoping you don't need the listener so that you don't need a helper at all.

kicktheken:
I've gotten rid of the listener and used the suggested approach from the Android docs to keep immersive mode.

pwnall:
I look forward to using this! :heart:

PR #176, opened 2015-05-22
CB-8954 Adds support for `requirements` command

vladimir-kotikov:
Implementation for https://issues.apache.org/jira/browse/CB-8954

### DIFF
--- a/bin/lib/check_reqs.js
+++ b/bin/lib/check_reqs.js
@@ -40,11 +40,13 @@ function forgivingWhichSync(cmd) {
     }
 }
 
-function tryCommand(cmd, errMsg) {
+function tryCommand(cmd, errMsg, catchStderr) {
     var d = Q.defer();
     child_process.exec(cmd, function(err, stdout, stderr) {
         if (err) d.reject(new Error(errMsg));
-        else d.resolve(stdout);
+        // Sometimes it is necessary to return an stderr instead of stdout in case of success, since
+        // some commands prints theirs output to stderr instead of stdout. 'javac' is the example
+        else d.resolve((catchStderr ? stderr : stdout).trim());
     });
     return d.promise;
 }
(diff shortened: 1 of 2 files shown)


bin/lib/check_reqs.js
```diff
@@ -196,20 +208,22 @@ module.exports.check_android = function() {
...
             process.env['PATH'] += path.delimiter + path.join(process.env['ANDROID_HOME'], 'platform-tools');
         }
         if (!process.env['ANDROID_HOME']) {
-            throw new Error('ANDROID_HOME is not set and "android" command not in your PATH. You must fulfill at least one of these conditions.');
+            throw new Error('Failed to find \'ANDROID_HOME\' environment variable. Try setting setting it manually.\n' +
+                'Failed to find \'android\' command in your \'PATH\'. Try update your \'PATH\' to include path to valid SDK directory.');
         }
         if (!fs.existsSync(process.env['ANDROID_HOME'])) {
-            throw new Error('ANDROID_HOME is set to a non-existant path: ' + process.env['ANDROID_HOME']);
+            throw new Error('\'ANDROID_HOME\' environment variable is leads to non-existant path: ' + process.env['ANDROID_HOME'] +
```
TimBarham: Typo - 'existent' (I know - not new code, but good to fix while you're here)

bin/lib/check_reqs.js
```diff
@@ -223,27 +237,87 @@ module.exports.check_android_target = function(valid_target) {
...
     });
 };
 
+/**
+ * Object thar represents one of requirements for current platform.
+ * @param {String} id         The unique identifier for this requirements.
+ * @param {String} name       The name of requirements. Human-readable field.
+ * @param {String} version    The version of requirement installed. In some cases could be an array of strings
+ *                            (for example, check_android_target returns an array of android targets installed)
+ * @param {Boolean} installed Indicates whether the reuirement is installed or not
```
TimBarham: Typo - 'requirement'

TimBarham: My IDE wants you to mark `version` and `installed` as optional :)

bin/lib/check_reqs.js
```diff
@@ -70,18 +72,23 @@ module.exports.get_target = function() {
...
 // Returns a promise. Called only by build and clean commands.
 module.exports.check_gradle = function() {
     var sdkDir = process.env['ANDROID_HOME'];
-    var wrapperDir = path.join(sdkDir, 'tools', 'templates', 'gradle', 'wrapper');
-    if (!fs.existsSync(wrapperDir)) {
-        return Q.reject(new Error('Could not find gradle wrapper within android sdk. Might need to update your Android SDK.\n' +
-            'Looked here: ' + wrapperDir));
-    }
-    return Q.when();
+    var wrapper = path.join(sdkDir, 'tools', 'templates', 'gradle', 'wrapper', 'gradlew');
```
TimBarham: If `ANDROID_HOME` is not defined, the error you will get here is:

    Gradle: not installed
    [TypeError: Arguments to path.join must be strings]

(since you end up passing null to `path.join()`)

While there is a previous check for `ANDROID_HOME` (as part of the Android SDK check), it would be nice to make this error more meaningful.

bin/lib/check_reqs.js
TimBarham: "is leads to" isn't good grammar :smile: ... Either leave it as it was ('is set to'), or just 'leads to' (I actually think it worked fine how it was).

TimBarham:
In general looks good. A few minor comments.

PR #177, opened 2015-05-30
Consider treating gamepad B button as a Back button

matb33:
I don't expect this to get merged, but am simply putting this out there for discussion. The pull request is simply here for convenience to quickly show you a diff.

The motivation behind this change is driven by how the `keyCode` value for many gamepad buttons is reported as `0` in `keydown` (and `keyup`) events in Javascript.  It makes it pretty much impossible to manually discern the `B` button from the others. The `B` button is, as far as I know, universally accepted as a "back" button.

I realize implementing gamepad code manually using the HTML5 APIs would allow one to discern buttons properly. The downside to doing this is that firing events from JS means that they are no longer "user-initiated", so mobile browsers will prevent things like onscreen keyboards from appearing when focusing or clicking on a text field.

By treating the `B` button as a `back` button, we can at least treat `keyCode` `0` events as "select", and safely handle the `B` button separately as the `backbutton` event.


infil00p:
Discussion about this should happen in JIRA or on the list.  This is a year old, and got completely ignored because nobody looked at it.  Since it's not intended on actually making it into Cordova, I'm going to close it since this isn't the right spot for open-ended feature discussions.

PR #178, opened 2015-06-03
Grant Lollipop permission requests

vbraun:
This patch overrides onPermissionRequest so that getUserMedia can be used inside the browser.

Since a hybrid app has to request permissions anyways via AndroidManifest.xml, I think it is unnecessary to have any further configuration for onPermissionRequest. Anything that the app is allowed to do should be possible from the JS side. Hence all requests are granted. This enables getUserMedia (and WebRTC) on Android Lollipop, without resorting to crosswalk.

The docs say that request.grant has to be called from the UI thread, but don't explicitly spell out whether onPermissionRequest is called from the UI thread. I think that this is so, the WebChromeClient of course makes its calls from the UI thread unless otherwise noted. So there is no need to post a runnable to the UI thread.

### DIFF
--- a/framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
+++ b/framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
@@ -18,6 +18,7 @@
 */
 package org.apache.cordova.engine;
 
+import java.util.Arrays;
 import android.annotation.TargetApi;
 import android.app.Activity;
 import android.content.ActivityNotFoundException;
@@ -36,6 +37,7 @@ import android.webkit.ValueCallback;
 import android.webkit.WebChromeClient;
 import android.webkit.WebStorage;
 import android.webkit.WebView;
+import android.webkit.PermissionRequest;
 import android.widget.LinearLayout;
 import android.widget.ProgressBar;
 import android.widget.RelativeLayout;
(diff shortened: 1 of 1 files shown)


vbraun:
Corresponding Apache JIRA issue: https://issues.apache.org/jira/browse/CB-9115

mccob:
How we are supposed to invoke permission request on Android M with cordova ? getUserMedia doesn't produce any permission request and leave errors from accessing media

dpogue:
Maybe we need a plugin that acts as a polyfill for the [W3C Permissions API](https://w3c.github.io/permissions/) `navigator.permission.request()` to programmatically invoke permission requests?

infil00p:
@dpogue That's exactly what needs to happen, and currently this is only half-done so I can't accept this pull request as is.

infil00p:
BTW: I'm going to close this and recommend that the approach dpogue is mentioning be taken.  We can't guarantee behaviour across multiple webviews.

anatoly-khomenko:
Hello,

@infil00p 
Dear Joe, could you point to some solution to handle this currently? We need to use camera from WebView in a Cordova application, but this does not seem to be possible as far as I understand it now.

infil00p:
Weird, I guess I decided to accept this after all.  I wish I had documentation as to why I accepted the PR instead of rejecting it.

PR #180, opened 2015-06-09
CB-9119 Fixing intermittent 'adb install' hanging issue

dblotsky:
Adding lib/retry.js for retrying promise-returning functions. Retrying 'adb install' in emulator.js to work around it hanging sometimes.

Related Android bug: https://code.google.com/p/android/issues/detail?id=10255

### DIFF
--- a/bin/templates/cordova/lib/emulator.js
+++ b/bin/templates/cordova/lib/emulator.js
@@ -21,13 +21,21 @@
 
 /* jshint sub:true */
 
-var exec  = require('./exec'),
-    Q     = require('q'),
-    os    = require('os'),
-    appinfo = require('./appinfo'),
-    build = require('./build'),
-    child_process = require('child_process');
+var exec       = require('./exec');
+var appinfo    = require('./appinfo');
+var retry      = require('./retry');
+var build      = require('./build');
 var check_reqs = require('./check_reqs');
+
+var Q             = require('q');
+var os            = require('os');
+var child_process = require('child_process');
+
+// constants
+var ONE_SECOND              = 1000; // in milliseconds
+var INSTALL_COMMAND_TIMEOUT = 120 * ONE_SECOND; // in milliseconds
+var NUM_INSTALL_RETRIES     = 3;
+var EXEC_KILL_SIGNAL        = 'SIGKILL';
 
 /**
  * Returns a Promise for a list of emulator images in the form of objects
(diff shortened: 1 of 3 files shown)


bin/templates/cordova/lib/exec.js
```diff
@@ -19,16 +19,42 @@
...
+//         child_process.exec expects
+//
+// NOTE:
+//      exec documented here - https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback
+module.exports = function(cmd, opt_cwd, options) {
+
     var d = Q.defer();
+
+    if (options === undefined) {
+        options = {};
```
nikhilkh: Is this changing the behavior of providing max buffer when no options are specified?

bin/templates/cordova/lib/emulator.js
```diff
@@ -298,37 +306,67 @@ module.exports.resolveTarget = function(target) {
...
-            return Q.reject('Failed to launch app on emulator: ' + err);
         });
+
+    // unlock screen
+    }).then(function () {
+
+        console.log('Unlocking screen...');
+        return exec('adb -s ' + target.target + ' shell input keyevent 82', os.tmpdir());
+
+    // launch the application
```
nikhilkh: Nit: I liked the placement of comments earlier - though - given we are emitting console log messages with the same text  - these comments are somewhat redundant.

bin/templates/cordova/lib/retry.js
```diff
@@ -0,0 +1,63 @@
...
+"use strict";
+
+/*
+ * Retry a promise-returning function attemts_left times, propagating its
+ * results on success or throwing its error on a failed final attempt.
+ *
+ * Takes a number of attempts, a promise-returning function, and all its
+ * arguments (as trailing arguments).
+ */
+module.exports.retryPromise = function (attemts_left, promiseFunction) {
```
nikhilkh: Consider writing unit tests for this one as it is fairly easy and isolated to write for this function.

bin/templates/cordova/lib/retry.js
```diff
@@ -0,0 +1,63 @@
...
+        // on rejection either retry, or throw the error
+        function onRejected(error) {
+
+            attemts_left -= 1;
+
+            if (attemts_left < 1) {
+                throw error;
+            }
+
+            console.log("call failed; retrying " + attemts_left + " more time(s)");
```
nikhilkh: Nit. use full stop instead of semicolon. This is not code but message to users and start the message with a capital letter.

nikhilkh:
Mostly looks good - some stylistic comments

bin/templates/cordova/lib/emulator.js
dblotsky: If the console logs get removed (for example, to make the function less verbose), the comments will remain.

bin/templates/cordova/lib/retry.js
dblotsky: Good idea. I created a JIRA task for this.

bin/templates/cordova/lib/retry.js
dblotsky: I'm conflicted on whether this log message should exist. It's there now because I chose to be more verbose rather than less verbose, but I'm not sure that it belongs here in general. If this function gets used in other places, these messages will quickly get out of hand. What's Cordova's way to print verbose logs only when asked?

bin/templates/cordova/lib/exec.js
dblotsky: Previously, a hard-coded `maxBuffer` was always passed to `child_process.exec`. Now, it's passed as a default when it's not specified in a given `options.maxBuffer`. Otherwise, the given value is respected.

bin/templates/cordova/lib/exec.js
```diff
@@ -19,16 +19,42 @@
...
+    }
+
+    // override cwd to preserve old opt_cwd behavior
+    options.cwd = opt_cwd;
+
+    // set maxBuffer
+    if (options.maxBuffer === undefined) {
```
nikhilkh: Comparing to undefined is a bad practice: http://stackoverflow.com/questions/27509/detecting-an-undefined-object-property

bin/templates/cordova/lib/emulator.js
nikhilkh: The comment should be added when the console log gets removed in that case. Following the DRY principle - Don't Repeat Yourself.

bin/templates/cordova/lib/exec.js
nikhilkh: Makes sense. I see that now.

bin/templates/cordova/lib/retry.js
nikhilkh: There is no unified way to do verbose logging - every platform & cordova-lib does it a little bit differently. I have not seen any code in Android platform doing it altogether. If you decide to keep these messages, consider reformatting.

bin/templates/cordova/lib/emulator.js
dblotsky: Don't Repeat Yourself applies to repeated code or logic, not to comments. This comment improves readability because you can clearly see what every `then` block does by looking only at the comment above it, and it is therefore useful.

bin/templates/cordova/lib/retry.js
dblotsky: Added capitals and periods.

bin/templates/cordova/lib/exec.js
dblotsky: Now using `typeof options.maxBuffer === "undefined"` instead.

bin/templates/cordova/lib/emulator.js
nikhilkh: It applies two any n things you need to keep in sync - be it code/comments/documentation etc. Anyway, this duplication is too trivial to make any measurable difference and it's best to move forward.

PR #181, opened 2015-06-09
CB-9135 fix the vulnerability bug

NoLongerLazyDhl:
This is fixed in 4.0.x, but not in 3.7.x. The property "LoadUrlTimeoutValue" still gets its value from intent not from configuration file. This commit will fix this bug.


nikhilkh:
@infil00p Can you please help review?

infil00p:
This pull request is for 3.7.x, which is no longer supported.  I'm going to be closing this PR.

PR #183, opened 2015-06-18
Update CordovaActivity.java

poocart:
This causes ionic framework to pause between splashscreen and app and this shows white screen and my fix solves it.


infil00p:
This is an old PR that removes functionality.  It also no longer merges correctly.  I'm going to close this PR.

PR #186, opened 2015-06-26
CB-9255: make getUriType case insensitive

cjpearson:


omefire:
the code looks good. However, I'm curious : is there any specific use case this breaks ? or is it more like cleanup ?

cjpearson:
The file transfer plugin uses this method to ensure that a url uses http or https. We would get INVALID_URL_ERR whenever a user entered a url with a capital h in Http. We've worked around this by lowercasing the scheme automatically, but this seems like something that could cause issues again.

purplecabbage:
Makes sense to me.

omefire:
@cjpearson, it does make sense. thanks for the explanation.

vladimir-kotikov:
LGTM. The build seems to fail due to some Travis internal issues with Android SDK.

PR #188, opened 2015-06-30
CB-9172 Improved emulator deploy stability

alsorokin:
This is a fix for https://issues.apache.org/jira/browse/CB-9172
- Use UUID to distinguish between launched emulators
- Wait for android.process.acore instead of init.svc.bootanim on emulator boot
- Increased retry timeout when installing app to the emulator
- If there is already a started/starting emulator, wait for it's boot instead of trying to deploy to it right away

### DIFF
--- a/bin/templates/cordova/lib/emulator.js
+++ b/bin/templates/cordova/lib/emulator.js
@@ -33,7 +33,8 @@ var child_process = require('child_process');
 
 // constants
 var ONE_SECOND              = 1000; // in milliseconds
-var INSTALL_COMMAND_TIMEOUT = 120 * ONE_SECOND; // in milliseconds
+var ONE_MINUTE              = 60 * ONE_SECOND; // in milliseconds
+var INSTALL_COMMAND_TIMEOUT = 5 * ONE_MINUTE; // in milliseconds
 var NUM_INSTALL_RETRIES     = 3;
 var EXEC_KILL_SIGNAL        = 'SIGKILL';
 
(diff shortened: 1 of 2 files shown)


bin/templates/cordova/lib/emulator.js
```diff
@@ -238,13 +242,13 @@ module.exports.wait_for_emulator = function(num_running) {
  */
```
nikhilkh: This comment needs to be updated.

bin/templates/cordova/lib/emulator.js
```diff
@@ -238,13 +242,13 @@ module.exports.wait_for_emulator = function(num_running) {
...
-    return exec('adb -s ' + emulator_id + ' shell getprop init.svc.bootanim', os.tmpdir())
+    return exec('adb -s ' + emulator_id + ' shell ps', os.tmpdir())
     .then(function(output) {
-        if (output.match(/stopped/)) {
+        if (output.match(/android\.process\.acore/)) {
             return;
         } else {
             process.stdout.write('.');
-            return Q.delay(3000).then(function() {
+            return Q.delay(7000).then(function() {
```
nikhilkh: Wonder if doing this every 3 seconds might still be a good idea - why decrease this polling frequency?

alsorokin: From my observations, `shell ps` (getting the list of all processes) is more time consuming than `shell getprop` (getting one property) so I increased the polling time.
This is only my feeling, we can safely return it back to 3s I think.

bin/templates/cordova/lib/emulator.js
alsorokin: Done

nikhilkh:
@dblotsky Can you please help review and merge?

dblotsky:
LGTM. This has been up on the staging CI for a while, and I haven't been seeing issues. Thus, letting it lose on the real CI.

dblotsky:
Merged. Feel free to delete the branch.

PR #191, opened 2015-07-03
CB-9286 Fixes build failure when ANDROID_HOME is not set.

vladimir-kotikov:
This fixes [CB-9286](https://issues.apache.org/jira/browse/CB-9286) which is a regression from 4bf705a.


nikhilkh:
LGTM

PR #192, opened 2015-07-06
CB-9298: Preferences are now passed to the Cordova Interface

infil00p:
I'm doing a pull request instead of just slamming this in because this is actually a pretty important change that will impact people.  I can't figure out another way to do this, and I think this will be the least intrusive.  Please look it over and let me know.


infil00p:
Didn't read the plugin info.  Closing.

PR #195, opened 2015-07-08
CB-9334 loop on CordovaPlugins before returning remapped Uri

spirann:
This is a fix for  https://issues.apache.org/jira/browse/CB-9334

Because sometimes you need more than one plugin to handle your Uri's.
Looping on all plugins before returning the remapped Uri.

### DIFF
--- a/framework/src/org/apache/cordova/PluginManager.java
+++ b/framework/src/org/apache/cordova/PluginManager.java
@@ -374,17 +374,16 @@ public class PluginManager {
     }
 
     Uri remapUri(Uri uri) {
-        Uri remap = null;
+        Uri remap = uri;
         for (CordovaPlugin plugin : this.pluginMap.values()) {
             if (plugin != null) {
-                Uri ret = plugin.remapUri(uri);
+                Uri ret = plugin.remapUri(remap);
                 if (ret != null) {
                     remap = ret;
-                    uri = ret;
                 }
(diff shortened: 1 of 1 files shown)


framework/src/org/apache/cordova/PluginManager.java
```diff
@@ -381,15 +374,16 @@ public void onReset() {
...
-                Uri ret = plugin.remapUri(uri);
+                Uri ret = plugin.remapUri(remap);
                 if (ret != null) {
-                    return ret;
+                    remap = ret;
                 }
             }
         }
-        return null;
+        return uri.equals(remap) ? null : remap;
```
omefire: I suppose this returns null if no plugins ends up remapping the uri. However, a comment as to why this is being done would be great !

spirann: It was the default behavior to return null if no plugins ends up remapping the uri. The only change is that it also return null if a plugin remap the uri to the same uri, because there's no need to act like there's a change when there's not (some plugins like JsHybuggerPlugin for example always remap every uri, to the same when no action is needed)

infil00p:
This is a pull request into a deprecated branch of Cordova.  I'm going to close the Pull Request.  Can you see if this issue still exists in the latest, and if so, open a new PR?

PR #196, opened 2015-07-09
window.location="http://remoteaddress" triggers a popup to choose application

mreis1:
As you can see in the below in later versions of cordova using window.location in android triggers the popup to chose the app. This would be ok if there was alternatives to change the main webview location.

http://stackoverflow.com/questions/31299338/how-to-use-window-location-http-remote-url-in-android-cordova-v3-6-4

PS: Sorry i think i messed up in the issue location


infil00p:
This commit is a complete mess.  I'm closing this!

PR #200, opened 2015-07-17
CB-9382 [Android] Fix KeepRunning setting when Plugin activity is showed

sgrebnov:
https://issues.apache.org/jira/browse/CB-9382

### DIFF
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -235,7 +235,10 @@ public class CordovaActivity extends Activity {
         LOG.d(TAG, "Paused the activity.");
 
         if (this.appView != null) {
-            this.appView.handlePause(this.keepRunning);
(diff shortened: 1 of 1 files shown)


vladimir-kotikov:
LGTM

omefire:
Why don't we set keepRunning to true here ? 
We seem to want to set keepRunning to false while still allowing things to run in the background under certain conditions. 
I'm wondering what's preventing us from just setting keepRunning to true.

PDACN:
Hi,
I have the same problem now,
when I start NewActivity (using this.cordova.startActivityForResult) from CordovaPlugin, 
after X minutes, usually 10/15 mins, CordovaActivity goes in pause,
if I send CordovaCallback from CordovaPlugin (using this.cordova.getThreadPool().execute(() -> cordovaCB.sendPluginResult(pluginResult))) nothing happens untill I close my NewActivity and return on CordovaActivity.

PR #202, opened 2015-07-22
CB-9185 Fixed an issue when unsigned apks couldn't be found

alsorokin:
https://issues.apache.org/jira/browse/CB-9185


vladimir-kotikov:
@alsorokin , could you please delete MSOT branch if not needed anymore

PR #208, opened 2015-08-20
CB-9519 Fix the engine destruction bug

NoLongerLazyDhl:


DILEEP-YADAV:
http://www.plugbunch.com/

nikhilkh:
@vladimir-kotikov Can you please review - as we look to do a new release - does this change need to be included?

vladimir-kotikov:
@nikhilkh, i'm not 100% sure - according to comments in the code it might sense to have this checked in, but looking at where this code was introduced [`087ec11`#L625](https://github.com/apache/cordova-android/commit/087ec11e6abc61acde4e2860f8ee363f3daba578#diff-08c0ff867a9ff2aaaa99017464f55684L625), it was always working this way - we never wait for `about:blank` being loaded.

@NoLongerLazyDhl, could you please add some details about the problem, which this change solves? Maybe repro steps?

NoLongerLazyDhl:
@vladimir-kotikov When I want to receive the "exit" message in my activity which extends CordovaActivity to do something, it make that I'm not able to receive the message. That's it solves.

DILEEP-YADAV:
ok

infil00p:
Can this be added to master and not to 4.0.x?

infil00p:
Can you please close this pull request and reformat it to be against master? Thanks!

PR #209, opened 2015-08-26
CB-9557 Fixes apk install failure when switching from debug to release build

vladimir-kotikov:
This fixes [CB-9557](https://issues.apache.org/jira/browse/CB-9557) and probably [CB-8834](https://issues.apache.org/jira/browse/CB-8834)

I've also caught the issue, described on SO (http://stackoverflow.com/questions/26794862/failure-install-failed-update-incompatible) which is resolved by this PR as well.

### DIFF
--- a/bin/templates/cordova/lib/device.js
+++ b/bin/templates/cordova/lib/device.js
@@ -96,10 +96,17 @@ module.exports.install = function(target, buildResults) {
     }).then(function(resolvedTarget) {
         var apk_path = build.findBestApkForArchitecture(buildResults, resolvedTarget.arch);
         var launchName = appinfo.getActivityName();
+        var pkgName = appinfo.getPackageName();
         console.log('Using apk: ' + apk_path);
-        console.log('Installing app on device...');
-        var cmd = 'adb -s ' + resolvedTarget.target + ' install -r "' + apk_path + '"';
-        return exec(cmd, os.tmpdir())
+        console.log('Uninstalling ' + pkgName + ' from device...');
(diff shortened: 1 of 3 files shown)


bin/templates/cordova/lib/device.js
```diff
@@ -96,10 +96,15 @@ module.exports.install = function(target, buildResults) {
     }).then(function(resolvedTarget) {
         var apk_path = build.findBestApkForArchitecture(buildResults, resolvedTarget.arch);
         var launchName = appinfo.getActivityName();
+        var pkgName = appinfo.getPackageName();
         console.log('Using apk: ' + apk_path);
-        console.log('Installing app on device...');
-        var cmd = 'adb -s ' + resolvedTarget.target + ' install -r "' + apk_path + '"';
-        return exec(cmd, os.tmpdir())
+        console.log('Uninstalling ' + pkgName + ' from device...');
+        return exec('adb -s ' + resolvedTarget.target + ' uninstall ' + pkgName, os.tmpdir())
```
nikhilkh: Wouldn't uninstalling always cause a failure - especially on the first run?

vladimir-kotikov: No. Even if `adb uninstall` fails to uninstall an app, it exits with error code 0, and thus `exec` promise is always resolved in this case.

nikhilkh: Ah - A comment would help clarify that.

vladimir-kotikov: Done.

nikhilkh:
LGTM

PR #210, opened 2015-08-29
CB-9577 Quoted specific device name for ADB

vksee:
This fixes [CB-9577](https://issues.apache.org/jira/browse/CB-9577).


nikhilkh:
@vladimir-kotikov Looks like your refactor made this PR irrevelant. Does your new code handle this correctly?

vladimir-kotikov:
@nikhilkh, yes, `superspawn` module handles arguments with spaces correctly. I can't test it on real device, but tested using some mocks, and it works.

PR #211, opened 2015-08-31
The extra digit need to be added to the end of the version code

fujunwei:
The version code of default config is generated by the environment variable
and the value from AndroidManifest.xml.
The test case is using the command line "cordova build -- --versionCode=100".


fujunwei:
@huningxin @infil00p PTAL, thanks.

huningxin:
LGTM. Thanks!

fujunwei:
@agrieve

PR #214, opened 2015-09-15
Network Error

achercasky:
Hi! I'm trying to load an url from the userwebview.class. The url is https://google.com (very simple hehe). I already add the <plugin name="cordova-plugin-whitelist" version="1" /> and <allow-navigation href="https://google.com/*" />.

When I'm trying to load the url, the dialog appear with an error. "Application Error. There was a network error".

I already have all the permission for access network and internet.


purplecabbage:
This is a pull request, not the way to ask for help with a specific issue you are having.
Please ask on StackOverflow, or join the slack.cordova.io

PR #215, opened 2015-09-23
CB-9608 cordova-android no longer builds on Node 0.10 or below

daserge:
Replaced `path.isAbsolute` usage with `path.resolve`.

[Jira issue](https://issues.apache.org/jira/browse/CB-9608)

PR #217, opened 2015-09-23
CB-9608 cordova-android no longer builds on Node 0.10 or below

daserge:
Replaced `path.isAbsolute` usage with `path.resolve`.

[Jira issue](https://issues.apache.org/jira/browse/CB-9608)


daserge:
This should work on OSX now as well (`path.join('.', '/some/abs/path.tofile')` was turning path to relative in #215 ).

vladimir-kotikov:
LGTM

PR #218, opened 2015-09-24
CB-9696: Add possibility to use non-immersive fullscreen in all Andro…

ttiurani:
…id versions

When setting the new preference "fullscreen-immersive" to false like this:

<preference name="fullscreen" value="true" />
<preference name="fullscreen-immersive" value="false" />

Cordova reverts to old fullscreen behaviour also on newer Android versions. This fixes issue
CB-9696.


ttiurani:
Any chance to pull this very minor change? Is there something more I could do? The new immersive mode is substantially different from the old, and there is no practical way to migrate my code to support immersive mode.

infil00p:
If you still want this feature, can you change it so it merges cleanly, and works with the other code that uses non-immersive as a fallback? Thanks!

PR #216, opened 2015-09-23
CB-9080 Cordova CLI run for Android versions 4.1.1 and lower throws e…

daserge:
…rror

[Jira issue](https://issues.apache.org/jira/browse/CB-9080)

PR #219, opened 2015-09-24
Fixed CB-9697 Cordova emulate for Android versions 4.1.1 does not work

mleoking:
After running 'cordova emulate android' the apk is not installed & shown on the emulator despite the log shows the app is successfully launched.

This issue is related to https://issues.apache.org/jira/browse/CB-9080 and https://issues.apache.org/jira/browse/CB-8912. These two previous issues fix the problem in 'device.js' and error in 'emulator.js' is still not corrected.

The `emulator.js` file try to run this command:

exec, 'adb -s ' + target.target + ' install -r -d "' + apk_path + '"', os.tmpdir(), execOptions

The '-d' option is not supported by Android 4.1.1

Remove the '-d' option will fix this issue.

### DIFF
--- a/bin/templates/cordova/lib/emulator.js
+++ b/bin/templates/cordova/lib/emulator.js
@@ -336,7 +336,7 @@ module.exports.install = function(givenTarget, buildResults) {
 
         var retriedInstall = retry.retryPromise(
             NUM_INSTALL_RETRIES,
-            exec, 'adb -s ' + target.target + ' install -r -d "' + apk_path + '"', os.tmpdir(), execOptions
+            exec, 'adb -s ' + target.target + ' install -r "' + apk_path + '"', os.tmpdir(), execOptions
         );
 
         return retriedInstall.then(function (output) {


daserge:
@mleoking thanks!
It should be fixed in `device.js` as well.
I believe this is a duplicate of #216 though.

mleoking:
@daserge If you visit https://github.com/apache/cordova-android/blob/4.1.x/bin/templates/cordova/lib/device.js in the `4.1.x branch`, you will see that `device.js` is already fixed (-d is removed). But `emulator.js` is not. So, this pull just fix `emulator.js` in the `4.1.x branch`.

I am not sure whether the `-d` option should be removed for more branches. It depends on whether other versions of android system support this option.

It would be really nice if someone can do a systematic test and fix them all. :)

daserge:
@mleoking Thanks for the explanation!
I think `-d` was also being removed in master before [this commit](https://github.com/apache/cordova-android/commit/bf57aa1df061b3fcd429bb8d673de02bcb26f872#diff-257a87acd541c25468767e36993508b9L101), so I'm not sure that this fix should only go to the `4.1.x` branch.
What does `-d` option do on >4.1.1 devices? 
I could not find anything more than 
`-d: Allow version code downgrade.`

daserge:
Is `-d` just about reinstalling an app with version less than the installed one has?

mleoking:
@daserge Thanks for letting me know about [the commit](https://github.com/apache/cordova-android/commit/bf57aa1df061b3fcd429bb8d673de02bcb26f872#diff-257a87acd541c25468767e36993508b9L101). It does not remove the `-d` option from `emulator.js` though.

I am not sure how `-d` is useful for the app development. If it is of no use, we can remove it from all branches?

Can someone explain this?

@infil00p @mwbrooks

nikhilkh:
We do not plan to make a release to Android 4.1.x. I do not see much point in merging this.

mleoking:
@nikhilkh I created this patch because the ionic framework by default download this branch (Android 4.1.x) of cordova when adding the android platform. 

As ionic is widely popular, I suppose this bug would affect many people. Personally, every time, I create a new ionic project and add the android platform, I need to manually fix this file. 

So, if you can please fix it.

PR #220, opened 2015-09-24
Current Cordova-Android 5.0 changes.

infil00p:
These are the changes to Cordova-Android that make it a 5.0 change.  This is a very minor major release, since we're literally just adding the ability for plugins to handle permissions, and adding a special hook for Geolocation so that browser geolocation can work.


framework/project.properties
```diff
@@ -10,7 +10,7 @@
 # Indicates whether an apk should be generated for each density.
 split.density=false
 # Project target.
-target=android-22
+target=android-23
```
nikhilkh: This will break http://ci.cordova.io. @alsorokin Can you please update the Windows slave to install android 23. @dblotsky Can you do the same for the iOS slave.

framework/src/org/apache/cordova/CordovaInterface.java
```diff
@@ -69,4 +69,7 @@ Licensed to the Apache Software Foundation (ASF) under one
      * Returns a shared thread pool that can be used for background tasks.
      */
     public ExecutorService getThreadPool();
+
+    public void requestPermission(CordovaPlugin plugin, int requestCode, String permission);
```
nikhilkh: JavaDoc would be great here - since this is public API. Eventually we should figure out how to publish the public API docs online.

framework/src/org/apache/cordova/CordovaInterfaceImpl.java
```diff
@@ -161,4 +164,35 @@ public ActivityResultHolder(int requestCode, int resultCode, Intent intent) {
...
+        permissionResultCallback = plugin;
+        String[] permissions = new String [1];
+        permissions[0] = permission;
+        getActivity().requestPermissions(permissions, requestCode);
+    }
+
+    public void requestPermissions(CordovaPlugin plugin, int requestCode)
+    {
+        permissionResultCallback = plugin;
+        String[] permissions = plugin.getPermissionRequest();
```
nikhilkh: It's kinda odd to have the permission array not being passed along instead we invoking a callback on plugin to get the permissions. Passing the permissions array would be a simpler design.

infil00p: This is actually in response to Andrew's feedback that we do this.  Most of the plugins don't actually have a set list of permissions, and most only have one permission.  This becomes far more clear when you read the plugins that have permissions.

framework/src/org/apache/cordova/CordovaPlugin.java
```diff
@@ -359,4 +361,38 @@ public boolean onReceivedClientCertRequest(CordovaWebView view, ICordovaClientCe
...
+     */
+
+    public String[] getPermissionRequest() {
+        return permissions;
+    }
+
+    public boolean hasPermisssion() {
+        for(String p : permissions)
+        {
+            if(PackageManager.PERMISSION_DENIED == cordova.getActivity().checkSelfPermission(p))
```
nikhilkh: Shouldn't this API call be guarded by a check for Android version M?

framework/src/org/apache/cordova/CordovaPlugin.java
```diff
@@ -359,4 +361,38 @@ public boolean onReceivedClientCertRequest(CordovaWebView view, ICordovaClientCe
...
+            }
+        }
+
+        return true;
+    }
+
+    /**
+     * Called by the system when the user grants permissions
+     *
+     * @param requestCode
```
nikhilkh: Another public method in a base class - would be great to fill the Javadoc comments.

framework/src/org/apache/cordova/CordovaPlugin.java
infil00p: Perhaps it should, although I haven't ran into any issues with this so far.

nikhilkh: Is it because geloc plugin overrides this method?

infil00p: That's probably the case.  I'm almost tempted to remove the code from this method entirely, since this really exists only for a Geolocation plugin.  The geolocation plugin is a special case, and no other plugins work this way.

nikhilkh: +1 I do find this confusing. I don't also understand why geolocation plugin has to be different. I understand some of this has to be because of Chrome webview handling here: https://github.com/apache/cordova-android/pull/220/files#diff-eb00315f6b4fb38aba0833096a86bb8aR190

The plublic

infil00p: The geolocation plugin ONLY handles the permissions, and since we require both the coarse location and the fine location to be requested, we send them both to the activity.  There isn't this sort of behaviour in any other plugin that we currently maintain.  However, it's possible that other plugins will want to send a bunch of permission as a batch, so we facilitate that here, which is why it's public.

That said, we don't do this for the media plugin.  I chose the one permission at a time approach because it's easier to control a plugin's callbackContext that way.

nikhilkh: I understand the need for multiple permissions with 1 callbackContext. An array of permisisons would lead to a simpler design:

``public void requestPermissions(CordovaPlugin plugin, int requestCode, string[] permissions);```

However, this code will need to be changed to either have private knowledge of what permissions the geoloc plugin needs or ask it somehow:
https://github.com/apache/cordova-android/pull/220/files#diff-eb00315f6b4fb38aba0833096a86bb8aR190

I think then we should be able to remove the `CordovaPlugin` base class methods and helpers for permissions.

infil00p: No, we still need the base class methods and helpers defined so that if we need to, we can pull up any plugin and access these methods without having to do some reflection to make sure that they still exist.

framework/project.properties
alsorokin: done for windows slave

DILEEP-YADAV:
http://www.plugbunch.com/

infil00p:
Going to close and issue a new pull request based on the prior pull request.

PR #223, opened 2015-09-30
Cordova-Android 5.0/Marshmallow Updates - Take 2 

infil00p:
This has some changes based on the feedback from Pull Request #220


framework/src/org/apache/cordova/CordovaInterface.java
```diff
@@ -69,4 +69,20 @@ Licensed to the Apache Software Foundation (ASF) under one
...
+
+    /**
+     * Sends a permission request to the activity for one permission.
+     */
+    public void requestPermission(CordovaPlugin plugin, int requestCode, String permission);
+
+    /**
+     * Sends a permission request to the activity for a group of permissions
+     */
+    public void requestPermissions(CordovaPlugin plugin, int requestCode, String [] permissions);
```
nikhilkh: Thanks for making the change! :+1:

framework/src/org/apache/cordova/CordovaPlugin.java
```diff
@@ -40,6 +42,7 @@ Licensed to the Apache Software Foundation (ASF) under one
     public CordovaInterface cordova;
     protected CordovaPreferences preferences;
     private String serviceName;
+    protected String [] permissions;
```
nikhilkh: I feel this is somewhat confusing. What is the derived plugin supposed to with this field? I do not see any of the core plugins doing anything with this field. Did I miss something?

nikhilkh: Each of the core plugins are creating their own version of this field either on the instance or `static`. This can be quite confusing. How would you explain to a plugin developer - what they are expected to do with this field.

framework/src/org/apache/cordova/CordovaPlugin.java
```diff
@@ -359,4 +361,38 @@ public boolean onReceivedClientCertRequest(CordovaWebView view, ICordovaClientCe
...
+
+    /**
+     * Called by the Plugin Manager when we need to actually request permissions
+     *
+     * @param requestCode   Passed to the activity to track the request
+     *
+     * @return              Returns the permission that was stored in the plugin
+     */
+
+    public void requestPermissions(int requestCode) {
```
nikhilkh: The design feels a bit hacky - designed around how geolocation plugin has to do this. In fact, your changes for camera plugin for Masrhmallow don't override this method and the `hasPermission` method. What's the guidance to a cordova plugin should they care to override this or not?

infil00p:
I rebased this into the repository, closing.

chaffeqa:
:+1:  thanks a ton @infil00p !

PR #226, opened 2015-10-12
CB-9782 Implements PlatformApi contract for Android platform

vladimir-kotikov:
This is an implementation for [CB-9782](https://issues.apache.org/jira/browse/CB-9782) - PlatformApi compatible api for android platform, which could be used by cordova and other tools as a separate NodeJS module.

The main entry point for new API is [/bin/templates/cordova/Api.js](https://github.com/MSOpenTech/cordova-android/blob/2023e87c93a7d2ea665b93e02a9034e3927c27c2/bin/templates/cordova/Api.js), which exposes all the methods, defined in PlatformApi spec at https://github.com/cordova/cordova-discuss/pull/12

Since the PR contains a lot of changes, here is the main points of interest:

1. [Api.js](https://github.com/MSOpenTech/cordova-android/blob/2023e87c93a7d2ea665b93e02a9034e3927c27c2/bin/templates/cordova/Api.js) which looks similar to PlatformApi polyfill implementation, but contains only android-specific code.
2. `build`/`run`/`check_reqs` scripts are reworked, so they are used as corresponding `Api` methods implementations (bound to `Api` instance at runtime)
3. 'prepare' logic from cordova-lib/metadata/android_parser copied into [`prepare`](https://github.com/MSOpenTech/cordova-android/blob/2023e87c93a7d2ea665b93e02a9034e3927c27c2/bin/templates/cordova/lib/prepare.js) module.
3. `build` script refactored and split into number of 'builders': [`GradleBuilder`](https://github.com/MSOpenTech/cordova-android/blob/2023e87c93a7d2ea665b93e02a9034e3927c27c2/bin/templates/cordova/lib/builders/GradleBuilder.js), [`AntBuilder`](https://github.com/MSOpenTech/cordova-android/blob/2023e87c93a7d2ea665b93e02a9034e3927c27c2/bin/templates/cordova/lib/builders/AntBuilder.js)
4. All the logic that deals with `Androidmanifest.xml` factored into [`AndroidManifest` class](https://github.com/MSOpenTech/cordova-android/blob/2023e87c93a7d2ea665b93e02a9034e3927c27c2/bin/templates/cordova/lib/AndroidManifest.js)
5. `exec` and `spawn` modules are removed in favor of `cordova-common`

This PR requires https://github.com/apache/cordova-lib/pull/320 to be checked first.
There is also PR to `cordova-coho` to set-up npm link to cordova-common: https://github.com/apache/cordova-coho/pull/96


infil00p:
Why is all of element-tree and the other npm dependencies checked in?  That seems very odd, and makes it harder to figure out what's going on.

bin/templates/cordova/Api.js
```diff
@@ -0,0 +1,506 @@
...
+var GENERIC_EVENTS = new (require('events').EventEmitter)()
+    .on('verbose', function (message) {
+        if (process.argv.indexOf('-d') >= 0 || process.argv.indexOf('--verbose') >= 0)
+            console.log(message);
+    })
+    .on('log', console.log)
+    .on('warn', console.warn)
+    .on('error', function (error) {
+        if (process.argv.indexOf('-d') >= 0 || process.argv.indexOf('--verbose') >= 0) {
+            console.error((error && error.message) || error);
```
nikhilkh: Should this be flipped? We should `stack` for verbose output and not otherwise? Also, does the message not show when we show stack? If I understand correctly, this is only for legacy scenarios - correct? Otherwise, the cli should eventually be doing the logging - correct?

nikhilkh: Or cordova-lib needs to do the logging not to break existing users of cordova-lib.

vladimir-kotikov: > Should this be flipped? We should stack for verbose output and not otherwise?

Yes, definitely. My bad.

> Also, does the message not show when we show stack?

The message is shown along with stack, yes. Just to clarify, the idea here is that until you don't have `-d` switch, you won't get full stacks, because they are only interesting for those, who working with cordova-lib/platform internals. Common users should see only meaningful error message , that explains what went wrong. This logic is borrowed [from CLI ](https://github.com/apache/cordova-cli/blob/c5acc3dcbea7ac736f778e58cde23a38b467572a/src/cli.js#L130)

>  If I understand correctly, this is only for legacy scenarios - correct? Otherwise, the cli should eventually be doing the logging - correct?

This code is required for cases, when caller of PlatformApi doesn't provide logger/eventEmitter instance. In this case we still need to log some events. In particular, this code is used in non-CLI workflow. For CLI workflow - yes, logging is already handled by CLI.

vladimir-kotikov:
@infil00p, the npm dependencies was already checked in (see https://github.com/apache/cordova-android/commit/7be1f018aa2729efaa569e422bee893686ad34b8 and https://github.com/apache/cordova-android/commit/fd6a1e5ed00baba902279dbc52cb9639b0e78eb3) as a part of [CB-3542](https://issues.apache.org/jira/browse/CB-3542). I just added two more dependencies (elementtree and nopt) and moved `node_modules` directory to repo root.

bin/templates/cordova/Api.js
vladimir-kotikov: @nikhilkh, reworked this logic completely in https://github.com/MSOpenTech/cordova-android/commit/6d9d4f4beb7a437f38391f10017669953fadede4

nikhilkh:
@alsorokin Can you please buddy test this? This is a significant change.

alsorokin:
@nikhilkh I have tested and reported found issues to @vladimir-kotikov 
Everything that I found is fixed now, LGTM

bin/create
```diff
@@ -1,49 +1,56 @@
...
+var argv = require('nopt')({
+    'help' : Boolean,
+    'cli' : Boolean,
+    'shared' : Boolean,
+    'link' : Boolean,
+    'activity-name' : [String, undefined]
+});
+
+if (argv.help || argv.argv.remain.length === 0) {
+    console.log('Usage: ' + path.relative(process.cwd(), path.join(__dirname, 'create')) + ' <path_to_new_project> <package_name> <project_name> [<template_path>] [--activity-name <activity_name>] [--link]');
```
stevengill: Would it make sense to use cordova logging instead of console.log here?

vladimir-kotikov: You mean, use `events.emit`? I think no, because this code should not be `require`d from anywhere, it is a regular executable script. This means that there will be no subscribers for these events.

stevengill: I did mean `events.emit`. Sounds good.

nikhilkh:
LGTM. Let's get this in! Is there a fix for the travis CI issue.

vladimir-kotikov:
@nikhilkh, those failures fixed in https://github.com/apache/cordova-android/commit/789c505a88a6bd2c7c5be596908c85681e457132 and https://github.com/MSOpenTech/cordova-android/commit/9c2d3817cf39c20411ddad007a663e3ca161200b

PR #227, opened 2015-10-15
Move CordovaWebView to CordovaWebInterface

macdonst:
As discussed on at the face to face we should rename this interface.


infil00p:
I'm pretty sure we were told we couldn't do this for some reason.  We should revisit it on the list.

PR #230, opened 2015-10-21
CB-9835 Downgrade `properties-parser` to prevent failures in Node < 4.x

vladimir-kotikov:
JIRA: https://issues.apache.org/jira/browse/CB-9835

This PR, once merged, should fix tests failures on our CI: https://ci.appveyor.com/project/Humbedooh/cordova-lib/build/1.0.1068#L462


vladimir-kotikov:
@TimBarham. @nikhilkh, @stevengill, please review

TimBarham:
Looks fine to me, as long as we're sure the older version doesn't lose any functionality we need.

vladimir-kotikov:
It does not. We use 0.2.3 in cordova-lib for `android_parser`: https://github.com/apache/cordova-lib/blob/master/cordova-lib/package.json#L32

nikhilkh:
Is there an appropriate e2e/unit test case that can be added to cordova-android to detect this? It's unfortunate cordova-lib tests are the only way we found it. Getting this in soon is a good idea - we should take a follow up item to add some tests here - @alsorokin

TimBarham:
:+1: (to getting this change in, and to then creating tests to guard against this going forwards).

TimBarham:
Just an FYI - The node 4.x.x dependency in `node-parser` has just been fixed and the fix released, so we can move to `0.3.1` safely if ever there is a reason to do so.

PR #233, opened 2015-10-27
CB-9844 Remove old .java after renaming activity

vladimir-kotikov:
This is a port of https://github.com/apache/cordova-lib/pull/328 to cordova-android. Need to be done in two places since the `prepare` logic was moved to platform.

More details on JIRA: [CB-9844](https://issues.apache.org/jira/browse/CB-9844) and [CB-9736](https://issues.apache.org/jira/browse/CB-9736)


vladimir-kotikov:
@gorkem, please take a look

gorkem:
LGTM

PR #237, opened 2015-10-29
CB-9909 Shouldn't escape spaces in paths on Windows.

TimBarham:
Instead on windows just wrap in quotes. Note that this value is only never executed directly - just displayed to the user. But this way they can copy and paste successfully on Windows.


TimBarham:
@stevengill - can you take a look at this? Thanks!

stevengill:
LGTM

omefire:
LGTM!

PR #238, opened 2015-11-06
CB-9831 CB-9835 CB-9932 Added e2e tests

alsorokin:
https://issues.apache.org/jira/browse/CB-9831
https://issues.apache.org/jira/browse/CB-9835
https://issues.apache.org/jira/browse/CB-9932


spec/e2e/update.spec.js
```diff
@@ -0,0 +1,96 @@
...
+            });
+        });
+    });
+}
+
+describe('preparing fixtures', function () {
+
+    it('cloning old platform', function (done) {
+        shell.rm('-rf', platformOld.path);
+        shell.exec('git clone ' + PLATFORM_GIT_URL + ' ' + platformOld.path +
```
vladimir-kotikov: I would use `--branch` option to checkout specific branch instead od doing this manually. Also `--depth` option decreased test time dramatically. This might not work in future, when we'll test multiple upgrade/downgrade variants (though `--no-single-branch` might help), but for now i'd prefer to have `--depth 1` enabled.
 
```javascript
var command = util.format('git clone %s --depth=1 --branch %s --single-branch %s',
    PLATFORM_GIT_URL, platformOld.version, platformOld.path);
```

vladimir-kotikov:
LGTM.

IMO it might be better to use promises in `projectActions` to reduce 'pyramids of doom' in [update.spec.js](https://github.com/apache/cordova-android/pull/238/files#diff-0f89b7fec99ccbdee182426eae6f10cdR34) but this is a matter of taste

PR #239, opened 2015-11-11
CB-8917: Added pending plugin callbacks to resume event payload

riknoll:
This is a redo of #236 after receiving some feedback. This relates to CB-8917 and CB-9189.

### Background
The issue at hand is that plugins can make calls to an external Activity and, if the device is low on memory, there is a chance that the CordovaActivity will get killed in the background causing the plugin to lose its state as well as its context for the callback. Activities in Android typically handle this situation by using `onSaveInstanceState()` and `onCreate()` methods to save and restore state respectively as the Activity is created and destroyed. This solution exposes that lifecycle to plugins, allowing them to save state and have it restored if necessary.

### Saving/Restoring plugin state
Two new methods are exposed to plugins that they can override to save/restore state.
```java
/**
 * Called when the Activity is being destroyed (e.g. if a plugin calls out to an external
 * Activity and the OS kills the CordovaActivity in the background). The plugin should save its
 * state in this method only if it is awaiting the result of an external Activity and needs
 * to preserve some information so as to handle that result; onRestoreStateForActivityResult()
 * will only be called if the plugin is the recipient of an Activity result
 *
 * @return  Bundle containing the state of the plugin or null if state does not need to be saved
 */
public Bundle onSaveInstanceState() {}

/**
 * Called when a plugin is the recipient of an Activity result after the CordovaActivity has
 * been destroyed. The Bundle will be the same as the one the plugin returned in
 * onSaveInstanceState()
 *
 * @param state             Bundle containing the state of the plugin
 * @param callbackContext   Replacement Context to return the plugin result to
 */
public void onRestoreStateForActivityResult(Bundle state, CallbackContext callbackContext) {}
```
The plugin is given a replacement CallbackContext as part of `onRestoreStateForActivityResult` that can accept the result the plugin would normally return and add it to the resume event payload for use in the js (see JSON below). Thus, it requires minimal modifications to existing plugins

_NOTE:_ When I mention that plugins are given the opportunity to restore state, I want to clarify that this only happens for plugins that are waiting for an external Activity result. This makes the API a little less intuitive, but otherwise we would be conflicting with the accepted behavior that plugins currently get destroyed (i.e. lose all of their state) and are selectively rebuilt whenever a new URL is loaded into the webview. If we restore state on resume, then we can end up with some awkward cases where part of the resuming involves loading a new page so the state gets lost again and so on and so forth. My thinking is that restoring the other state is better left to app developers

### Saving/Restoring js state
We already send out pause and resume events. This solution enhances these events in the case of Activity destruction by adding to them the result of any pending Plugin calls. The resume event is of the form
```
{
    action: "resume",
    pendingResult: {
        pluginServiceName: <plugin service name e.g. "Camera">,
        pluginStatus: <description of result' status (see PluginResult.java)>,
        result: <argument(s) that would have been given to the callback>
    }
}
```
It is the responsibility of the application developer to properly use these events and save their state as well as keep information about what plugin results they have pending. We should provide guidance for this in the Android documentation and plugin documentation should clearly communicate when it is necessary.

#### Benefits:

* It requires minimal updates to existing plugins (and no updates at all if the plugin doesn't use an external Activity)
* It is a general solution/pattern that plugins can follow rather than forcing them to include platform specific methods in their APIs

#### Downsides:

* The resume callback will only ever get received after the initial page loads (potential page flickering)
* The pending result part of the event object doesn't provide much context (so it puts more responsibility on the app developer to keep state so they know what they're getting)

In the core plugins, this is mostly relevant to the Camera plugin which previously would crash upon receiving the Activity result if the CordovaActivity had been killed by the OS while a picture was being taken/chosen (CB-9189). The updated Camera Plugin can be found in [this branch](https://github.com/MSOpenTech/cordova-plugin-camera/tree/save-state-plugin-only) and a (trivial) example application that uses this API + instructions for testing can be found [here](https://gist.github.com/riknoll/d0ab25583b1dc8ab22af).

### DIFF
--- a/framework/src/org/apache/cordova/CordovaInterfaceImpl.java
+++ b/framework/src/org/apache/cordova/CordovaInterfaceImpl.java
@@ -19,7 +19,6 @@
 
 package org.apache.cordova;
 
-import android.Manifest;
 import android.app.Activity;
 import android.content.Intent;
 import android.content.pm.PackageManager;
@@ -28,6 +27,7 @@ import android.os.Bundle;
 import android.util.Log;
 
 import org.json.JSONException;
+import org.json.JSONObject;
 
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
@@ -46,6 +46,8 @@ public class CordovaInterfaceImpl implements CordovaInterface {
     protected CordovaPlugin permissionResultCallback;
     protected String initCallbackService;
     protected int activityResultRequestCode;
+    protected boolean activityWasDestroyed = false;
+    protected Bundle savedPluginState;
 
     public CordovaInterfaceImpl(Activity activity) {
         this(activity, Executors.newCachedThreadPool());
@@ -95,12 +97,28 @@ public class CordovaInterfaceImpl implements CordovaInterface {
     }
 
     /**
-     * Dispatches any pending onActivityResult callbacks.
+     * Dispatches any pending onActivityResult callbacks and sends the resume event if the
+     * Activity was destroyed by the OS.
      */
     public void onCordovaInit(PluginManager pluginManager) {
         this.pluginManager = pluginManager;
         if (savedResult != null) {
             onActivityResult(savedResult.requestCode, savedResult.resultCode, savedResult.intent);
+        } else if(activityWasDestroyed) {
+            // If there was no Activity result, we still need to send out the resume event if the
+            // Activity was destroyed by the OS
+            activityWasDestroyed = false;
+
+            CoreAndroid appPlugin = (CoreAndroid) pluginManager.getPlugin(CoreAndroid.PLUGIN_NAME);
+            if(appPlugin != null) {
+                JSONObject obj = new JSONObject();
+                try {
+                    obj.put("action", "resume");
+                } catch (JSONException e) {
+                    LOG.e(TAG, "Failed to create event message", e);
+                }
+                appPlugin.sendResumeEvent(new PluginResult(PluginResult.Status.OK, obj));
+            }
         }
     }
 
@@ -115,6 +133,10 @@ public class CordovaInterfaceImpl implements CordovaInterface {
             savedResult = new ActivityResultHolder(requestCode, resultCode, intent);
             if (pluginManager != null) {
                 callback = pluginManager.getPlugin(initCallbackService);
+                if(callback != null) {
+                    callback.onRestoreStateForActivityResult(savedPluginState.getBundle(callback.getServiceName()),
+                            new ResumeCallback(callback.getServiceName(), pluginManager));
+                }
             }
         }
         activityResultCallback = null;
@@ -126,7 +148,7 @@ public class CordovaInterfaceImpl implements CordovaInterface {
             callback.onActivityResult(requestCode, resultCode, intent);
             return true;
         }
-        Log.w(TAG, "Got an activity result, but no plugin was registered to receive it" + (savedResult != null ? " yet!": "."));
+        Log.w(TAG, "Got an activity result, but no plugin was registered to receive it" + (savedResult != null ? " yet!" : "."));
         return false;
     }
 
@@ -147,6 +169,8 @@ public class CordovaInterfaceImpl implements CordovaInterface {
             String serviceName = activityResultCallback.getServiceName();
             outState.putString("callbackService", serviceName);
         }
+
+        outState.putBundle("plugin", pluginManager.onSaveInstanceState());
     }
 
     /**
@@ -154,6 +178,8 @@ public class CordovaInterfaceImpl implements CordovaInterface {
      */
     public void restoreInstanceState(Bundle savedInstanceState) {
         initCallbackService = savedInstanceState.getString("callbackService");
+        savedPluginState = savedInstanceState.getBundle("plugin");
+        activityWasDestroyed = true;
     }
 
     private static class ActivityResultHolder {
@@ -209,6 +235,4 @@ public class CordovaInterfaceImpl implements CordovaInterface {
             return true;
         }
     }
-
-
 }
--- a/framework/src/org/apache/cordova/CoreAndroid.java
+++ b/framework/src/org/apache/cordova/CoreAndroid.java
@@ -19,10 +19,6 @@
 
 package org.apache.cordova;
 
-import org.apache.cordova.CallbackContext;
-import org.apache.cordova.CordovaPlugin;
-import org.apache.cordova.LOG;
-import org.apache.cordova.PluginResult;
 import org.json.JSONArray;
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -45,6 +41,8 @@ class CoreAndroid extends CordovaPlugin {
     protected static final String TAG = "CordovaApp";
     private BroadcastReceiver telephonyReceiver;
     private CallbackContext messageChannel;
+    private PluginResult pendingResume;
+    private final Object messageChannelLock = new Object();
 
     /**
      * Send an event to be fired on the Javascript side.
@@ -112,7 +110,13 @@ class CoreAndroid extends CordovaPlugin {
                 this.exitApp();
             }
 			else if (action.equals("messageChannel")) {
-                messageChannel = callbackContext;
+                synchronized(messageChannelLock) {
+                    messageChannel = callbackContext;
+                    if (pendingResume != null) {
+                        sendEventMessage(pendingResume);
+                        pendingResume = null;
+                    }
+                }
                 return true;
             }
 
@@ -313,10 +317,13 @@ class CoreAndroid extends CordovaPlugin {
         } catch (JSONException e) {
             LOG.e(TAG, "Failed to create event message", e);
         }
-        PluginResult pluginResult = new PluginResult(PluginResult.Status.OK, obj);
-        pluginResult.setKeepCallback(true);
+        sendEventMessage(new PluginResult(PluginResult.Status.OK, obj));
+    }
+
+    private void sendEventMessage(PluginResult payload) {
+        payload.setKeepCallback(true);
         if (messageChannel != null) {
-            messageChannel.sendPluginResult(pluginResult);
+            messageChannel.sendPluginResult(payload);
         }
     }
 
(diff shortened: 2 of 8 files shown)


riknoll:
@infil00p @purplecabbage @jasongin please review. This incorporates feedback received on the first PR. Example plugin and application code are linked at bottom of the description

framework/src/org/apache/cordova/CordovaInterface.java
```diff
@@ -22,6 +22,7 @@ Licensed to the Apache Software Foundation (ASF) under one
 import android.content.Intent;
 
 import org.apache.cordova.CordovaPlugin;
+import org.json.JSONObject;
```
jasongin: It looks like there were no real changes to this file. Was this just leftover from a previous iteration?

riknoll: Yep! I'll update the PR.

framework/src/org/apache/cordova/CordovaInterfaceImpl.java
```diff
@@ -147,6 +151,8 @@ public void onSaveInstanceState(Bundle outState) {
             String serviceName = activityResultCallback.getServiceName();
             outState.putString("callbackService", serviceName);
         }
+
+        outState.putBundle("plugin", pluginManager.onSaveInstanceState());
```
jasongin: Do you need to check if pluginManager is not null? I ask because there is a check for that on line 116.

framework/src/org/apache/cordova/CoreAndroid.java
```diff
@@ -113,6 +110,10 @@ else if (action.equals("exitApp")) {
             }
 			else if (action.equals("messageChannel")) {
                 messageChannel = callbackContext;
+                if(pendingResume != null) {
```
jasongin: nit: space after 'if'

framework/src/org/apache/cordova/CordovaInterfaceImpl.java
```diff
@@ -115,6 +115,10 @@ public boolean onActivityResult(int requestCode, int resultCode, Intent intent)
             savedResult = new ActivityResultHolder(requestCode, resultCode, intent);
             if (pluginManager != null) {
                 callback = pluginManager.getPlugin(initCallbackService);
+                if(callback != null) {
+                    callback.onRestoreStateForActivityResult(savedPluginState.getBundle(callback.getServiceName()),
```
jasongin: Are you sure savedPluginState will always be non-null at this point? And that it will have the expected bundle?

framework/src/org/apache/cordova/ResumeCallback.java
```diff
@@ -0,0 +1,71 @@
...
+        super("resumecallback", null);
+        this.serviceName = serviceName;
+        this.pluginManager = pluginManager;
+    }
+
+    @Override
+    public void sendPluginResult(PluginResult pluginResult) {
+        synchronized (this) {
+            if (finished) {
+                return;
```
jasongin: Should there be a warning here, as in the base class implementation?

framework/src/org/apache/cordova/CordovaInterfaceImpl.java
riknoll: I believe so. This will only be called after the Activity was destroyed in which case `onSaveInstanceState()` will have been called and the `savedPluginState` bundle saved. I believe Activity results are only delivered after `onCreate()` is called, so the Bundle will have already been retrieved (but I can double check that). It is the responsibility of the plugin to check if the bundle they receive is null.

framework/src/org/apache/cordova/ResumeCallback.java
```diff
@@ -0,0 +1,71 @@
...
+
+            event.put("action", "resume");
+            event.put("pendingResult", pluginResultObject);
+        } catch (JSONException e) {
+            LOG.e(TAG, "Unable to create resume object for Activity Result");
+        }
+
+        PluginResult eventResult = new PluginResult(PluginResult.Status.OK, event);
+
+        List<PluginResult> result = new ArrayList<PluginResult>();
```
jasongin: I'm confused about why you're making a list of 2 PluginResults and wrapping them in another PluginResult below. Why wouldn't you just send the eventResult object directly?

framework/src/org/apache/cordova/CordovaInterfaceImpl.java
riknoll: pluginManager should be created when the first URL is loaded. In the default Android project, this happens in `onCreate()` so it shouldn't be an issue.

framework/src/org/apache/cordova/ResumeCallback.java
riknoll: I'll add a comment to the code explaining this. It has to do with how we encode plugin results. If we were to just send one event object, we would have to first translate the PluginResult obtained by the ResumeCallback into something we could place in the event plugin result's JSONObject. Instead of duplicating our PluginResult decoding code on the Native side of things, I just send back two results and handle it in platform.js

framework/src/org/apache/cordova/ResumeCallback.java
riknoll: Yup, that sounds reasonable

Titoine:
Hi,
First of all, thanks from the great work!
But I'm getting a lot of crash from this problem.

Do you have any idea when this will be merge ?

riknoll:
Hey all, I've added a PR to the docs repo that documents the result callback part of this change as well as general Android lifecycle considerations: https://github.com/apache/cordova-docs/pull/428

infil00p:
@Titoine Are you having problems after the patch or before? I'm not sure.

Titoine:
@infil00p Before the patch.

Bnaya:
@riknoll, after going over the docs (which looks great btw) its seems it supports only single plugin result
What if you wait for two or more plugins responses? like file system operation and camera plugin result?

riknoll:
@Bnaya That is correct and it was intentional. Is there a need for extending this to all plugin callbacks currently pending? The way I see it, the burden is on the app developer for being careful about what plugins are launching external Activities and maintaining their state accordingly (for example, not launching the camera activity while waiting on a file transfer). Plugins have access to the Activity lifecycle on the native side, so they should be able to handle whatever is happening. I guess plugins lose the ability to notify the javascript side about what's going on, (like a file transfer failing) but hopefully the app developer is being careful so that those situations don't occur.

riknoll:
Also, I don't think you can get into a situation where you have two Activity results pending (correct me if I'm wrong on that one).

Bnaya:
I will extend the scope of that issue.

This change can/should also help the apps developers to recover from situations that the app sent to the background not only by activity from the app, but also if the user hit the home button/notifications etc
And lets say the app had several IO ongoing operations by plugins. even 2 operations by the same plugin.
The operating system can kill the activity also in such case, and then if the user navigates back to the app and the webview will be reloaded with resume event.

infil00p:
@Bnaya @riknoll Unless you're getting a result back from an Activity, the plugin processes belong to the activity that was put in the background, and the plugins would die with the activity that was killed.  Therefore, there could only be one result returning to the application.  This issue is related to communication between two activities, not trying to resurrect plugin code, which is really just class that's not even guaranteed to have its own thread (although it should).

I think the problems are different enough that the scope shouldn't be changed here, and that the second problem that your mentioning needs to be fleshed out more.

riknoll:
Alright, I plan on merging this in soon. I want to do some more testing and write some plugin-side docs before I do, but since this PR's been open a while and I don't think this will really be much of a breaking change, I am going to push forward. I plan to merge it in early next week. Let me know if there are any objections and I will try to address them!

riknoll:
I will submit a PR that updates the camera plugin to use the save-restore API as well, which should fix CB-9189. I'll also take a look at the contacts plugin, because I believe it uses external activities.

riknoll:
Apologies for not getting to this last week like I said I would (I didn't really have much of an Internet connection). I did some more testing on this PR today and found a few bugs/edge cases that I've now fixed. This PR has been rebased to master and I pushed a new commit that handles the following:

* PluginResults with multipart messages are now handled properly in the `resume` payload
* The `resume` event is now fired when the Activity is destroyed even if there isn't a pending result
* A potential concurrency issue in CoreAndroid has been handled

Please review this new commit. Since I pushed some changes I will wait to merge this in until end of day tomorrow. Let met know if there are any objections. For the record, I tested as follows:

* Ran mobilespec auto tests on a 6.0.0 device and a 5.1.1 emulator (minus the file transfer tests) and confirmed output matched master (one orientation failure on 5.1.1, but I believe it is the emulator's fault)
* Ran the platform unit tests on a 6.0.0 device and a 5.1.1 emulator and confirmed output matched master (one failure)
* Did some manual testing with a 6.0.0 device and a 5.1.1 emulator

jasongin:
LGTM

Titoine:
When I try my app with "Don't keep activities" on, my app crash and I have the "unfortunately MyApp has stopped" message. The app doesn't restart. Will it restart with your fix or on some devices the problem still occur ?
On Wiko Cink Five with Android 4.1.2

Bnaya:
@Titoine This is not a complete fix, and its not released yet
There's also corresponding changes to the camera plugin this change
When it will be released you will need to update your android platform and your camera plugin.

For now, you can try my fix for that issue, which includes more fixes for other edge cases with the plugin
https://github.com/photomania/cordova-plugin-camera/commits/prevent-crash-on-external-image-selection

Note that its not rebased with the lasted code of the camera plugin but its works ok and we use if in production

Titoine:
Ok! With your fix or in futur update, when the app crash, is the user have to launch it again or does It launch itself ? Or both of the case could happen ?

Bnaya:
The app will not crash but just restart itself without the user doing anything.

The thing is that, if you want to use the picture he choose before the restart you will need another code path to handle it. the callback you gave to getPicture will not be called.

You can try an just let him to select photo again, but then the app might restart soon

Titoine:
Thanks, I understand now.
Hope to see this implemented soon. It's a critical issue for me.
Anyway, thanks you all for the great work!

PR #240, opened 2015-11-11
CB-9971 Redirect gradlew stderr to current process stdout

vladimir-kotikov:
This is a fix for https://issues.apache.org/jira/browse/CB-9971


jasongin:
Does java.exe output compilation errors and warnings to stderr or stdout? If they go to stderr then we should NOT redirect those.

bin/templates/cordova/lib/builders/GradleBuilder.js
```diff
@@ -211,3 +209,46 @@ module.exports = GradleBuilder;
...
+ */
+function spawnWithStderrRedirect(cmd, args) {
+    return Q.Promise(function (resolve, reject) {
+        // Work around spawn not being able to find .bat files.
+        var opts = { stdio: [0,1,1] };
+        if (process.platform === 'win32') {
+            var joinedArgs = [cmd]
+                .concat(args)
+                .map(function(a){
+                    if (/^[^"].* .*[^"]/.test(a)) return '"' + a + '"';
```
dblotsky: What if `a` contains `"` characters? For example, if `a` is `program" files "folder`? Is there a cross-platform shell-escaping library we can use instead?

bin/templates/cordova/lib/builders/GradleBuilder.js
```diff
@@ -211,3 +209,46 @@ module.exports = GradleBuilder;
...
+                    return a;
+                })
+                .join(' ');
+
+            args = ['/s', '/c'].concat('"' + joinedArgs + '"')/*, '"' + [cmd].concat(args).map(function(a){if (/^[^"].* .*[^"]/.test(a)) return '"' + a + '"'; return a;}).join(' ')+'"']*/;
```
dblotsky: Is the comment at the end of the line necessary?

bin/templates/cordova/lib/builders/GradleBuilder.js
```diff
@@ -211,3 +209,46 @@ module.exports = GradleBuilder;
...
+            cmd = 'cmd';
+            opts.windowsVerbatimArguments = true;
+        }
+        require('child_process')
```
dblotsky: Please put the `require` at the top of the file.

dblotsky:
Also :+1: on what @jasongin said.

bin/templates/cordova/lib/builders/GradleBuilder.js
```diff
@@ -211,3 +209,46 @@ module.exports = GradleBuilder;
 function isAutoGenerated(file) {
     return fs.existsSync(file) && fs.readFileSync(file, 'utf8').indexOf(MARKER) > 0;
 }
+
+/**
+ * A special superspawn-like implementation, required to workaround the issue
```
jasongin: Why not just add another option to the existing spawn() function?

vladimir-kotikov: Not even an additional option, but an ability to accept any valid `stdio` option - `superspawn` now knows only `default`, `ignore` and `inherit`. I agree, this will be the most correct solution, but will require patching this in cordova-common and releasing it before next android release.  

@jasongin, what do you think?

bin/templates/cordova/lib/builders/GradleBuilder.js
vladimir-kotikov: This escaping works exactly the same way in `superspawn` and we hadn't had any problems with whitespaces, caused by this implementation, for a long time, so i'd prefer just not to change it.

vladimir-kotikov:
@jasongin, i will double check this

bin/templates/cordova/lib/builders/GradleBuilder.js
dblotsky: Fair enough. However if `superspawn` code has a bug and is updated, then this code will be out of date. It's better not to duplicate code and instead modify and reuse `superspawn`.

bin/templates/cordova/lib/builders/GradleBuilder.js
dblotsky: It looks like `superspawn` should be modified to take the same parameters as `child_process.spawn`. Refer [here](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options) for Node 5.0.x, and [here](https://nodejs.org/docs/latest-v0.12.x/api/child_process.html#child_process_child_process_spawn_command_args_options) for Node 0.12.x.

bin/templates/cordova/lib/builders/GradleBuilder.js
```diff
@@ -211,3 +212,54 @@ module.exports = GradleBuilder;
...
+ *   with java printing some necessary information to stderr instead of stdout.
+ *   This function redirects all stderr messages to current process stdout. See
```
jasongin: Update the comments based on the revised code that only redirects the one line.

bin/templates/cordova/lib/builders/GradleBuilder.js
```diff
@@ -211,3 +212,54 @@ module.exports = GradleBuilder;
...
+        cmd = 'cmd';
+        opts.windowsVerbatimArguments = true;
+    }
+
+    return Q.Promise(function (resolve, reject) {
+        var proc = child_process.spawn(cmd, args, opts);
+
+        proc.stdout.on('data', process.stdout.write.bind(process.stdout));
+        proc.stderr.on('data', function (data) {
+            var isRedirectNeeded = data.toString().match(/^Picked up _JAVA_OPTIONS: /i);
```
jasongin: /regex/.test(str) is faster when all you need is a boolean result, not the detailed info that match() returns.

bin/templates/cordova/lib/builders/GradleBuilder.js
jasongin: This is not really a consideration anymore now that the redirection has to be more specialized to this particular case

bin/templates/cordova/lib/builders/GradleBuilder.js
```diff
@@ -211,3 +212,54 @@ module.exports = GradleBuilder;
...
+            process[isRedirectNeeded ? 'stdout' : 'stderr'].write(data);
```
jasongin: Are we sure that putting that line on stdout will not cause different problems? Would it be better to completely suppress it? I don't think anyone actually cares to see it.

dblotsky:
Can't we still use superspawn and just add something like the `stderr.on('data')` filter to catch and discard the undesired line? Superspawn can return a promise that receives the stdout and stderr of the spawned command.

vladimir-kotikov:
@dblotsky, i propose to proceed with current approach for now, otherwise we'll have to release 'cordova-common' with updated superspawn before this fix.

I created [CB-10052](https://issues.apache.org/jira/browse/CB-10052) for `superspawn` improvement, and once it is implemented, we could remove this cusom logic and switch to superspawn again.

dblotsky:
Doesn't superspawn already return a promise whose output can be processed? Superspawn doesn't need to be modified.

vladimir-kotikov:
.It does, right, but there is two problems:

1. Superspawn returns output only _after_ the underlying process exits - if the build lasts long enough, this might be confusing for the user,

2. Superspawn either resolves with stdout or rejects with some error message that contains stderr output - there is no way to get both stdout and stderr after the promise completed

dblotsky:
I'm pretty sure we can modify this line: https://github.com/MarcDiethelm/superspawn/blob/master/lib/index.js#L166 to return both the stdout and stderr. To me it seems like writing a whole new function is going to result in maintenance bugs in the future.

sgrebnov:
I think current implementation is good enough so let's merge this PR.

:+1:  to the idea of improving superspawn function so it can cover this case as well, but I think this is out of scope of this PR so let's continue discussion at CB-10052.

PR #241, opened 2015-12-02
CB-10112 Parse additional CLI arguments properly

vladimir-kotikov:
This fixes [CB-10112](https://issues.apache.org/jira/browse/CB-10112) by updating `build.js` to take an addtional platform options (passed behing `--`) from correct options object


stevengill:
Thanks for the quick fix!

PR #242, opened 2015-12-07
Update CordovaWebViewImpl.java

wangsongyan:
use resetCoreAndroidPlugin method to set variable null when opening other url.
when I use cordova-android framework in my project,it is troubling me that the backbutton event doesn't work if I trriger the backbutton on other url.
ps:I import the cordova.js into my html page more than once.

### DIFF
--- a/framework/src/org/apache/cordova/CordovaWebViewImpl.java
+++ b/framework/src/org/apache/cordova/CordovaWebViewImpl.java
@@ -135,6 +135,8 @@ public class CordovaWebViewImpl implements CordovaWebView {
         if (recreatePlugins) {
             // Don't re-initialize on first load.
             if (loadedUrl != null) {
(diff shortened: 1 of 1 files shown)


infil00p:
The only way the back button breaks is if you override the behaviour on the document, and it is the developer's responsibility to manage that, not the framework.  For that reason, we're going to not accept this one, since you shouldn't have to change the CordovaWebView's implementation to fix the issue that you are having.

wangsongyan:
I found that all plugins including CoreAndroid will be initialized when jumping to other page,but the variable appPlugin of CordovaWebViewImpl cannot be changed if appPlugin is not a null variable.Not only the back button but also the volumeKeyDown and the volumeKeyUp cannot work.
Is it the only way to solve my problem that using the cordova.js only once?

PR #243, opened 2015-12-09
Include native libraries from armeabi directory when producing armv7 apk

mbektchiev:
If a Cordova Android project contains `.so` files located in the `lib/armeabi` directory, they will not be included in the final apk when building for the `armv7` architecture with `cdvBuildMultipleApks` set to `true`

### DIFF
--- a/bin/templates/project/build.gradle
+++ b/bin/templates/project/build.gradle
@@ -176,7 +176,7 @@ android {
             armv7 {
                 versionCode defaultConfig.versionCode + 2
                 ndk {
-                    abiFilters "armeabi-v7a", ""
+                    abiFilters "armeabi-v7a", "armeabi"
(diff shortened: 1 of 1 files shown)


mbektchiev:
@infil00p @stevengill @vladimir-kotikov 
Can someone review this?

vladimir-kotikov:
I don't have an experience in this area, so i'd prefer to @infil00p or @stevengill took a look at this

infil00p:
Do you have a test case to hand us? So far the only native code that we've seen used is Crosswalk, and it'd be good to have like a libexample.so or something for testing.

infil00p:
This appears to be abandoned, so I'm going to close this.  If not, can you please give us a test cases so we can test this PR?

PR #244, opened 2015-12-09
CB-10140 Use application id instead of package name when running

CookieCookson:


vladimir-kotikov:
@CookieCookson, LGTM, though I'm not sure about situations when this would be useful.

However the problem here is that aapt might be not in the $PATH, so probably you need to check this and set up $PATH correspondingly in [check_reqs](https://github.com/apache/cordova-android/blob/master/bin/lib/check_reqs.js)

CookieCookson:
@vladimir-kotikov Thanks for the response! I have created the PR as a fix to an issue with a plugin I am developing. See https://github.com/CookieCookson/cordova-plugin-lifecycle and it should explain its use. 

I shall investigate check_reqs.

vladimir-kotikov:
Thanks for explanation, @CookieCookson

infil00p:
@CookieCookson This seems to have fallen by the wayside for quite a few months.  Are you still looking to add this, or should I close it since it no longer merges cleanly?

CookieCookson:
@infil00p I haven't had the time available to work any more on this, and have got around the issue by other means! I would say close it for now.

fredgalvao:
@infil00p This issue still exists. For anyone changin applicationId on android, to use flavours or whitelabel-like solutions, running cordova will always try to launch the base package name. It should definitely launch the applicationId instead. I mistakenly created [another issue](https://issues.apache.org/jira/browse/CB-11374) a while ago for this, and I just now discovered this issue because of the mailing list notifications.

I'd love to see this topic being fixed/improved. There are a few other comments in [this issue](https://issues.apache.org/jira/browse/CB-11373) as well, regarding this subject.

infil00p:
@fredgalvao The PR was closed, not the issue.

janpio:
Was this added via another PR in the meantime? Seems like a useful change.

PR #245, opened 2015-12-09
Update theme to Theme.DeviceDefault.NoActionBar

dpogue:
This uses the device system theme instead of the old Gingerbread-era theme for dialogs and alerts.

Available as of API 14, which is now the minSDK version for cordova-android.

### DIFF
--- a/bin/templates/project/AndroidManifest.xml
+++ b/bin/templates/project/AndroidManifest.xml
@@ -35,7 +35,7 @@
         <activity android:name="__ACTIVITY__"
                 android:label="@string/activity_name"
                 android:launchMode="singleTop"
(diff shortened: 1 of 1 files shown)


riknoll:
LGTM. Makes sense

infil00p:
Makes sense to me as well.  Let's create a JIRA issue for this for housekeeping and pull this in.

nikhilkh:
@dpogue Could you change the title to point to this JIRA: CB-10295. Looks like a good one to merge.

dpogue:
Already merged as 73fd9e4dfa5dffde2eb14ebd6c736ca462d1e29f

PR #247, opened 2015-12-21
CB-10014: Set gradle applicationId to package name.

dpogue:
This fixes an issue that's affecting several Cordova Android apps where Google Play services causes conflicts that prevent multiple apps from being installed.

JIRA issue: https://issues.apache.org/jira/browse/CB-10014
See also: https://code.google.com/p/android/issues/detail?id=193567
http://tools.android.com/tech-docs/new-build-system/applicationid-vs-packagename

### DIFF
--- a/bin/templates/project/build.gradle
+++ b/bin/templates/project/build.gradle
@@ -163,6 +163,8 @@ android {
 
     defaultConfig {
         versionCode cdvVersionCode ?: Integer.parseInt("" + privateHelpers.extractIntFromManifest("versionCode") + "0")
+        applicationId privateHelpers.extractStringFromManifest("package")
+
         if (cdvMinSdkVersion != null) {
             minSdkVersion cdvMinSdkVersion
         }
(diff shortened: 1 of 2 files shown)


Bnaya:
Consider using this code,
https://github.com/one-signal/OneSignal-Cordova-SDK/commit/dd9fb05cfb570a8e01e40f1ebf1ce4ed42079882
its based on xml full parse and not string matching

dpogue:
I think I'd agree that `def manifest = new XmlSlurper().parse(file("AndroidManifest.xml"))` is better. In this case I was just trying to copy what already existed for reading the version number.

It looks like the helpers use a slightly different method for parsing the config.xml:
```gradle
def doGetConfigXml() {
    def xml = file("res/xml/config.xml").getText()
    // Disable namespace awareness since Cordova doesn't use them properly
    return new XmlParser(false, false).parseText(xml)
}
```

I'm not sure which approach is best. Maybe @infil00p or one of the Crosswalk devs has a better idea of gradle best practices? (I'm pretty sure none of this is considered "best practices" though)

macdonst:
@dpogue I'm currently doing the same thing in the push plugin to fix the issue. So I'm :+1: on this change.

infil00p:
LGTM, I'll merge it in.

dpogue:
Merged in fb9cf60c4100d5bc6f9320747611da1ca0ced289

PR #251, opened 2016-01-21
CB-10157 Uninstall app from device/emulator only when signed apk is already installed

vladimir-kotikov:
JIRA: [CB-10157](https://issues.apache.org/jira/browse/CB-10157)


nikhilkh:
LGTM - Not clear why travis test is failing.

PR #252, opened 2016-01-26
CB-10421 Fixes exception when calling run script with --help option

vladimir-kotikov:
See [CB-10421](https://issues.apache.org/jira/browse/CB-10421)


vladimir-kotikov:
@sgrebnov, could you please take a look

sgrebnov:
LGTM

PR #253, opened 2016-01-27
Clearing appPlugin when loading pages.

Dannysw:
When changing a page the appPlugin becomes stale, causing the "backbutton" and other events to no longer be passable from the CordovaWebView. By setting it to null when a page is loaded it will force the updated appPlugin when it is needed next.


Dannysw:
Jira ticket https://issues.apache.org/jira/browse/CB-10614

PR #254, opened 2016-01-27
CB-10138 Adds missing plugin metadata to plugin_list module

vladimir-kotikov:
This is a fix for [CB-10138](https://issues.apache.org/jira/browse/CB-10138)


vladimir-kotikov:
@sgrebnov, PTAL

sgrebnov:
:+1:

PR #257, opened 2016-02-03
CB-10498: Resume event should be sticky if it has a plugin result

riknoll:
Changes the resume event behavior as follows:

1. When the app starts and there is no plugin result to be delivered, the **resume event will not fire** (same as current behavior)
2. When the app is left and then resumed (e.g. press the home button and then return to the app), the **resume event will fire but will not be sticky** (same as current behavior)
3. When the app is recreated after a plugin Activity finishes (i.e. the event payload contains a pending result), the **resume event will fire and will be sticky**

### DIFF
--- a/cordova-js-src/platform.js
+++ b/cordova-js-src/platform.js
@@ -19,6 +19,9 @@
  *
 */
 
+// The last resume event that was received that had the result of a plugin call.
+var lastResumeEvent = null;
+
 module.exports = {
     id: 'android',
     bootstrap: function() {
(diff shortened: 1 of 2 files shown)


cordova-js-src/platform.js
```diff
@@ -58,6 +61,19 @@ module.exports = {
         bindButtonChannel('volumeup');
         bindButtonChannel('volumedown');
 
+        // The resume event is not "sticky", but it is possible that the event
+        // will contain the result of a plugin call. We need to ensure that the
+        // plugin result is delivered even after the event is fired (CB-10498)
+        var cordovaAddEventListener = document.addEventListener;
```
adamduren: This is really the original document.addListener. Maybe the name should be changed to make that more clear.

I think this is a simple and clean solution. However, I am a little weary about monkey patching `document.addEventListener`. I'm not sure what a better alternative would be though without adding unnecessary complexity.

riknoll: @adamduren We already patch `addEventListener` in the common Cordova js code (hence the `cordovaAddEventListener` name). See https://github.com/apache/cordova-js/blob/master/src/cordova_b.js#L48

Come to think of it, I should probably be doing the method in window as well. I'll update the PR. Thanks for taking a look!

riknoll: Actually, it doesn't seem like Cordova events fire in the window...

riknoll:
@infil00p I'd appreciate it if you could review this one

csantanapr:
@riknoll I tested the pull request and it's working for me.
lastResume keeps the data, any new listener that get's attached get's the data.

csantanapr:
@riknoll you are going to merge this or there is more work to be done?

riknoll:
Sorry @csantanapr, I was dealing with some plugin fixes. I will merge this by the end of today; I need to test on a device.

csantanapr:
@riknoll sure no problem, I was just curious if there was more you wanted me to look at.

adamduren:
Any idea when this will be released?

PR #258, opened 2016-02-11
CB-10510: Add an optional timeout to emu start script

riknoll:
The `emulator.js` script waits forever for the emulator to boot. If the emulator gets stuck, it will never terminate. The reason for adding this timeout is that the Android emulator on Windows occasionally hangs while booting up. This causes problems in our CI, where the run step would go on forever until buildbot eventually cut it off (see [here](https://ci.apache.org/builders/cordova-android-win/builds/1588/steps/running-tests/logs/stdio) for an example run). I have an upcoming PR for cordova-medic that switches from using `cordova run` to start the emulator to requiring this script directly so that we can add retry logic for the emulator start. I don't think it really makes sense to expose the timeout option in the CLI itself, so this PR does not alter the CLI's behavior at all (it will still hang). I'm not sure if we should add in a default timeout or not.

### DIFF
--- a/bin/templates/cordova/lib/emulator.js
+++ b/bin/templates/cordova/lib/emulator.js
@@ -40,6 +40,7 @@ var ONE_SECOND              = 1000; // in milliseconds
 var ONE_MINUTE              = 60 * ONE_SECOND; // in milliseconds
 var INSTALL_COMMAND_TIMEOUT = 5 * ONE_MINUTE; // in milliseconds
 var NUM_INSTALL_RETRIES     = 3;
+var CHECK_BOOTED_INTERVAL   = 3 * ONE_SECOND; // in milliseconds
 var EXEC_KILL_SIGNAL        = 'SIGKILL';
 
 /**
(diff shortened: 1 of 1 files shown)


riknoll:
@dblotsky @nikhilkh please review

nikhilkh:
LGTM

dblotsky:
LGTM.

bin/templates/cordova/lib/emulator.js
```diff
@@ -227,18 +236,25 @@ module.exports.wait_for_emulator = function(uuid) {
...
+            return true;
+        } else if (time_remaining === 0) {
+            return false;
         } else {
             process.stdout.write('.');
-            return Q.delay(3000).then(function() {
-                return self.wait_for_boot(emulator_id);
+
+            // Check at regular intervals
+            return Q.delay(time_remaining < CHECK_BOOTED_INTERVAL ? time_remaining : CHECK_BOOTED_INTERVAL).then(function() {
```
dblotsky: Maybe use `Math.min` here instead of the ternary operator? It's a little hard to read long ternary conditions.

riknoll:
@vladimir-kotikov can you take a quick look at this? I want to make sure I'm not messing with the platform API at all given that we don't have any tests for this code path.

riknoll:
@omefire

PR #259, opened 2016-02-15
CB-10618 Handle gradle frameworks on plugin installation/uninstallation

vladimir-kotikov:
An implementation for [CB-10618](https://issues.apache.org/jira/browse/CB-10618)

This PR brings into platform logic for processing gradle files after plugin installation (instead of keeping it in LIB).

It also removes `prepBuildFiles` method from `build` module exports, and by this disables this logic being called from LIB (as there is a check for `build.prepBuildFiles` existence - see https://github.com/apache/cordova-lib/blob/master/cordova-lib/src/plugman/install.js#L625)

Also see apache/cordova-lib#389 for corresponding LIB changes


nikhilkh:
LGTM. How can we add tests for this scenario?

vladimir-kotikov:
@nikhilkh, added tests

PR #260, opened 2016-02-16
CB-10628 Fix emulate android --target

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-10628)

### DIFF
--- a/bin/templates/cordova/lib/run.js
+++ b/bin/templates/cordova/lib/run.js
@@ -27,6 +27,19 @@ var path  = require('path'),
     device   = require('./device'),
     Q = require('q');
 
+function getInstallTarget(runOptions) {
+    var install_target;
+    if (runOptions.target) {
+        install_target = runOptions.target;
+    } else if (runOptions.device) {
+        install_target = '--device';
+    } else if (runOptions.emulator) {
+        install_target = '--emulator';
+    }
+
+    return install_target;
+}
+
 /**
  * Runs the application on a device if available. If no device is found, it will
  *   use a started emulator. If no started emulators are found it will attempt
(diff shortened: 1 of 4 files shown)


nikhilkh:
How can we test this?

bin/templates/cordova/lib/run.js
```diff
@@ -41,8 +41,9 @@ var path  = require('path'),
 
     var self = this;
 
-    var install_target = runOptions.device ? '--device' :
-        runOptions.emulator ? '--emulator' :
+    var install_target = !runOptions.target ?
```
vladimir-kotikov: IMO this statement looks very complex - there is 3 ternary operators in one expression. Could you please try to simplify this

daserge:
@vladimir-kotikov, addressed, thanks.
@nikhilkh, I've tried to test private `getInstallTarget` with `rewire` but faced an issue caused by the fact that the test-time directory structure differs from the run-time one (so that `check-reqs` fails to load when rewire requires run -> emulator -> check_reqs):

```javascript
var rewire = require("rewire");
var run = rewire("../../bin/templates/cordova/lib/run");
var getInstallTarget = run.__get__("getInstallTarget");

describe("run", function () {
  describe("getInstallTarget", function() {
    var targetOpts = { target: "emu" };
    var deviceOpts = { device: true };
    var emulatorOpts = { emulator: true };
    var emptyOpts = {};

    it("should select correct target based on the run opts", function() {
        expect(getInstallTarget(targetOpts)).toBe("emu");
        expect(getInstallTarget(deviceOpts)).toBe("--device");
        expect(getInstallTarget(emulatorOpts)).toBe("--emulator");
        expect(getInstallTarget(emptyOpts)).toBeUndefined();
    });
  });
});
```

daserge:
I will take a look at this once more tomorrow.

daserge:
Worked on this with @vladimir-kotikov - there is a way to solve this using `require.cache` but it looks like an overkill for this case. 
Decided to require `check_reqs` at runtime instead.

daserge:
@nikhilkh, please take a look.

nikhilkh:
LGTM

PR #261, opened 2016-02-22
CB-10660 fixed the exception when removing a non-existing directory

bso-intel:
It is ALWAYS safer to check if the directory exists before it tires to remove the directory.
The uninstallation of the plugin with a <framework> tag failed because of this issue.


bso-intel:
@vladimir-kotikov ,
Could you review this?

vladimir-kotikov:
LGTM, merging

bso-intel:
Thanks for merging, Vladimir.

PR #263, opened 2016-02-26
CB-10669 cordova run --list cannot find 'adb'

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-10669)


nikhilkh:
LGTM

PR #265, opened 2016-02-29
CB-10729 Move plugin handlers tests into platform's repo

vladimir-kotikov:
JIRA: https://issues.apache.org/jira/browse/CB-10729


bin/templates/cordova/lib/pluginHandlers.js
```diff
@@ -31,7 +31,7 @@ var handlers = {
             if (!obj.src) throw new CordovaError('<source-file> element is missing "src" attribute for plugin: ' + plugin.id);
             if (!obj.targetDir) throw new CordovaError('<source-file> element is missing "target-dir" attribute for plugin: ' + plugin.id);
             var dest = path.join(obj.targetDir, path.basename(obj.src));
-            copyNewFile(plugin.dir, obj.src, project.projectDir, dest, options && options.link);
+            copyNewFile(plugin.dir, obj.src, project.projectDir, dest, !!(options && options.link));
```
nikhilkh: Did the test find a bug here?

nikhilkh:
LGTM - though one test is failing.

bin/templates/cordova/lib/pluginHandlers.js
vladimir-kotikov: Just some divergence between expected spy arguments and actual ones, Not a bug probably, but i preferred to change parameter type rather than tests.

PR #266, opened 2016-02-29
Update SystemWebViewEngine

alin1771:
There is no point of having font size under 1px (which is default minimum size) BUT when we use EM and REM units (font-size: 0.1em)there might be some UI issues when the equivalent size in PX is smaller than 1


infil00p:
Is there a JIRA issue corresponding to this? Is there an example of px being less than one somehow?

alin1771:
No, I didn't create a jira issue. I've just had this issue on my own apps

infil00p:
Please create a JIRA issue with repro steps so that we can see what's actually being fixed, otherwise we can't accept this pull request.

alin1771:
That is too much for me and I don't have enough time to learn how to do that, I've tried to help others by fixing an issue that I had an others might have too.
Please reject my pull request,
Regards

PR #267, opened 2016-03-02
CB-10710 Respect plugman's www option when installing plugins

vladimir-kotikov:
JIRA [CB-10710](https://issues.apache.org/jira/browse/CB-10710)

This change returns back an ability to specify alternative destination for `www` files when installing plugin using Plugman CLI.

### DIFF
--- a/bin/templates/cordova/Api.js
+++ b/bin/templates/cordova/Api.js
@@ -515,4 +515,6 @@ Api.prototype._processPluginOptions = function (options) {
             this.locations.www :
             this.locations.platformWww;
     }
+
+    return options;
 };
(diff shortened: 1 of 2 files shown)


dblotsky:
Is this supported in other platforms too?

dblotsky:
Are there tests for this? I don't know how to test this to verify that the fix works.

vladimir-kotikov:
> Is this supported in other platforms too? If not, should it be?

It sound like platform-agnostic Plugman's feature, so my thinking is that it should. The PRs for other platforms is on their way.

> Are there tests for this? I don't know how to test this to verify that the fix works

Here is verification steps: 

    cordova create foo && cd foo
    cordova platform add android@https://github.com/msopentech/cordova-android#CB-10710
    plugman -d install --platform android --project platforms/android --plugin cordova-plugin-statusbar --www my/custom/www/dir

Then verify the contents of `/my/custom/www/dir`. Plugin js files should be there along `with cordova_plugins.js` Also there should be no `cordova-plugin-statusbar` files neither in platforms/android/platform_www nor in platforms/android/assets/www

dblotsky:
Okay, those steps passed for me on OS X.

PR #271, opened 2016-03-08
5.1.x fullscreen no immersive

oratthumzap:
hi,
we are using Cordova and would like to have the ability to set the Cordova activity in full screen mode, and not in an immersive mode regardless of the Android version (as was the case in previous Cordova versions).
therefore, we added a new configuration key to the Cordova config: "FullscreenNotImmersive".
when set to true the Cordova activity will be started using the full-screen-not-immersive flags.
otherwise, it will be started using the relevant flags according to the Android version (the original v5.1.x implementation). by default, this configuration key is set to false (= original implementation).
we would appreciate if you could merge this pull request so that other developers would be able to enjoy this logic ;)


infil00p:
Can you clean up this pull request?

oratthumzap:
@infil00p, do you want me to refactor the code or remove the pull request altogether?

daserge:
@oratthumzap, you should rebase your commit on master so that only relevant changes are shown.

PR #273, opened 2016-03-10
[CB-10600] 'cordova run --release' uses wrong apk

dpolivy:
When doing `cordova run android --release`, the CLI attempts to use the
android-release-unsigned.apk, which naturally fails to install on the
device because it is unsigned. The release build does also generate the
properly signed android-release.apk, and that should be used instead.

The apk-finding logic for "debug" excludes unsigned builds, so the
"release" logic should follow a similar pattern.

Fixes https://issues.apache.org/jira/browse/CB-10600

### DIFF
--- a/bin/templates/cordova/lib/builders/GenericBuilder.js
+++ b/bin/templates/cordova/lib/builders/GenericBuilder.js
@@ -109,7 +109,7 @@ function findOutputApksHelper(dir, build_type, arch) {
             return /-debug/.exec(apkName) && !/-unaligned|-unsigned/.exec(apkName);
         }
         if (build_type === 'release') {
-            return /-release/.exec(apkName) && !/-unaligned/.exec(apkName);
+            return /-release/.exec(apkName) && !/-unaligned|-unsigned/.exec(apkName);
         }
         return true;
     })


nikhilkh:
@daserge Can you please review?

daserge:
Tested - master version works with both `--buildConfig` and `-- --keystore`.
With this PR the CLI outputs the following:
```
BUILD SUCCESSFUL

Total time: 18.57 secs
Built the following apk(s):

Error: Could not find apk architecture: x86 build-type: release
```
which is IMO misleading as `android-release-unsigned.apk` is actually built (and can be signed manually). 
Probably we can improve [this message](https://github.com/apache/cordova-android/blob/82582e5a5bb3b7472d51812d902b5074227aa41b/bin/templates/cordova/lib/build.js#L235) instead to state about the need of apk signing directly.

dpolivy:
@daserge Thanks for looking at this. What's the exact cmd line you're using that generates that error? Is that due to the change I made? It seems like in that case the signing is not happening, so the unsigned build isn't being picked up, but I don't think the unsigned build will run anyways so you're right the error message should probably change to reflect the need to sign. Do you want me to update this PR with that change or do you want to handle it separately?

daserge:
I've tested this way:
- `created project, added master android platform`
- `λ cordova run android --release` ->
```
Built the following apk(s):
        f:/coho/CB-10600/test/platforms/android/build/outputs/apk/android-release-unsigned.apk
Using apk: f:/coho/CB-10600/test/platforms/android/build/outputs/apk/android-release-unsigned.apk
Error: Failed to install apk to emulator: WARNING: linker: libdvm.so has text relocations. This is wasting memory and is a security risk. Please fix.

        pkg: /data/local/tmp/android-release-unsigned.apk

Failure [INSTALL_PARSE_FAILED_NO_CERTIFICATES]
```
- `λ cordova run android --release -- --keystore=../my-release-key.keystore --storePassword=password --alias=alias_name --password=password` ->
```
BUILD SUCCESSFUL

Total time: 6.286 secs
Built the following apk(s):
        f:/coho/CB-10600/test/platforms/android/build/outputs/apk/android-release-unsigned.apk
        f:/coho/CB-10600/test/platforms/android/build/outputs/apk/android-release.apk
Using apk: f:/coho/CB-10600/test/platforms/android/build/outputs/apk/android-release-unsigned.apk
Error: Failed to install apk to emulator: WARNING: linker: libdvm.so has text relocations. This is wasting memory and is a security risk. Please fix.

        pkg: /data/local/tmp/android-release-unsigned.apk

Failure [INSTALL_PARSE_FAILED_NO_CERTIFICATES]
```
So the first issue is that even though it has built a signed apk it uses an unsigned one because of last modification time sorting probably.

If we do `cordova clean android` and run `-- --keystore` again it will succeed (`--buildConfig` option works the same way).

Using your PR version makes `unsigned` apk to not be listed so that successive builds after `cordova run android --release` work although I would better change the sorting logic to prefer a signed build and output a message about signing requirement in case of `run` command if the filter has returned `unsigned` apk.

dpolivy:
@daserge That makes sense; in the case where a release build is built, but not automatically signed as part of the process, I can see that run would not pick up any APK with this PR.

You know this code better than I, so I defer to you on a proper fix. Feel free to reject this PR and propose your own fix. It would be good to have this addressed soon, though!

daserge:
@dpolivy, I've sent #276

infil00p:
I like #276 better, can you guys pop over, give it a look and comment on that? Thanks!

dpolivy:
@daserge Thanks! I will close this one.

PR #276, opened 2016-03-16
CB-10600 'cordova run android --release' does not use signed and zip-…

daserge:
…aligned version of APK

De-prioritize unsigned builds
Adds an actionable hint for INSTALL_PARSE_FAILED_NO_CERTIFICATES error
Removes filtering by architecture when not specified

[Jira issue](https://issues.apache.org/jira/browse/CB-10600)

### DIFF
--- a/bin/templates/cordova/lib/emulator.js
+++ b/bin/templates/cordova/lib/emulator.js
@@ -368,8 +368,14 @@ module.exports.install = function(givenTarget, buildResults) {
                         if (err) reject(new CordovaError('Error executing "' + command + '": ' + stderr));
                         // adb does not return an error code even if installation fails. Instead it puts a specific
                         // message to stdout, so we have to use RegExp matching to detect installation failure.
-                        else if (/Failure/.test(stdout)) reject(new CordovaError('Failed to install apk to emulator: ' + stdout));
-                        else resolve(stdout);
+                        else if (/Failure/.test(stdout)) {
+                            if (stdout.match(/INSTALL_PARSE_FAILED_NO_CERTIFICATES/)) {
+                                stdout += 'Sign the build using \'-- --keystore\' or \'--buildConfig\'' +
+                                    ' or sign and deploy the unsigned apk manually using Android tools.';
+                            }
+                            
+                            reject(new CordovaError('Failed to install apk to emulator: ' + stdout));
+                        } else resolve(stdout);
                     });
                 });
             }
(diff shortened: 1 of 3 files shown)


infil00p:
LGTM

bin/templates/cordova/lib/emulator.js
```diff
@@ -363,8 +363,14 @@ module.exports.install = function(givenTarget, buildResults) {
...
-                        else if (/Failure/.test(stdout)) reject(new CordovaError('Failed to install apk to emulator: ' + stdout));
-                        else resolve(stdout);
+                        else if (/Failure/.test(stdout)) {
+                            if (stdout.match(/INSTALL_PARSE_FAILED_NO_CERTIFICATES/)) {
+                                stdout += 'Sign the build using \'-- --keystore\' or \'--buildConfig\'' +
+                                    ' or sign and deploy the unsigned apk manually using Android tools.';
+                            }
+                            
+                            reject(new CordovaError('Failed to install apk to emulator: ' + stdout));
+                        } else resolve(stdout);
```
dblotsky: Does it make sense to de-duplicate this error string since it also occurs in `Adb.js`, or would that be too cumbersome?

daserge: I agree but this should be probably done along with error codes/messages introducing task (can't find Jira for it; dev list thread: http://apache.markmail.org/thread/qsyocgri3zq6nkrh).

dblotsky: Cool. Thanks for explaining. Do you have the context on this task in order to file a JIRA? If not, I'll read up and file it.

daserge: That would be great, thanks - I don't have enough context unfortunately.

dblotsky: Created a JIRA: https://issues.apache.org/jira/browse/CB-10903

pke:
This problem seems to be still existing in 5.1.1
I get only `release-unsigned.apk` built.

daserge:
@pke, the fix is not released yet.
You can use master version for now: `cordova platform add https://github.com/apache/cordova-android`.

pke:
ah ok, thanks. For the future: Are fixed + released PR marked here with a special label?

daserge:
The tags a commit is released under are below the commit description, for example:

https://github.com/apache/cordova-android/commit/bc9e895e2537f212756eb8e18559fac2cb02c536
`master (#276)`

https://github.com/apache/cordova-android/commit/ce2525d4d8b2b7c2af0b467f99490529c14beb1d
`master (#260)  rel/5.1.1 5.1.1`

PR #281, opened 2016-03-18
CB-10837 Support platform-specific orientation on Android

vladimir-kotikov:
This removes processing of "orientation" preference lets user to specify any platform-specific value for Android.

See [CB-10837](https://issues.apache.org/jira/browse/CB-10837)

### DIFF
--- a/bin/templates/cordova/lib/prepare.js
+++ b/bin/templates/cordova/lib/prepare.js
@@ -131,7 +131,7 @@ function updateProjectAccordingTo(platformConfig, locations) {
     var orig_pkg = manifest.getPackageId();
 
     manifest.getActivity()
-        .setOrientation(findOrientationValue(platformConfig))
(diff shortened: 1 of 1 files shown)


vladimir-kotikov:
Note that to fix orienattion issues ultimately we'll need also https://github.com/apache/cordova-lib/pull/411 and https://github.com/apache/cordova-lib/pull/412

omefire:
What happens if the orientation provided is not supported by the Android platform ?

vladimir-kotikov:
When set invalid value gradle failed with

```
:processDebugResources
E:\PROJECTS\Temp\foo\platforms\android\build\intermediates\manifests\full\debug\AndroidManifest.xml:35:40-43 : AAPT: String types not allowed (at 'screenOrientation' with value 'foo').
```

omefire:
Ok, its better we leave the validation to gradle then.

omefire:
LGTM!

PR #283, opened 2016-03-18
CB-10897 Refactor URI Parsing for Whitelist

ktop:
Fix for CB-10897. 

I moved out the code that parses the 'origin' from the whitelist to a separate function so that it can also be used to parse the URLs that request access instead of using Android's Uri class parse function. Android's Uri parse function does not parse 'host' the same way that Cordova parses the host in origin for whitelist, so host ends up being null and the comparison fails. Using the same parsing methods is the way to go to give accurate comparisons of URLs to those in the whitelist. 

Also, I can have specific non http/https URLs in my whitelist now. 
Something like this works now without the wildcard:
`<allow-intent href="market://details?id=com.google.android.apps.maps" />`

### DIFF
--- a/framework/src/org/apache/cordova/Whitelist.java
+++ b/framework/src/org/apache/cordova/Whitelist.java
@@ -26,8 +26,6 @@ import java.util.regex.Pattern;
 
 import org.apache.cordova.LOG;
 
-import android.net.Uri;
-
 public class Whitelist {
     private static class URLPattern {
         public Pattern scheme;
@@ -79,7 +77,7 @@ public class Whitelist {
             }
         }
 
-        public boolean matches(Uri uri) {
+        public boolean matches(URLParts uri) {
             try {
                 return ((scheme == null || scheme.matcher(uri.getScheme()).matches()) &&
                         (host == null || host.matcher(uri.getHost()).matches()) &&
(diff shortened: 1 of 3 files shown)


ktop:
@infil00p can you take a look at this PR?

infil00p:
Can we have tests in the test project for this?  It looks fine, but since it's new functionality in cordova-android, adding a test would be good.

ktop:
I've added a test now. 

Should we open another jira issue to update the Whitelist in the test project?
It's still using the <access> tags in config.xml and in order to test my changes for the Whitelist tests, I had to manually add the whitelist plugin. Not sure if there's an easy way to pull down the latest Whitelist plugin and add it to the test project. Maybe it's a manual step we can document in the readme.

ktop:
@infil00p can you review again? see my previous comment.

infil00p:
We should be able to unit test the methods on the Whitelist object itself without plugins.  I also don't see the commit where the test is.

infil00p:
What's the status for this? I refactored the unit tests so we can actually have tests that back this code now, as opposed to before where it wasn't quite as straight forward.

infil00p:
This is closed as abandoned.  It's sad, because asking for tests shouldn't cause anything to become abandoned.

PR #284, opened 2016-03-25
CB-10862 Cannot set minsdkversion

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-10862)


vladimir-kotikov:
LGTM

nikhilkh:
Another scenario where we need a test case - perhaps an end-to-end one - thoughts?

PR #285, opened 2016-03-25
CB-10963: Handle overlapping permission requests from plugins

riknoll:
Adds support for overlapping permission requests from plugins (i.e. requests that are made while another request is still pending). The old behavior was to just return the last permission request, which meant that plugins could inadvertently cancel requests made by other plugins.

### DIFF
--- a/framework/src/org/apache/cordova/CallbackMap.java
+++ b/framework/src/org/apache/cordova/CallbackMap.java
@@ -43,7 +43,7 @@ public class CallbackMap {
      * @return              A unique request code that can be used to retrieve this callback
      *                      with getAndRemoveCallback()
      */
-    public int registerCallback(CordovaPlugin receiver, int requestCode) {
+    public synchronized int registerCallback(CordovaPlugin receiver, int requestCode) {
         int mappedId = this.currentCallbackId++;
         callbacks.put(mappedId, new Pair<CordovaPlugin, Integer>(receiver, requestCode));
         return mappedId;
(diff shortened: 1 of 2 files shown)


riknoll:
@infil00p @jasongin please review when you have a moment.

framework/src/org/apache/cordova/CallbackMap.java
```diff
@@ -0,0 +1,65 @@
...
+     * Stores a CordovaPlugin and request code and returns a new unique request code to use
+     * in a permission request.
+     *
+     * @param receiver      The plugin that is making the request
+     * @param requestCode   The original request code used by the plugin
+     * @return              A unique request code that can be used to retrieve this callback
+     *                      with getAndRemoveCallback()
+     */
+    public int registerCallback(CordovaPlugin receiver, int requestCode) {
+        int mappedId = this.currentCallbackId++;
```
jasongin: Can this be called concurrently? If so, consider using java.util.concurrent.atomic.AtomicInteger.

jasongin: Also then you'd need a collection tolerant of concurrency instead of SparseArray.

framework/src/org/apache/cordova/CallbackMap.java
```diff
@@ -0,0 +1,65 @@
...
+
+import android.util.Pair;
+import android.util.SparseArray;
+
+/**
+ * Provides a collection that maps unique request codes to CordovaPlugins and Integers.
+ * Used to ensure that when plugins make requests for runtime permissions, those requests do not
+ * collide with requests from other plugins that use the same request code value.
+ */
+public class CallbackMap {
```
jasongin: Why public? It's only used in this package right?

framework/src/org/apache/cordova/CallbackMap.java
riknoll: Updated to synchronize the methods themselves

framework/src/org/apache/cordova/CallbackMap.java
jasongin: Never mind, it looks like basically everything in this package is public/protected to allow for easy extensibility.

jasongin:
LGTM

infil00p:
Can we somehow create a test for this?  I don't know how this is testable, since I still don't know exactly how to use JUnit to test onActivityResult.

riknoll:
Perhaps we can mock the Activity's request permissions method? I don't have a lot of experience writing JUnit tests for Android.

PR #286, opened 2016-04-06
CB-10964 Handle build.json file starting with a BOM

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-10964)


nikhilkh:
This needs to be handled for other platforms too.

PR #288, opened 2016-04-07
CB-8383 Handle pause event when keeprunning=false

GitRubb:
I added this pull request to hopefully get some attention to this bug https://issues.apache.org/jira/browse/CB-8383. When keeprunning = false in config.xml, any pause event processing is halted until the app returns. Not only is the pause event missed on app backgrounding, but on return it processes the pause and resume events.


infil00p:
This does not work, since all actions on a WebView thread MUST happen on the UI thread.  Also, you don't actually start the thread at all and this patch breaks the pause/resume functionality of KeepRunning entirely.

Adding a thread just for the sake of adding a thread is generally a bad idea, especially when you have to pause something that runs on the UI thread.

infil00p:
Closing pull request due to it disabling pause/resume behaviour.

PR #289, opened 2016-04-08
CB-11022 Duplicate www files to both destinations on plugin add/rm

vladimir-kotikov:
This is a supplementary PR for https://github.com/apache/cordova-lib/pull/423 which removes requirement for running prepare to copy plugin's files to `www` dir.


jasongin:
LGTM

PR #291, opened 2016-04-15
CB-11036 'args.slice is not a function' when building windows with ot…

daserge:
…her platform

Avoid parseOpts side effects on options obj passed in as it's then passed to other platforms

[Jira issue](https://issues.apache.org/jira/browse/CB-11036)


nikhilkh:
Should the cloning happen in `cordova-lib` instead - to ensure isolation?

daserge:
You are right, thanks.
Filed https://github.com/apache/cordova-lib/pull/430

PR #292, opened 2016-04-15
CB-11015 Error adding plugin with gradle extras

daserge:
[Jira issue](https://issues.apache.org/jira/browse/CB-11015)


daserge:
@vladimir-kotikov, could you please review?

vladimir-kotikov:
LGTM

PR #293, opened 2016-04-18
Update CoreAndroid.java

dheran:
No modifier (package-private) class to public class


infil00p:
I agree, this should be made public.

PR #295, opened 2016-04-19
CB-11117: Use FileUpdater to optimize prepare for android platform

jasongin:
This uses the FileUpdater module added in https://github.com/apache/cordova-lib/pull/429 to optionally skip copying files that didn't change. Some refactoring was required because previously the target directories would just be wiped before copying; now we need to map out the source and target directories so the FileUpdater has the necessary information to determine the optimal set of file operations.


jasongin:
@vladimir-kotikov @infil00p please review

bin/templates/cordova/lib/prepare.js
```diff
@@ -26,26 +26,43 @@ var AndroidManifest = require('./AndroidManifest');
...
 
     var self = this;
+    var platformResourcesDir = path.relative(cordovaProject.root, path.join(this.locations.root, 'res'));
 
     this._config = updateConfigFilesFrom(cordovaProject.projectConfig,
         this._munger, this.locations);
 
     // Update own www dir with project's www assets and plugins' assets and js-files
-    return Q.when(updateWwwFrom(cordovaProject, this.locations))
+    return Q.when(updateWww.call(self, cordovaProject))
```
vladimir-kotikov: I'd rather not to pass `this` reference to this method, especially if the only thing we need is a couple of properties from 'locations' object (`this.locations.platformWww` and `this.locations.www` at L120). Maybe just pass these properties as an arguments, or revert to previous signature?

bin/templates/cordova/lib/prepare.js
```diff
@@ -26,26 +26,43 @@ var AndroidManifest = require('./AndroidManifest');
...
     .then(function () {
         // update project according to config.xml changes.
         return updateProjectAccordingTo(self._config, self.locations);
     })
     .then(function () {
-        handleIcons(cordovaProject.projectConfig, self.root);
-        handleSplashes(cordovaProject.projectConfig, self.root);
+        updateIcons.call(self, cordovaProject, platformResourcesDir);
+        updateSplashes.call(self, cordovaProject, platformResourcesDir);
```
vladimir-kotikov: Same as above. I can't find any usage of `this` inside of these functions, so binding them to `this` is redundant

bin/templates/cordova/lib/prepare.js
```diff
@@ -84,29 +101,49 @@ function updateConfigFilesFrom(sourceConfig, configMunger, locations) {
...
+/**
+ * Cleans all files from the platform 'www' directory.
+ */
+function cleanWww(projectRoot) {
+    var targetDir = path.relative(projectRoot, this.locations.www);
+    events.emit('verbose', "Cleaning " + targetDir);
+
+    // No source paths are specified, so mergeAndUpdateDir() will clear the target directory.
+    FileUpdater.mergeAndUpdateDir(
+        [], targetDir, { rootDir: projectRoot, all: true }, logFileOp);
```
vladimir-kotikov: Passing logger function here would produce a lot of output if `www` contains a lot of files. IMO everything that end user would need to know is that we're going to wipe all `www` contents

vladimir-kotikov: Also is there any advantage of using `FileUpdater` rather than just calling `shell.rm()`?

bin/templates/cordova/lib/prepare.js
```diff
@@ -26,26 +26,43 @@ var AndroidManifest = require('./AndroidManifest');
...
+    // but the project root dir and config can be resolved here easily.
+    var projectRoot = path.resolve(this.root, "../..");
+    var projectConfig = new ConfigParser(this.locations.configXml);
+    var platformResourcesDir = path.relative(projectRoot, path.join(this.locations.root, 'res'));
+
+    var self = this;
+    return Q().then(function () {
+        cleanWww.call(self, projectRoot);
+        cleanIcons.call(self, projectRoot, projectConfig, platformResourcesDir);
+        cleanSplashes.call(self, projectRoot, projectConfig, platformResourcesDir);
```
vladimir-kotikov: Same here - `this` reference in not used inside of the functions above

bin/templates/cordova/lib/prepare.js
```diff
@@ -26,26 +26,43 @@ var AndroidManifest = require('./AndroidManifest');
...
     })
     .then(function () {
-        events.emit('verbose', 'updated project successfully');
+        events.emit('verbose', 'Prepared project successfully');
+    });
+};
+
+module.exports.clean = function (options) {
+    // Unfortunately the cordovaProject isn't passed into the clean() function,
+    // but the project root dir and config can be resolved here easily.
```
vladimir-kotikov: The `cordovaProject` is not passed here because the `clean` command is also accessible in non-CLI workflow through shell script (as opposed to `prepare`, which is intended to be called from CLI only). In this case resolving project's root is a bad idea, as there might not be any project at all.

Also, calling this in non-CLI project would destroy the project, as these files are not copied from the root project but were added by developer manually. IMO this functionality should be hidden behind the flag, so user would not wipe these files unintentionally.

bin/templates/cordova/lib/prepare.js
```diff
@@ -295,25 +344,47 @@ function handleIcons(projectConfig, platformRoot) {
...
-function deleteDefaultResourceAt(baseDir, resourceName) {
-    shell.ls(path.join(baseDir, 'res/drawable-*'))
+function cleanIcons(projectRoot, projectConfig, platformResourcesDir) {
+    var icons = projectConfig.getIcons('android');
+    if (icons.length > 0) {
+        var resourceMap = mapImageResources(projectRoot, platformResourcesDir, 'icon.png');
+        events.emit('verbose', 'Cleaning icons at ' + platformResourcesDir);
+
+        // No source paths are specified in the map, so updatePaths() will delete the target files.
+        FileUpdater.updatePaths(
```
vladimir-kotikov: Just like in `cleanWww` - is there any advantage of calling this rather than doing `shell.rm` on these files? This would  eliminate the need to know how `mapImageResources` and `FileUpdater.updatePaths` works internally and IMO improve readability a lot.

bin/templates/cordova/lib/prepare.js
jasongin: I guess I don't understand the non-CLI workflow very well. Why would 'cordova clean' be called in a non-CLI workflow, rather than 'gradle clean' or equivalent? Doesn't "non-CLI workflow" mean the cordova CLI commands like 'cordova clean' aren't being used?

In a normal cordova CLI workflow, I think users would expect that the files copied by prepare would be cleaned up by 'cordova clean' without any options; I was surprised to find they were not already.

bin/templates/cordova/lib/prepare.js
jasongin: I had been using `this` in a few more places, before I moved some things around. I agree it is not really needed now; I'll clean that up everywhere.

bin/templates/cordova/lib/prepare.js
vladimir-kotikov: In fact there is no big difference between calling `gradle clean` and `cordova/clean.bat`but non-CLi workflow assumes that there is a contract between platform and user, according to which platform offers a set of shell scripts to control the aspects related to build system, and the `clean` is one of these scripts just as `build`, `run`, etc - you can find them inside of 'android/cordova' folder

One of advantages of calling `clean` rather that `gradle clean` is that you might not have `gradle` available in path, but `clean` contains some logic (similar to build) to find gradle wrapper, and call it with appropriate arguments

vladimir-kotikov: In CLI world i think there is not much sense in cleaning www folder because we treat platforms as build artifacts and hence this is it's something that is out of user control - instead we just need to ensure that every time when 'prepare' completes the user's 'www' dir and platform's 'www' dir are in sync.

jasongin: I don't think everyone using CLI workflow treats platforms as build artifacts. Regardless, I think this kind of clean behavior is not needed now that incremental prepare does a stronger sync (copying different file times rather than just newer times). So I'll remove it for now. We can discuss the merits of something like 'cordova prepare --clean' later.

bin/templates/cordova/lib/prepare.js
jasongin: I was mostly going for consistent logging: why log some file operations but not others? Our customer feedback indicates they would like more visibility into the cordova build process, and this kind of verbose logging will help with that.

Anyway this is going away now that I'm removing the clean function.

bin/templates/cordova/lib/prepare.js
jasongin: After discussion with Nikhil, we'd really like to keep the clean functionality. I'm adding a check for invocation via the non-CLI clean script; I hope that addresses your concern.

jasongin:
I updated the PR based on feedback. As mentioned in the other comment I've kept the clean functionality, but it is now skipped when not invoked via the CLI.

vladimir-kotikov:
LGTM

PR #296, opened 2016-04-21
CB-11132 Fix Error: Cannot read property 'match' of undefined in cord…

daserge:
…ova-android emulator.js

Please find the details in the [Jira issue](https://issues.apache.org/jira/browse/CB-11132).


sarangan12:
LGTM

PR #298, opened 2016-04-22
CB-9465 replace Integer.parseInt with BigInteger so that you can use longer Android version codes

pa-de-solminihac:
I used to have DNS like version codes (YYYYmmmddXX format) for my apps builds, and I can't go back in older apps because Google Play wont allow me to upload inferior version codes, so I thing we should to use BigInteger instead of Integer.parseInt

### DIFF
--- a/bin/templates/project/build.gradle
+++ b/bin/templates/project/build.gradle
@@ -162,7 +162,7 @@ android {
     }
 
     defaultConfig {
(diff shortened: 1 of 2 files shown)


pa-de-solminihac:
I am afraid this commit was useless (if not harmful), as [the documentation](https://developer.android.com/studio/publish/versioning.html#appversioning) states:
> Warning: The greatest value Google Play allows for versionCode is 2100000000.

PR #299, opened 2016-04-26
CB-11153 Use Gradle 2.8 correctly for building Android projects

nikhilkh:
This improves gradle build performance by 1 second on my machine

@jasongin; @infil00p; @riknoll  to help review


infil00p:
Why 2.8? Why not 2.10?   Is 2.10 slower than 2.8?  Will this break other users somehow?

nikhilkh:
I just ended up moving to version 2.13 which is the latest - and the perf is even a bit better than version 2.8. I looked at Android docs and did not find anything about one version being more compatible than other - so we should be fine.

infil00p:
Good call on updating this to Gradle 2.13, I just updated to 2.1.0 beta, and Android Studio just showed me a bunch of complaints about importing the project.

infil00p:
OK, I merged this in, because I need it for something I'm working on :P

PR #300, opened 2016-04-27
CB-10096: Upgrading to Gradle Plugin 2.1.0

infil00p:
This is a minor change, but I need people running stable to make sure it works. @riknoll Can you take a quick look?

### DIFF
--- a/bin/templates/project/build.gradle
+++ b/bin/templates/project/build.gradle
@@ -31,7 +31,7 @@ buildscript {
     // http://tools.android.com/tech-docs/new-build-system/version-compatibility
     // and https://issues.apache.org/jira/browse/CB-8143
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.5.0'
+        classpath 'com.android.tools.build:gradle:2.1.0'
     }
    }
 
--- a/framework/build.gradle
+++ b/framework/build.gradle
@@ -24,7 +24,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:1.5.0'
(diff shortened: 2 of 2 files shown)


riknoll:
Sure, I'll try it out

riknoll:
I built/ran mobilespec with and without crosswalk and I didn't encounter any issues. One weird thing that did happen was that I wasn't able to increase the size of the gradle daemon heap from within gradle.properties. Gradle's build output was advising me to edit my gradle.properties file so that I could use the new dexing-in-process feature, but I couldn't get that to work.  Passing it as an argument worked:
```
cordova build android -- --gradleArg=-Dorg.gradle.jvmargs=-Xmx2048M
```

infil00p:
@riknoll Did you test Instant Run on Android Studio? I don't expect this to work with Crosswalk.  We need help from them getting Crosswalk to work with the new Gradle stuff

riknoll:
I tried it out a bit just now (without crosswalk). Java code changes refreshed fine for me. The webview gets restarted of course, but my java changes were reflected as expected and the app did not get restarted. Changes to js, html, etc. were less speedy (app gets killed/restarted). I was just building mobilespec so I didn't really see any drastic improvements in that scenario since it already builds pretty quickly. Are there any specific scenarios you want me to try out?

infil00p:
@riknoll It should skip HTML and JS changes.  If not, this isn't really worth doing other than correctness.

nikhilkh:
@infil00p  Let's get this in - I get a prompt in Android Studio without this.

infil00p:
OK, I pushed this in.  Closing

PR #301, opened 2016-04-28
CB-11138 Reuse PluginManager from common to add/rm plugins

vladimir-kotikov:
A change to use `PluginManager` class to manage plugin installation/uninstallation instead of custom logic

JIRA: [CB-11138](https://issues.apache.org/jira/browse/CB-11138)


nikhilkh:
@jasongin Can you please help review.

jasongin:
Can I ignore all the changes under node_modules? I don't understand why that is checked in to source control...

vladimir-kotikov:
Sure, you can skip them safely. This is because we ship them as 'bundled' dependencies, since we don't run `npm install` after downloading the platform.

jasongin:
:+1:

PR #302, opened 2016-04-28
CB-11163 Android Builds are failing in CI

daserge:
CB-11133 Handle android emulator start failure (reverted from commit abcaee24f9bd0e680f8d06ced5e0d363d63cffb7)


nikhilkh:
Can you explain what was the issue here and why this fixes it?

daserge:
Reverted https://github.com/apache/cordova-android/commit/abcaee24f9bd0e680f8d06ced5e0d363d63cffb7 to unblock the CI.
Reopened the [CB-11133](https://issues.apache.org/jira/browse/CB-11133) - the proper fix is to be investigated.

The issue was that the main process in emulator.js was waiting for adb emulator child process end because of our subscription to stdout (in which the ERROR message was being caught) and was never returned as emulator is running in the background.

PR #304, opened 2016-05-04
CB-6936: fix crash when calling methods on a destroyed webview

uareurapid:
Hi guys. This has happened in some of my apps, not often, but a few times. I also have see some reports on the web.
Take a look at this stack overflow thread for instance:
http://stackoverflow.com/questions/36239052/nullpointerexception-in-cordova-app-reported-by-fabric-sdk
I think that better than just doing a try/catch is just check if the webView is valid/null or not


infil00p:
Do you have a stack trace for this? This is old, but it doesn't conflict, so I might pull it in, but it would help to see an error report or something before throwing this over the fence.

infil00p:
1. We need a JIRA for this
2. This has become more interesting since we need to add an Android O version of this code to handle it better, because the renderer goes away, not the whole webview now.

PR #305, opened 2016-05-11
Improving Android build output and verbose logging

riknoll:
Improving grammar, consistency of language, detail, etc. for verbose and regular output. I'm going to use Android as an example for Windows and iOS to follow. There wasn't much that was missing log-wise, mostly just needed to be clearer about what was happening. The changes in #295 will also improve output a lot (though they might conflict with this PR a bit). Also see https://github.com/apache/cordova-lib/pull/439


jasongin:
:shipit:

riknoll:
Sorry @jasongin, I made a few more updates. Mind taking a look?

jasongin:
LGTM

PR #306, opened 2016-05-20
CB-10096: Upgrade test project to Gradle Plugin 2.1.0

tony--:


infil00p:
Why did this break travis???

infil00p:
This looks completely unrelated to the pull request itself.  We should just merge this in.

infil00p:
OK, this is a pre-req of #307, so I merged them both together.

PR #307, opened 2016-05-20
CB-11292 fix broken MessageChannel after plugins are recreated

tony--:
also added a new test to prevent regression - probably overkill but the issue was hard to isolate


tony--:
AppVeyor timed out, not a failure related to the PR.

PR #308, opened 2016-05-31
CB-11343: Handle Java versions after JEP-223

ankon:


infil00p:
So, how do we test this one? Is this a Mac OS X or Linux issue?

ankon:
> So, how do we test this one? Is this a Mac OS X or Linux issue?

It's a "whereever you are able to run Java 9 (EA)"-issue.

infil00p:
Yeah, I'm amazed this fails, since I've been working on a different issue with this regex for a day finding out that check_reqs.js needs to be redone entirely because it just wants a result, literally any result.

ankon:
I'm not sure I follow here. In any case, what is needed to get this forward, and prepare the plugin for actually working with JDK 9?

infil00p:
I couldn't even get Java 9 to work with Android when I installed it, even when it did pass the check, so for now I don't think I can accept this request right now.  If you can try to update the Regex with the new branch, and show that you can actually compile Hello World using Java 9 Early Access, I might reconsider.

PR #309, opened 2016-06-06
CB-11383: Update to gradle for using jcenter and use app plugin

infil00p:
This is a housekeeping pull request so we can use gradle more easily.


macdonst:
@infil00p should we leave `mavenCentral` in just in case?

infil00p:
@macdonst Yes, we should.

infil00p:
OK, Looks good, I'm going to merge this.

PR #310, opened 2016-06-07
[CB-10916] Support display name for Android

dpogue:
This is the Android Platform API support side of https://github.com/apache/cordova-lib/pull/453.

### DIFF
--- a/bin/templates/cordova/lib/prepare.js
+++ b/bin/templates/cordova/lib/prepare.js
@@ -167,9 +167,16 @@ function cleanWww (projectRoot, locations) {
  */
 function updateProjectAccordingTo (platformConfig, locations) {
     // Update app name by editing res/values/strings.xml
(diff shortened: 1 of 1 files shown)


infil00p:
Looks fine, but we should wait for it all to go in together if possible.  @dpogue You have commit rights, so can you make sure that this gets pulled in when the parent issue gets resolved?

dpogue:
Will do :+1:

shazron:
iOS, cordova-lib has this pulled in, and also docs has a PR. Just waiting on this one to be pulled in.

infil00p:
This appears to have fell by the wayside.  Why hasn't this been pulled into cordova-android?

dpogue:
Sorry, been on vacation the past few weeks, but it looks like we've already bundled in the required version of cordova-common, so this is good to go!

PR #312, opened 2016-06-24
CB-11481:android-library is deprecated use com.android.library instead

vivekkiran:


infil00p:
This seems trivial enough to accept, done!

PR #314, opened 2016-07-06
CB-9489 Fixed "endless waiting for emulator" issue

alsorokin:
https://issues.apache.org/jira/browse/CB-9489

Only emulators with API level 22+ are affected by this.

Fixed by changing the wait logic: we don't poll every started emulator anymore. Instead, we're launching the emulator on a specific port and poll only this instance, waiting for boot to complete.


vladimir-kotikov:
LGTM

PR #321, opened 2016-07-28
CB-8978 Prepare copy resource-files from config.xml

dpogue:
See also cordova/cordova-discuss#6
Depends on https://github.com/apache/cordova-lib/pull/468

This copies files specified by `<resource-file>` elements in config.xml into the Android project when preparing. This allows providing arbitrary files such as special notification-sized icons, or API configuration JSON files.

/cc @jasongin @jcesarmobile

### DIFF
--- a/bin/templates/cordova/lib/prepare.js
+++ b/bin/templates/cordova/lib/prepare.js
@@ -48,6 +48,7 @@ module.exports.prepare = function (cordovaProject, options) {
     .then(function () {
         updateIcons(cordovaProject, path.relative(cordovaProject.root, self.locations.res));
         updateSplashes(cordovaProject, path.relative(cordovaProject.root, self.locations.res));
+        updateFileResources(cordovaProject, path.relative(cordovaProject.root, self.locations.root));
     })
     .then(function () {
         events.emit('verbose', 'Prepared android project successfully');
@@ -72,6 +73,7 @@ module.exports.clean = function (options) {
         cleanWww(projectRoot, self.locations);
         cleanIcons(projectRoot, projectConfig, path.relative(projectRoot, self.locations.res));
         cleanSplashes(projectRoot, projectConfig, path.relative(projectRoot, self.locations.res));
+        cleanFileResources(projectRoot, projectConfig, path.relative(projectRoot, self.locations.root));
     });
 };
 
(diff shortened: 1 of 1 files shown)


purplecabbage:
bin\templates\cordova\lib\prepare.js: line 81, col 9, 'cleanFileResources' is not defined.
bin\templates\cordova\lib\prepare.js: line 433, col 10, 'cleanResourceFiles' is defined but never used.

dpogue:
Thanks! I've corrected that function name

stevengill:
Needs review with master cordova-common. Aka cordova-common@2 (releasing soon)

dpogue:
Yeah, I'm going to try to get it set up locally with cordova-common master and ensure it's all working as intended

dpogue:
cordova-common@2.0.0 is released now, which was the prerequisite for this.
Should we bump the cordova-common dependency in package.json and merge this in for the next Android release?

shazron:
@dpogue @infil00p  is this out in the latest cordova-android? Are there docs for this, or is it a copy-paste from plugin.xml?

dpogue:
It is in cordova-android@6.2.1.
No docs yet (wanted to get it supported in more platforms first), but it works exactly the same way that it does in plugin.xml.

bin/templates/cordova/lib/prepare.js
```diff
@@ -401,6 +403,44 @@ function mapImageResources(rootDir, subDir, type, resourceName) {
...
+
+    events.emit('verbose', 'Updating resource files at ' + platformDir);
+    FileUpdater.updatePaths(
+        resourceMap, { rootDir: cordovaProject.root }, logFileOp);
+}
+
+
+function cleanFileResources(projectRoot, projectConfig, platformDir) {
+    var files = projectConfig.getFileResources('android');
+    if (files.length > 0) {
```
shazron: @dpogue @infil00p  I noticed that files.length is always zero here when I do a `cordova clean` even though I have a `<resource-file>` tag in config.xml.  Is this function useless then effectively?

shazron: (same result in iOS in my PR)

shazron:
Leaving some notes here for future reference (in case there are bugs):
1. Firstly `cleanFileResources` is called, but effectively useless during runtime, even if you have a `<resource-file>` tag in config.xml since prepare does not send a project config (see my line comments above and https://github.com/apache/cordova-android/blob/dbce4de5b6313b8f10db64166612a8bcb56d898b/bin/templates/cordova/lib/prepare.js#L59)
2. Because of item 1 above, this part of the feature cannot be tested
3. No platform unit tests for this feature, which is important because of item 2 above (no effective test available)
4. The corresponding iOS implementation of this feature (https://issues.apache.org/jira/browse/CB-8980 and https://github.com/apache/cordova-ios/pull/305) has an example of unit tests

dpogue:
With regards to point 1 above, I've found the root cause:
In a project-level config.xml, `resource-file` is only supported under a `platform` tag. These files get copied into the native platforms.

Within each native platform, there's a config.xml specific to that platform. When the `resouce-file` tags are copied into this file, they are not nested under a `platform` tag. This means `ConfigParser.getFileResources` was not finding them, and hence returning an empty list of files to clean.

I've opened https://github.com/apache/cordova-lib/pull/547 to add an option to `ConfigParser.getFileResources` to return the top-level `resource-file` tags.
I've also opened https://github.com/apache/cordova-ios/pull/309 (iOS) and https://github.com/apache/cordova-android/pull/377 (Android) to pass in that new option when cleaning.

I will look into getting unit tests for Android's prepare step set up to address the rest of the feedback.

PR #323, opened 2016-08-11
Changes for Cordova-Android 6.0.0 - Android Modules as Plugins

infil00p:
Redoing this branch as gradle_respect, where we don't nuke the gradle files of AARs, Library Projects and other things that have gradle files.  This should resolve a lot of irritating errors, but slows down plugin installation with file checks. :/

We do need a better solution to this, but it would require a refactor of how we do builds.

### DIFF
--- a/bin/templates/cordova/lib/builders/GradleBuilder.js
+++ b/bin/templates/cordova/lib/builders/GradleBuilder.js
@@ -69,16 +69,17 @@ GradleBuilder.prototype.prepBuildFiles = function() {
     var pluginBuildGradle = path.join(this.root, 'cordova', 'lib', 'plugin-build.gradle');
     var propertiesObj = this.readProjectProperties();
     var subProjects = propertiesObj.libs;
+    var checkAndCopy = function(subProject, root) {
+      var subProjectGradle = path.join(root, subProject, 'build.gradle');
+      fs.exists(subProject, function(exists) {
+        if (!exists) {
+          shell.cp('-f', pluginBuildGradle, subProjectGradle);
+        }
+      });
+    };
     for (var i = 0; i < subProjects.length; ++i) {
         if (subProjects[i] !== 'CordovaLib') {
-            var subProjectGradle = path.join(this.root, subProjects[i], 'build.gradle');
-            // Only copy the gradle if it doesn't exist for the library
-            fs.exists(subProjectGradle, function(exists) {
-              if (!exists)
-                {
-                  shell.cp('-f', pluginBuildGradle, path.join(this.root, subProjects[i], 'build.gradle'));
-                }
-            });
+          checkAndCopy(subProjects[i], this.root);
         }
     }
     var name = this.extractRealProjectNameFromManifest();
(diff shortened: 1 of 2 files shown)


bin/templates/cordova/lib/plugin-build.gradle
```diff
@@ -21,24 +21,15 @@
...
-    } else if (gradle.gradleVersion >= "2.1") {
-        dependencies {
-            classpath 'com.android.tools.build:gradle:0.14.0+'
-        }
-    } else {
-        dependencies {
-            classpath 'com.android.tools.build:gradle:0.12.0+'
-        }
+    dependencies {
+        classpath 'com.android.tools.build:gradle:1.0.0+'
```
tony--: I'm not very familiar with gradle compatibility issues, but it appears that this change removes backwards compatibility.  I'm guessing anything <1.0.0 is deprecated/EOL and no switching logic is required anymore.

It seems like the preceding comment should be updated or removed since there is no more switching logic and the information in CB-8143 is obsolete?  I don't know enough to update the comment appropriately, sorry.

tony--:
Just checking my understanding...

1. checkAndCopy only copies when the subProject does not have a build.gradle
2. insertExclude ensures that there is only one reference to CordovaLib in the project by inspecting each subProject's build.gradle, finding references and excluding them when present

bin/templates/cordova/lib/builders/GradleBuilder.js
```diff
@@ -93,10 +102,24 @@ GradleBuilder.prototype.prepBuildFiles = function() {
     // Update dependencies within build.gradle.
     var buildGradle = fs.readFileSync(path.join(this.root, 'build.gradle'), 'utf8');
     var depsList = '';
+    var root = this.root;
+    var insertExclude = function(p) {
+          var gradlePath = path.join(root, p, 'build.gradle');
+          var projectGradleFile = fs.readFileSync(gradlePath, 'utf-8');
+          if(projectGradleFile.indexOf('CordovaLib') != -1) {
```
tony--: What if there is a reference to MyCordovaLib or CordovaLib appears in a comment?
Will it be safe to add the exclude if there is not really a reference to CordovaLib?

bin/templates/cordova/lib/plugin-build.gradle
infil00p: You're correct about the switching logic being removed and it not being required.  The information in CB-8143 is obsolete, and we already did this in the main gradle file.  We should be using the latest gradle tools and keeping up to date with the stable Android Studio updates, which updates extremely often, and usually updates automatically.

bin/templates/cordova/lib/builders/GradleBuilder.js
infil00p: It should be safe to add the exclude, since CordovaLib itself will be included in the project.  We require this exclusion because if you make plugins modules, you have to include the CordovaLib file somehow for Android Studio to work properly.  In the near future, we should just use the AAR and not have to deal with this slower solution.  However, it's good to have this just in case.

infil00p:
This is TOO SLOW for creating a project after I started rebasing it with master.  I'm going to close this and come back to it a bit later, since it's REALLY slow right now and I don't want to slow down the workflow.

infil00p:
Re-opening pull request to run CI to see if there's any change in speed since the last changes to the build.

infil00p:
Locally, the creation of projects is faster.  Time to test plugin installation.

PR #327, opened 2016-09-01
API 24 Changes Pull Request

infil00p:
This is an outstanding PR for updating Cordova so that it works with API Level 24.  This is important since Nougat is actually out.


infil00p:
Whoops, there's something seriously wrong with this PR, closing.

PR #328, opened 2016-09-01
Nougat update - Updates required for API Level 24

infil00p:
This includes the ability to grab the build config and deliver it to plugins, which is important for obtaining the Application ID.

### DIFF
--- a/framework/src/org/apache/cordova/CoreAndroid.java
+++ b/framework/src/org/apache/cordova/CoreAndroid.java
@@ -30,6 +30,7 @@ import android.content.IntentFilter;
 import android.telephony.TelephonyManager;
 import android.view.KeyEvent;
 
+import java.lang.reflect.Field;
 import java.util.HashMap;
 
 /**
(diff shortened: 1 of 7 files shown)


framework/src/org/apache/cordova/CoreAndroid.java
```diff
@@ -357,4 +358,33 @@ public void sendResumeEvent(PluginResult resumeEvent) {
...
+      /*
+     * This needs to be implemented if you wish to use the Camera Plugin or other plugins
+     * that read the Build Configuration.
+     *
+     * Thanks to Phil@Medtronic and Graham Borland for finding the answer and posting it to
+     * StackOverflow.  This is annoying as hell!
+     *
+     */
+
+    public static Object getBuildConfigValue(Context ctx, String key)
```
pbakondy: Notice:  if you want to use this method within a ProGuard-obfuscated build the application will exit with `java.lang.ClassNotFoundException`

infil00p: This class can't be used with ProGuard obfuscation, since it's a Cordova plugin in the first place.

PR #330, opened 2016-09-09
Bridge Fixes for Android Jellybean

infil00p:
Updating the Javascript to detect whether we're using Jellybean or not using the userAgent, and pick the right bridge based on the version of Android.  This isn't the best solution, but it's good enough until people stop using Jellybean.

### DIFF
--- a/bin/templates/project/assets/www/cordova.js
+++ b/bin/templates/project/assets/www/cordova.js
@@ -901,7 +901,7 @@ var cordova = require('cordova'),
         EVAL_BRIDGE: 3
     },
     jsToNativeBridgeMode,  // Set lazily.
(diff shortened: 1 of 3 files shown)


dpogue:
Is there a good way to check if we're using the system webview when making this change? I suspect Crosswalk could use the evaluateJavascript bridge even on Jellybean?

infil00p:
@dpogue Again, it would be in the userAgent, but I'm fine with leaving ONLINE_BRIDGE as the default for Crosswalk, since we also force Crosswalk to use prompt for the jsToNative bridge even though it could probably use addJavascriptInterface without everything burning down.

Closing PR for now, I merged this due to priority.

PR #332, opened 2016-09-18
Enable background start of Cordova Android apps

macdonst:
This is to allow a push notification to tell the app to start in the background.


fredgalvao:
@macdonst Hmm, I think I found something strange.

This is still open, even though the code is on master. Commit https://github.com/apache/cordova-android/commit/a4f6c8bf7bf5ba49855f31831662f7964b17f962 "closes" it, but it mentions the wrong issue number (#322 vs #332).

PR #333, opened 2016-09-19
CB-11869 Fix cordova-js android exec tests

daserge:
Useragent Regexp result is null in node environment - add a check for it
Updated cordova-js

[CB-11869](https://issues.apache.org/jira/browse/CB-11869)


vladimir-kotikov:
LGTM

PR #335, opened 2016-09-25
older android fix (v2.3.6)

coderofsalvation:
I was playing around with my old android phone (v2.3.6), which did seem to install fine.
However, the app wouldn't launch.
`adb logcat` printed a nullpointerexception, which i fixed in this commit.
I also noticed more people had the same issue on stackoverflow, so i posted the patch there as well: 

http://stackoverflow.com/questions/19350062/java-lang-nosuchmethoderror-android-webkit-cookiemanager-setacceptfileschemecoo/39671863#39671863

The `// REALLY`-comment above the`setAcceptFileSchemeCookies`-call might indicate a bugfix which wasn't tested on older devices.

### DIFF
--- a/framework/src/org/apache/cordova/engine/SystemCookieManager.java
+++ b/framework/src/org/apache/cordova/engine/SystemCookieManager.java
@@ -37,10 +37,9 @@ class SystemCookieManager implements ICordovaCookieManager {
         webView = webview;
         cookieManager = CookieManager.getInstance();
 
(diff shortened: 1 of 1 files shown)


infil00p:
Cordova dropped support for Android 2.3.6 about two years ago:
https://cordova.apache.org/docs/en/latest/guide/platforms/android/

coderofsalvation:
ok we could just close this PR, at least the case is documented for and accessible by google.

PR #336, opened 2016-09-26
CB-11907: Upgrading the Gradle Plugins so they work with Android Studio 2.2

infil00p:
This is a PR to run against our CI before I put these into master.


alsorokin:
Looks like updating the gradle has introduced a nasty bug:
https://issues.apache.org/jira/browse/CB-11964

PR #337, opened 2016-09-29
CB-11935 Call onPause & onResume on WebView

spirann:
Does a best-effort attempt to pause any processing that can be paused safely, such as animations and geolocation.
Fix the potential rejection from Play Store

### DIFF
--- a/framework/src/org/apache/cordova/engine/SystemWebViewEngine.java
+++ b/framework/src/org/apache/cordova/engine/SystemWebViewEngine.java
@@ -315,8 +315,10 @@ public class SystemWebViewEngine implements CordovaWebViewEngine {
     @Override
     public void setPaused(boolean value) {
         if (value) {
+            webView.onPause();
             webView.pauseTimers();
         } else {
+            webView.onResume();
             webView.resumeTimers();
         }
     }


infil00p:
BTW: Don't add a commit that says "This closes #337" on this.  I want to be certain that this actually closes it and sign off on the commit.

vinzscam:
The #337 has broken the firing of `pause` event. 

Now the `pause` event fires when the native platform pulls the application out from the background (just before firing the `resume` event).

Maybe is better to rollback the changes keeping consistent to the [documentation](https://cordova.apache.org/docs/en/latest/cordova/events/events.html#pause).

infil00p:
Sorry, but this is more important than the pause event, this stays.

vinzscam:
It makes sense :)
I just want to point out that it is not just a broken API: it could affects the functionality of existing apps (in my case I was using it to connect/disconnect a persistent connection to a server).

infil00p:
It's obviously a bug, and we should probably make sure the pause event fires before we pause the timers somehow, since it's clear there's a threading issue.  That said, broken apps in the store are better than no apps in the store.

infil00p:
https://issues.apache.org/jira/browse/CB-12601 is the bug for this new issue.

dpogue:
The Android System WebView (in Android 5.0+ anyhow) also fires `visibilitychange` when the app is focused/backgrounded, which might be an alternative (and is, arguably, the spec-compliant event to use for this sort of thing).

Unfortunately the CrossWalk webview does not fire `visibilitychange`, nor do the older WebViews on Android 4.x, so it's not a perfect solution

PR #338, opened 2016-09-29
CB-11640: Fixing the checks so they check for Java 1.8 and higher FINAL VERSIONS

infil00p:
Right now, we test for only final and not pre-release versions of the Java language.  Therefore, no Java 9 early candidates.


infil00p:
Seriously, I can't even find info on Java 9 working with Android, and jenv is not letting me switch between Java 8 and Java 9, so it's definitely not ready for primetime. :/

PR #339, opened 2016-10-06
CB-11964: Do a clean when installing a plugin to get around the bug

infil00p:
This is a quick fix for the problem, and it might fix some other build errors.  I'm not sure about doing it here, because it might break more stuff, but hey, that's what PRs and CI are for. :)

### DIFF
--- a/bin/templates/cordova/Api.js
+++ b/bin/templates/cordova/Api.js
@@ -360,7 +360,6 @@ Api.prototype.run = function(runOptions) {
  */
 Api.prototype.clean = function(cleanOptions) {
     var self = this;
-    try {
     return require('./lib/check_reqs').run()
       .then(function () {
           return require('./lib/build').runClean.call(self, cleanOptions);
@@ -368,16 +367,6 @@ Api.prototype.clean = function(cleanOptions) {
       .then(function () {
           return require('./lib/prepare').clean.call(self, cleanOptions);
       });
-    }
-    catch (err) {
-      return require('../../lib/check_reqs').run()
-      .then(function () {
-          return require('./lib/build').runClean.call(self, cleanOptions);
-      })
-      .then(function () {
-          return require('./lib/prepare').clean.call(self, cleanOptions);
-      });
-    }
 };
 
 /**
--- a/bin/templates/cordova/Api.js
+++ b/bin/templates/cordova/Api.js
@@ -354,13 +354,24 @@ Api.prototype.run = function(runOptions) {
  */
 Api.prototype.clean = function(cleanOptions) {
     var self = this;
+    try {
     return require('./lib/check_reqs').run()
-    .then(function () {
-        return require('./lib/build').runClean.call(self, cleanOptions);
-    })
-    .then(function () {
-        return require('./lib/prepare').clean.call(self, cleanOptions);
-    });
+      .then(function () {
+          return require('./lib/build').runClean.call(self, cleanOptions);
+      })
+      .then(function () {
+          return require('./lib/prepare').clean.call(self, cleanOptions);
+      });
+    }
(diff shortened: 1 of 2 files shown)


infil00p:
OK, so, WTF was that? It appears that we have a test suite problem.

bin/templates/cordova/Api.js
```diff
@@ -212,6 +212,11 @@ Api.prototype.addPlugin = function (plugin, installOptions) {
       installOptions.android_studio = true;
     }
 
+    //CB-11964: Do a clean when installing the plugin code to get around
+    //the Gradle bug introduced by the Android Gradle Plugin Version 2.2
+    //TODO: Delete when the next version of Android Gradle plugin comes out
+    
+    this.clean();
```
alsorokin: Is it out of promise chain intentionally? I think that `this.clean()` returns a promise.

I've tried to merge it to the promise chain, but got the following exception upon `cordova platform add`:
```
Error: ENOENT: no such file or directory, open 'C:\Cordova\foo\platforms\android\assets\www\cordova_plugins.js'
    at Error (native)
    at Object.fs.openSync (fs.js:549:18)
    at Object.fs.writeFileSync (fs.js:1156:15)
    at C:\Cordova\foo\platforms\android\cordova\node_modules\cordova-common\src\PluginManager.js:137:12
    at _fulfilled (C:\Cordova\foo\platforms\android\cordova\node_modules\q\q.js:834:54)
    at self.promiseDispatch.done (C:\Cordova\foo\platforms\android\cordova\node_modules\q\q.js:863:30)
    at Promise.promise.promiseDispatch (C:\Cordova\foo\platforms\android\cordova\node_modules\q\q.js:796:13)
    at C:\Cordova\foo\platforms\android\cordova\node_modules\q\q.js:857:14
    at runSingle (C:\Cordova\foo\platforms\android\cordova\node_modules\q\q.js:137:13)
    at flush (C:\Cordova\foo\platforms\android\cordova\node_modules\q\q.js:125:13)
```

I wonder why it works the way you put it and if it is dangerous to leave it like this.

bin/templates/cordova/Api.js
```diff
@@ -349,13 +354,24 @@ Api.prototype.run = function(runOptions) {
  */
 Api.prototype.clean = function(cleanOptions) {
     var self = this;
+    try {
```
alsorokin: I think instead of doing this `try catch` in product code we could just mock `Api.prototype.clean()` in tests. Please see my attempt at doing so:
https://github.com/alsorokin/cordova-android/commit/008f46a1f3e807ee0ec1b576cd1a5d84e779d838

bin/templates/cordova/Api.js
infil00p: I agree it's dangerous, but I can't even get it to clean inside the promise chain. If you can try and get this to work in a safer way, that would be awesome.

bin/templates/cordova/Api.js
infil00p: That is a better solution.

infil00p:
Moving the example into core, tested it locally, everything appears to check out on MacOS.  Hopefully there's no weird Windows issues.

infil00p:
OK, there's something wrong with how we're cleaning in the chain.  I think we need to do it later in the promise chain.

infil00p:
I'm not getting the same errors locally when I run the same tests, could this be a weird travis timeout issue?

infil00p:
@alsorokin I'm going to keep the commits exploded and as is on this PR.  If this looks good to you, I'll merge this in tonight.

alsorokin:
Tested it on Windows, it works! LGTM

PR #340, opened 2016-10-06
CB-8722 - Move app icons from drawable to mipmap

dpogue:
https://issues.apache.org/jira/browse/CB-8722

Reasoning behind the change: https://androidbycode.wordpress.com/2015/02/14/goodbye-launcher-drawables-hello-mipmaps/

/cc @infil00p 
This probably counts as a breaking change, so we might want to get this into 6.0


infil00p:
If someone knew nothing about Android, and they just fiddled with their icons, would this change, or would we have to delete their old drawable for this to work?

dpogue:
If they were specifying icons via config.xml, this should just work.
If they had the platform already installed and had modified files, this shouldn't affect those existing files.

AxelNennker:
Does not work for me. Steps to reproduce are in the CB-8722

ghenry22:
I have the same problem, after upgrading the platform I now get default cordova icons for my android build, even though with the previous version of the platform all icons and splash screen images were in place and all specified in config.xml.

I use ionic framework to auto generate all splash / icon images and insert them into the config.xml

dpogue:
That is a known issue with cordova-android@6.0.0, unrelated to this change.

It has been fixed in master, and will be included in cordova-android@6.1.0 which is due to be released this week or next.

PR #357, opened 2017-02-03
CB-12421: (android)The CoreAndroid's loadUrl method should load web p…

wlmzfx:
### Platforms affected
Android

### What does this PR do?
Upate CoreAndroid's loadUrl method to load web pages in the UI thread

Built and run the generated application, calling {color:red}navigator.app.loadUrl('http://domain.com', { "clearHistory": true}){color} , verify that there is not more errors and navigation can be made successfully.

### Checklist
- [x] Commit message follows the format: "CB-12421: The CoreAndroid's loadUrl method should load web pages in the UI thread"
- [x] Added automated test coverage as appropriate for this change. (don't know how?)

### DIFF
--- a/framework/src/org/apache/cordova/CoreAndroid.java
+++ b/framework/src/org/apache/cordova/CoreAndroid.java
@@ -201,7 +201,16 @@ public class CoreAndroid extends CordovaPlugin {
                 e.printStackTrace();
             }
         }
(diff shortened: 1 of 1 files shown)


infil00p:
I'm pretty sure this is already being done on the UI thread.

wlmzfx:
I think this may not an issue when working with the system default webview, I met this issue when using the crosswalk together with cordova.

PR #358, opened 2017-02-03
CB-12424: (android) Fix encoding of multipart messages.

kalliste:
Currently when I try to assemble a multipart message cordova-android assembles the data in NativeToJsMessageQueue.java as [""] but if I call pMessageLoopResponse(callbackContext, 3, 5) it should generate [3, 5]

    private void pMessageLoopResponse(final CallbackContext ctx, int iMsg, int iTag) {
        List<PluginResult> results = new ArrayList<PluginResult>();
        results.add(0, new PluginResult(PluginResult.Status.OK, iMsg));
        results.add(1, new PluginResult(PluginResult.Status.OK, iTag));
        PluginResult result = new PluginResult(PluginResult.Status.OK, results);
        result.setKeepCallback(true);
        ctx.sendPluginResult(result);
    }

We split off encodeMessageAsJsMessage as its own method to support multipart messages including ones containing array buffer, binary string, or other multipart messages.


infil00p:
Can we have better method names?  encodeJsMessage and encodeMessageAsJsMessage looks confusing.

kalliste:
Sure! I can edit this and submit a new pull request --- What would be least confusing to you?

shankari:
Note that the multi-part handling appears to be a regression.
When I was on android@5.1.1, the email composer plugin, which returns a multi-part result from `isAvailable` worked. When I upgraded to android@6.1.2, it stopped working.
https://github.com/shankari/cordova-plugin-email-composer/blob/master/src/android/EmailComposer.java#L112

```
                List<PluginResult> messages = new ArrayList<PluginResult>();

                messages.add(new PluginResult(PluginResult.Status.OK, available[0]));
                messages.add(new PluginResult(PluginResult.Status.OK, available[1]));

                PluginResult result = new PluginResult(
                        PluginResult.Status.OK, messages);
```

Note further that multipart messages seem to be handled just fine in `encodeAsMessageHelper` (and, by implication, in `encodeAsMessage`) but not in `encodeAsJsMessage`. Is this due to the switch in the default bridge?

infil00p:
@kalliste Perhaps buildJsMessage, or convertToJsMessage?  You don't need to create a new PR, just add a commit to this branch, and the change should propogate through.

infil00p:
@kalliste BTW: Have you signed the ICLA with Apache? https://www.apache.org/licenses/#clas

kalliste:
I have now submitted an ICLA and also an updated pull request with a more reasonable method name.


> Is this due to the switch in the default bridge?

That is the impression I got.

infil00p:
@kalliste Thanks for submitting the ICLA.  Is it possible for you to add a Unit Test to the new Unit Testing framework for this patch? People have been complaining about this being a regression despite the fact that we have zero documentation and zero tests.  (In fact the refactored bridge had absolutely no unit tests until now, which is why I haven't pulled this in yet and have been working on getting some tests up.)  The test project is in tests, and it's a standard Android project using JUnit and Espresso.

kalliste:
Ok - I've added a unit test for this now

infil00p:
@kalliste This test is based on the old test framework.  Can you pull down the latest and take a look at this test and see if you can fit it in here?  https://github.com/apache/cordova-android/blob/master/test/app/src/test/java/org/apache/cordova/unittests/NativeToJsMessageQueueTest.java

kalliste:
Ok - I've pulled the new code and updated this test for it.

I have not yet found a gradle task in the new version that will show the results of the tests.

PR #362, opened 2017-02-22
CB-12453: Remove unnecessary double quotes from .bat files which are the causes of crash if project path contains spaces

matrosov-nikita:
### Platforms affected
self

### What does this PR do?
This PR fixes bug with crash if project path contains spaces

Current auto-tests don't work if project path contains spaces, this PR fixes it

### Checklist

### DIFF
--- a/bin/android_sdk_version.bat
+++ b/bin/android_sdk_version.bat
@@ -18,7 +18,7 @@
 @ECHO OFF
 SET script_path="%~dp0android_sdk_version"
 IF EXIST %script_path% (
-        node "%script_path%" %*
+        node %script_path% %*
 ) ELSE (
     ECHO.
     ECHO ERROR: Could not find 'android_sdk_version' script in 'bin' folder, aborting...>&2
--- a/bin/check_reqs.bat
+++ b/bin/check_reqs.bat
@@ -18,7 +18,7 @@
 @ECHO OFF
 SET script_path="%~dp0check_reqs"
 IF EXIST %script_path% (
-        node "%script_path%" %*
+        node %script_path% %*
 ) ELSE (
     ECHO.
     ECHO ERROR: Could not find 'check_reqs' script in 'bin' folder, aborting...>&2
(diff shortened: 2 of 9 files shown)


infil00p:
@filmaj Can you look at this?

filmaj:
Yep, will take a look.

I would feel better about merging in #391 first, as that would run the end to end tests on appveyor (windows-based CI), which would validate some of this stuff in an automated way rather than relying on manual testing.

I'm surprised removing quotes from a .bat file is the solution, but then again I don't know a lot about .bat syntax :P

filmaj:
@matrosov-nikita can you rebase with latest master in here?

filmaj:
I've merged in the e2e testing changes on Appveyor (Windows) into master, so a rebase should trigger running these changes in a Windows environment via our CI.

matrosov-nikita:
This bug affects user only if project path includes spaces.

Repro steps:
1) Create project in folder with space in name (i.e `C:\my projects\sample`)
2) Run `cordova run android --list`


On [this line](https://github.com/apache/cordova-android/blob/master/bin/templates/cordova/lib/list-devices.bat#L21) `%script_path%` already enclosed with double quotes. When we make additional enclosing, double quotes in the start and end of the script path are collapsed (""C:\my projects\sample\platforms\android\cordova\lib\list-devices"") and it leads to error like this: `Error: Cannot find module 'C:\my'`.

matrosov-nikita:
Regarding changes in `spec/e2e/helpers/projectActions.js`, script path within this [command](https://github.com/apache/cordova-android/blob/master/spec/e2e/helpers/projectActions.js#L48) is not enclosed with double quotes and if [platformpath](https://github.com/apache/cordova-android/blob/master/spec/e2e/helpers/projectActions.js#L42) is not defined, it affects `create.spec.js` and `update.spec.js`.

matrosov-nikita:
@filmaj, done.

PR #366, opened 2017-03-08
CB-12546: More robust support spawning the emulator with newer Android SDK

filmaj:
### Platforms affected

Android

### What does this PR do?

Two things related to starting emulator:

1. Adds more versatile support for the changing structure of the Android SDK. In particular, if spawning the `android` command exits with a non-zero status code (as it does in SDK Tools 25.3.1), attempts to use `avdmanager` and massages its output to conform to the old output of `android list avd`.
2. Ensures to spawn the emulator with a `cwd` parameter, to workaround the SDK bug described here: https://code.google.com/p/android/issues/detail?id=235461

Not enough! Unfortunately, I cannot downgrade my Android SDK to test that this works on older SDK versions. Thus why I attempted not to change any of the old logic leveraging the `android` binary, and rather just built support for the new SDK on top of the old one, trying to use feature / failure detection along the way.

I was, however, able to successfully spawn an emulator when running `./cordova run --emulator` from a generated Android project using the newest SDK tools.

### Checklist
- [X] [CB-12546](https://issues.apache.org/jira/browse/CB-12546)

### DIFF
--- a/bin/android_sdk_version
+++ b/bin/android_sdk_version
@@ -19,7 +19,7 @@
        under the License.
 */
 
-var android_sdk_version = require('./lib/android_sdk_version');
+var android_sdk_version = require('./lib/android_sdk');
 
 android_sdk_version.run().done(null, function(err) {
     console.log(err);
--- a/bin/lib/android_sdk.js
+++ b/bin/lib/android_sdk.js
@@ -62,3 +62,17 @@ module.exports.run = function() {
     return Q.all([get_sdks()]);
 };
 
+module.exports.version_string_to_api_level = {
+    '4.0': 14,
+    '4.0.3': 15,
+    '4.1': 16,
+    '4.2': 17,
+    '4.3': 18,
+    '4.4': 19,
+    '4.4W': 20,
+    '5.0': 21,
+    '5.1': 22,
+    '6.0': 23,
+    '7.0': 24,
+    '7.1.1': 25
+};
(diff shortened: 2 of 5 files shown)


filmaj:
Ah, I forgot to ping you: please review @infil00p

infil00p:
@filmaj This needs to at least pass Appveyor before I we should really look at it.  I'm sure it works, but jshint is showing a lot of quote errors.  (Related: I've been burned by this more times than I can count, maybe we should get rid of this check?)

filmaj:
@infil00p OK, I'll try to get to the bottom of that. I'll ping you once more when I do.

filmaj:
> Related: I've been burned by this more times than I can count, maybe we should get rid of this check?

Nah, I think it's a good check. It runs fast and it enforces a certain JS style - I think there's benefits to it. It was my bad to not run it.

filmaj:
Alright @infil00p, fixed that up. Please review, thanks!

filmaj:
OK I've rebased and tweaked based on some upcoming changes I'm prepping for CB-12554.

@infil00p can you review one more time, please?

infil00p:
This doesn't work in Windows, so there's still work to be done.

filmaj:
I see. I will take a look at it on my Windows VM.

filmaj:
OK, I had to add more robust support for automatically adding the location of the `avdmanager` and `sdkmanager` binaries to the PATH, and I broke the tests once more. At this point, I'm changing the behaviour so much that I don't feel comfortable pushing further updates to this without getting tests for this.

I am going to look at test coverage for this today.

filmaj:
I'm actually going to close this pull request and re-open once I have added tests and thoroughly tested on both Windows and Mac.

gonzaloserrano:
Any news on the fix? 

Got bugged when installing everything via Homebrew in MacOS, where the binaries are symlinked, like:
```
❯ realpath /usr/local/bin/emulator
/usr/local/Caskroom/android-sdk/3859397,26.0.1/emulator/emulator
```

PR #367, opened 2017-03-09
CB-12524: Working around Android SDK update by finding Gradle in Android Studio itself, if that fails, we look for a locally installed Gradle

infil00p:
### Platforms affected
Android

### What does this PR do?
This PR allows Cordova users to be able to build Android projects again using the platform CLI tools.  

Testing has been done on the build scripts to make sure things work, but testing on the CLI integration still needs to be done.

### Checklist
- [ x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ x] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.


infil00p:
Ah, how did that rogue commit get in there.  I'll be doing some git gardening before merging this, FYI

bin/templates/cordova/lib/builders/GradleBuilder.js
```diff
@@ -65,6 +65,20 @@ GradleBuilder.prototype.getArgs = function(cmd, opts) {
...
 };
 
+/*
+ * This returns a promise
+ */
+
+GradleBuilder.prototype.runGradleWrapper = function(gradle_cmd) {
+	var gradlePath = path.join(this.root, 'gradlew');
+	if(fs.existsSync(gradlePath)) {
+		;
```
filmaj: I think the spacing is off, too, yeah?

bin/lib/check_reqs.js
```diff
@@ -78,21 +79,46 @@ module.exports.check_ant = function() {
     });
 };
 
+module.exports.get_gradle_wrapper = function() {
+    var androidStudioPath;
+    if(os.platform() == "darwin") {
```
infil00p: I have no idea where Android Studio is installed on a Linux box.  It literally could be anywhere!  If you're not using Android Studio, you HAVE to have Gradle installed.  It's either one or the other from here on out.

bin/templates/cordova/lib/builders/GradleBuilder.js
infil00p: I spent an hour trying to get this stupid code to work in Windows.  This is so we don't keep running the wrapper task every time we build, clean or whatever else.  It looks weird because fs.existsSync kept forever returning true.

bin/lib/check_reqs.js
dpogue: A particular version of gradle, or just gradle in general?

Ubuntu gradle versions vary from release to release: http://packages.ubuntu.com/search?keywords=gradle
ArchLinux and homebrew (macOS) are both shipping the latest version (3.4)
Latest Android Studio ships 3.3

infil00p: @dpogue Honestly, we're just installing the wrapper, so any will work.  I wanted to delete https://github.com/apache/cordova-android/blob/master/bin/templates/cordova/lib/builders/GradleBuilder.js#L187 for a while, since it replaces everything with an old version of Gradle that just gets updated anyway, but we might just have to keep it in and update it to the latest Gradle supported.

FYI: My homebrew keeps fetching 2.2, and Android Studio 2.3 is packaged with 3.2 on my Windows and Mac partitions.

shazron:
Did tests on the old sdk (pre-G-eff-up), build was fine. Now upgrading my Android Studio and updating sdks

shazron:
Success! (Android Studio 2.2 -> Android Studio 2.3 upgrade, and update to latest sdk tools)

infil00p:
Tested against a Linode box with Android Tools updated to latest and Gradle Installed: (via http://exponential.io/blog/2015/03/30/install-gradle-on-ubuntu-linux/)

I have no idea how to get this to work on Travis, though!

mlaffargue:
Please we need this.

For a workaround you can download old version of Command Line Tools here : https://developer.android.com/studio/index.html#downloads

Then just rename in **Android SDK** your current _tools_ folder "tools_backup" and put the one in the zip instead.

At least could make _Ionic2 run_ work

infil00p:
So....I had to delete a test that no longer will ever pass thanks to the Android SDK tools, and the other problem we're running into now is that we DO in fact require Gradle to be higher than 2.14, and it doesn't matter when we update the wrapper, due to build.grade having the android plugin instantiated.

infil00p:
OK, it passes here: https://travis-ci.org/infil00p/cordova-android, so it's time to tidy this up and merge this in.

bin/lib/check_reqs.js
```diff
@@ -246,7 +273,11 @@ module.exports.check_android_target = function(originalError) {
     //   Google Inc.:Glass Development Kit Preview:20
     var valid_target = module.exports.get_target();
     var msg = 'Android SDK not found. Make sure that it is installed. If it is not at the default location, set the ANDROID_HOME environment variable.';
-    return tryCommand('android list targets --compact', msg)
+    //   Changing "targets" to "target" is stupid and makes more code.  Thanks Google!
+    var cmd = 'android list targets --compact';
+    if(forgivingWhichSync('avdmanager').length > 0)
+      cmd = 'avdmanager list target --compact';
```
filmaj: I just realized this but I think what we want here is to use `sdkmanager` and not `avdmanager`, right @infil00p ? `avdmanager list targets` returns either AVD device definitions or existing AVD IDs.

filmaj: Actually, don't worry about it, I already have a fix incoming anyways.

PR #370, opened 2017-03-21
CB-12595: Fix Android Studio Detection (may need refactoring)

infil00p:
### Platforms affected
Android

### What does this PR do?
Allows for the installation of multiple versions of Android Studio on a system without breaking the Cordova Build.

Tested it on local machine.  This is a critical fix for CB-12524 after realizing that future Android Studio versions will break that fix.

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/check_reqs.js
+++ b/bin/templates/cordova/lib/check_reqs.js
@@ -95,10 +95,31 @@ module.exports.check_ant = function() {
 
 module.exports.get_gradle_wrapper = function() {
     var androidStudioPath;
+    var i = 0;
+    var foundStudio = false;
+    var program_dir;
     if (module.exports.isDarwin()) {
-        androidStudioPath = path.join('/Applications', 'Android Studio.app', 'Contents', 'gradle');
+        program_dir = fs.readdirSync('/Applications');
+        while(i < program_dir.length && !foundStudio) {
+          if(program_dir[i].startsWith('Android Studio')) {
+            //TODO: Check for a specific Android Studio version, make sure it's not Canary
+            androidStudioPath = path.join('/Applications', program_dir[i], 'Contents', 'gradle');
+            foundStudio = true;
+          } else { ++i; }
+        }
     } else if (module.exports.isWindows()) {
-        androidStudioPath = path.join(process.env['ProgramFiles'],'Android', 'Android Studio', 'gradle');
+    var androidPath = path.join(process.env['ProgramFiles'], 'Android');
+        program_dir = fs.readdirSync(androidPath + '/');
+    console.log(path.join(process.env['ProgramFiles'], 'Android'));
+        while(i < program_dir.length && !foundStudio) {
+      console.log(program_dir[i]);
+          if(program_dir[i].startsWith('Android Studio')) {
+            foundStudio = true;
+            androidStudioPath = path.join(process.env['ProgramFiles'],'Android', program_dir[i], 'gradle');
+        console.log(androidStudioPath);
+          }
+          else { ++i; }
+        }
     }
 
     if (androidStudioPath !== null && fs.existsSync(androidStudioPath)) {
--- a/bin/templates/cordova/lib/check_reqs.js
+++ b/bin/templates/cordova/lib/check_reqs.js
@@ -110,13 +110,10 @@ module.exports.get_gradle_wrapper = function() {
     } else if (module.exports.isWindows()) {
     var androidPath = path.join(process.env['ProgramFiles'], 'Android');
         program_dir = fs.readdirSync(androidPath + '/');
-    console.log(path.join(process.env['ProgramFiles'], 'Android'));
         while(i < program_dir.length && !foundStudio) {
-      console.log(program_dir[i]);
           if(program_dir[i].startsWith('Android Studio')) {
             foundStudio = true;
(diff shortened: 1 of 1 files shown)


shazron:
Tested on macOS Sierra with Android Studio 2.3 with latest updates -- builds fine.

filmaj:
For reference, you can get canary and beta versions of the SDK here: https://developer.android.com/studio/preview/index.html

A summary of using the canary and regular versions of the SDK side by side is here: https://developer.android.com/studio/preview/install-preview.html

Sounds like we might want to expand the logic in here to cover Linux as well? (i.e. search the home directory). Otherwise, [Linux will rely on the right `gradle` being in the `PATH`](https://github.com/apache/cordova-android/pull/370/files#diff-3dc86ff94c734381565a43a0fe6bde98R132).

I'll try grabbing a Linux VM and with the SDKs installed side-by-side in Mac and Windows too. Will report back.

infil00p:
@filmaj The reason we didn't do the same thing for Linux is because:

1. Android Studio could literally be installed anywhere on Linux. (No, seriously.  The instructions for installing Canary are crazy vague)
2. Linux users know how to install Gradle on their platforms.
3. Linux users are an edge case, and they don't need the hand-holding that Windows and Mac users have come to expect.

filmaj:
Results of my testing with this.

# Mac

Side-by-side installations of Android Studio + Android Studio Preview:

```
➔ l /Applications/Android*
/Applications/Android Studio 2.4 Preview.app:
total 0
drwxr-xr-x@  3 maj   staff   102B 18 Mar 21:10 .
drwxrwxr-x+ 69 root  admin   2.3K 22 Mar 18:06 ..
drwxr-xr-x@ 12 maj   staff   408B 18 Mar 21:16 Contents

/Applications/Android Studio.app:
total 0
drwxr-xr-x@  3 maj   admin   102B 24 Feb 19:03 .
drwxrwxr-x+ 69 root  admin   2.3K 22 Mar 18:06 ..
drwxr-xr-x@ 12 maj   admin   408B 24 Feb 11:14 Contents```

Running the relevant method directly in node:

```
➔ node
> cr=require('./templates/cordova/lib/check_reqs')
> cr.get_gradle_wrapper()
'/Applications/Android Studio 2.4 Preview.app/Contents/gradle/gradle-3.4.1/bin/gradle'
```

Note that in my environment, `gradle` is not on my `PATH`:

```
➔ which gradle
gradle not found
```

# Windows

I installed the preview beside the existing Android Studio under `C:\Program Files\Android`:

```
C:\Program Files\Android>dir

03/22/2017  11:26 PM    <DIR>          .
03/22/2017  11:26 PM    <DIR>          ..
03/13/2017  06:12 PM    <DIR>          Android Studio
03/22/2017  08:24 PM    <DIR>          Android Studio Preview
```

Running the relevant function in node:

```
C:\src\cordova-android\bin>node
> cr = require('./templates/cordova/lib/check_reqs')
> cr.get_gradle_wrapper()
C:\Program Files\Android
Android Studio
C:\Program Files\Android\Android Studio\gradle
```

`gradle` is also not on the `PATH` in Windows:

```
C:\src\cordova-android\bin>where gradle
INFO: Could not find files for the given pattern(s).
```

So, it doesn't pick the preview version, just the regular Android Studio version - unlike on Mac. Not sure if that's a showstopper @infil00p. Also worth noting the extra output (`console.logs` in this code).

Still setting up my Linux VM so can't report back for that at the moment.

filmaj:
Thanks for cleaning up the logging. However, this point still stands:

> So, it doesn't pick the preview version, just the regular Android Studio version - unlike on Mac.

What's the desired behaviour? Pick Preview over Regular studio?

infil00p:
@filmaj I don't think it matters since we're overwriting whatever gradle-wrapper specifies with our own version of Gradle location anyway because we can't guarantee the version of Gradle (because of the Apache requirement to not bundle our own gradle-wrapper.jar).  All gradle has to do is install the gradle wrapper.

filmaj:
Got it, sounds good then.

PierBover:
I suspect this update is breaking code-push.

Hours ago I updated Android SDK tools to prevent a problem with USB and then had to `cordova platform add https://github.com/apache/cordova-android` since Cordova couldn't find graddle.

Now whenever I deploy to code-push the update can't be installed.

I'm investigating this and will report back.

filmaj:
Thanks for letting us know, @PierBover! We will keep our eyes peeled for your report.

PierBover:
When compiling locally and trying the `.apk` on a device everything works fine.

After deploying to code-push my users can't install the update. When trying the `.apk` in production (the one my users have installed) I can't update from code-push either.

I have explaind the problem in detail in a code-push issue since I'm guessing they will be able to point out what may be happening.

https://github.com/Microsoft/cordova-plugin-code-push/issues/238

infil00p:
@PierBover Based on the information provided in your issue, it's very unlikely that your issue was caused by this pull request.  The issue is related to compilation, not Javascript or Javascript bridge execution, which is what the log you have is currently showing.  I would recommend downgrading to the latest released version of Cordova and the Android Tools and seeing if the problem persists.

PierBover:
I really don't know @infil00p, it could be a combination of factors. I'll keep working on this and report back if I get new information.

PierBover:
I still don't know what is it, but I've changed the Android platform to a previous version and still getting the same problem.

Sorry for the noise guys.

PR #375, opened 2017-04-14
CB-12657 android:  fixed android 4.3 exit app, WebView.destroy() call…

Thomsen:
…ed while still attached

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/engine/SystemWebViewEngine.java
+++ b/framework/src/org/apache/cordova/engine/SystemWebViewEngine.java
@@ -28,6 +28,7 @@ import android.content.IntentFilter;
 import android.content.pm.ApplicationInfo;
 import android.os.Build;
 import android.view.View;
+import android.view.ViewGroup;
 import android.webkit.ValueCallback;
 import android.webkit.WebSettings;
 import android.webkit.WebSettings.LayoutAlgorithm;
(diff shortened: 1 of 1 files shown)


framework/src/org/apache/cordova/engine/SystemWebViewEngine.java
```diff
@@ -326,6 +329,12 @@ public void setPaused(boolean value) {
     @Override
     public void destroy() {
         webView.chromeClient.destroyLastDialog();
+        nativeToJsMessageQueue.reset();
+        ViewGroup vgParent = (ViewGroup) webView.getParent();
```
infil00p: Shouldn't we remove the webView's child views BEFORE removing the WebView?

framework/src/org/apache/cordova/engine/SystemWebViewEngine.java
```diff
@@ -326,6 +329,12 @@ public void setPaused(boolean value) {
     @Override
     public void destroy() {
         webView.chromeClient.destroyLastDialog();
+        nativeToJsMessageQueue.reset();
```
infil00p: Why are we resetting the nativeToJsMessageQueue on destroy?  We may want to send some events before we kill the WebView.

PR #378, opened 2017-05-02
Start of GradleBuilder.js specs

filmaj:
FYI @infil00p


infil00p:
I can't cleanly merge this, it looks like it's already in.

PR #385, opened 2017-06-11
CB-12835: add a Context getter in CordovaInterface

dreifachstein:
A custom engine may live outside of the Activity's lifecycle and the
Activity instance may not always be available. This getter allows
Context accesses in all cases.

### DIFF
--- a/framework/src/org/apache/cordova/CordovaInterface.java
+++ b/framework/src/org/apache/cordova/CordovaInterface.java
@@ -19,6 +19,7 @@
 package org.apache.cordova;
 
 import android.app.Activity;
+import android.content.Context;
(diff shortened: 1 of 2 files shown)


filmaj:
Does anything need to change in the plugin authoring documentation for Android as a result of this?

https://cordova.apache.org/docs/en/latest/guide/platforms/android/plugin.html

Maybe the Android Integration section needs updating? It references that the context is no longer available?

infil00p:
@filmaj Yeah, that definitely needs an update.

infil00p:
Yeah, this isn't that big of a deal to document, I'm going to merge this in now.

PR #386, opened 2017-06-14
CB-12895 : added eslint and removed jshint / reformatted with eslint 

audreyso:
### What does this PR do?

Added eslint and removed jshint / reformatted with eslint.

### Checklist
- [X] Added automated test coverage as appropriate for this change.


audreyso:
Thanks! @filmaj That makes sense and seems easier than having to add in /*eslint-env:  jasmine*/ in every file.  I made some changes and added .eslint configs where they are needed. Let me know if you have any other suggestions or feedback!

filmaj:
I think one more tweak would be good:

So by putting `root:true` into the spec/**/ directories' eslintrc files, when eslint searches up the directory tree to find an eslintrc file, once it finds the file, it'll stop searching further. eslint, however, supports "cascading" or combining multiple config files into one. We can leverage this to define one project-level eslintrc file in the root of the project dir, but then also have additional eslintrc files in project subdirectories if there are specific rules we want to apply per-directory.

So I think what we want to do is:
 - have the project-root-directory-level eslintrc file specify `root:true`
 - any subdirectory-level eslintrc files _not_ to specify `root:true`
   - this would then cause eslint to apply the rules from subdirectories, but keep going to search for further rules, and in our case, it would keep going until it hit the project-root-directory-level eslintrc, apply those rules, and because that one has `root:true`, stop there.
 - additionally, in this repo, both the spec/e2e and spec/unit directories need just a single additional rule: `env: jasmine: true`. So we could actually put a single eslintrc file under `/spec/.eslintrc.yml` and have it contain only the one `env: jasmine: true` rule. Since root is not set to true for the `spec/.eslintrc.yml` file, it'll go up one more directory and apply the project-level eslintrc.

audreyso:
Ohh thank you for that clarification @filmaj . So I did the following: 
1.  Left the project root level eslint file to keep--> root: true
2. Removed eslint config files from spec/e2e & spec/unit and just placed it in spec/.eslintrc.yml
3. Took out root:true from spec/.eslintrc.yml and just included ONE rule env: jasmine: true

filmaj:
Ahh! That's it! Very nice. Works flawlessy on my end!

images

infil00p:
LGTM

audreyso:
I rebased this morning! @filmaj @infil00p If anyone has time to merge it in, that would be great :)

filmaj:
Do you not have permissions to merge, or do you have your hands full with other work? Happy to merge this in for you but if it's a permission issue please let me know so that I can fix that :)

audreyso:
I think I have permission, but haven't ever done it before! @filmaj

filmaj:
I believe in you, you can do it!

i-have-hope

bin/templates/cordova/Api.js
```diff
@@ -112,16 +110,13 @@ Api.createPlatform = function (destination, config, options, events) {
     events = setupEvents(events);
     var result;
     try {
-        result = require('../../lib/create')
-        .create(destination, config, options, events)
-        .then(function (destination) {
+        result = require('../../lib/create').create(destination, config, options, events).then(function (destination) {
```
purplecabbage: This is an unfortunate workaround for eslint-ing.  I would much prefer this be broken up, ala ...

    var libCreate = require('../../lib/create');
    var result = libCreate.create(destination, config, options, events);
    result.then(function (destination) {
        // ...
    });
    ...
    return result;

PR #391, opened 2017-06-27
CB-12950: E2E test tweaks + Appveyor now runs E2E tests

filmaj:
Lots of changes to how tests are run. Biggest one is now Appveyor runs end-to-end tests on Windows.

I have one comment in [CB-12950](https://issues.apache.org/jira/browse/CB-12950) about how to speed up the testing by reducing the number of end-to-end `create` tests - I think a lot of them are redundant. We can combine all of the create tests into one to significantly reduce runtime (> 50%), but we'd lose the ability to distinguish which of unicode, ascii or spaces in the application name parameter passed to `create` cause an issue. Check out the detail link above for more info / discussion.

Please review @infil00p, FYI @purplecabbage, @alsorokin, @dpogue.


filmaj:
I just realized... on Windows, we run against both node v4 and node v6. On travis, only against node v6. We should make this consistent. Which one? The `package.json` file _does_ specify as supporting node v4+...

purplecabbage:
Yeah, iOS doesn't even care ...
https://github.com/apache/cordova-ios/blob/master/.travis.yml
Windows does 4+6 ( without `nvm` )
https://github.com/apache/cordova-windows/blob/master/.travis.yml
Browser also 4+6
https://github.com/apache/cordova-browser/blob/master/.travis.yml

filmaj:
Might be tough as we tell travis to use the Android image preset via `language: android` so it comes pre-installed with some bits from the Android SDK...

purplecabbage:
Yeah that might be the same issue with iOS ...

alsorokin:
Take a look at this: http://austinpray.com/ops/2015/09/20/change-travis-node-version.html
So, we're basically download nvm and pass it a node version stored in the environment variable.
I've already done it for the plugins repos (only node 4.2 so far though):
https://github.com/apache/cordova-plugin-device/blob/master/.travis.yml#L55

alsorokin:
As I've already done this for plugins and have the knowledge, I can do the same thing for iOS and Android if you like, after this is merged.

filmaj:
@alsorokin perfect! Yeah no worries, I can do it too. How about let's file an issue for it first and then go from there?

As for this particular PR, today I will attempt to start an emulator in appveyor + travis and run the native cordova-android instrumented tests. That would end up running _all_ the tests we have in cordova-android in CI, on both Windows and Linux, which would be a big win, I think.

Further, I'd like to direct people to the original issue thread where I commented about how we can speed up the integration tests (by removing a bunch of the create tests). Still looking for feedback on that - if you have an opinion (either for or against it), please speak up!

alsorokin:
Here's a JIRA issue:
https://issues.apache.org/jira/browse/CB-12960

infil00p:
LGTM

PR #392, opened 2017-06-27
WIP: Run native android unit tests in CI

filmaj:
Still a work in progress, let's see how appveyor and travis react to this.


infil00p:
Did this get merged in separately? Why is there an appveyor conflict?

filmaj:
Nah, this was a work-in-progress @infil00p. I think we can kill this. 

I do think we should salvage running the java unit tests from this PR, but we can do that in a new PR. I'll whip something up and ping you on that later.

filmaj:
Anyone still interested in this can take a look at #405

PR #393, opened 2017-07-11
Fix for Jira issue CB-12981

AndyClausen:
https://issues.apache.org/jira/browse/CB-12981

### Platforms affected
Android

### What does this PR do?
A method that checks API level now uses a regex instead of split and replace (the split returned null and therefore replace threw a null pointer exception)

Tested with Ionic framework on Ubuntu 16.04 with the latest version of Android Studio.

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/emulator.js
+++ b/bin/templates/cordova/lib/emulator.js
@@ -196,7 +196,7 @@ module.exports.best_image = function () {
         for (var i in images) {
             var target = images[i].target;
             if (target) {
-                var num = target.split('(API level ')[1].replace(')', '');
+                var num = target.match(/\d+/)[0];
                 if (num === project_target) {
                     return images[i];
                 } else if (project_target - num < closest && project_target > num) {


infil00p:
This somehow broke the project creation test.  Can you figure out what happened with your PR that caused this to break?

filmaj:
I think if the PR author rebases w/ latest master, they will find that test no longer exists / no longer applicable.

I bet it was Travis VM being too slow / our timeout for the create/compile loop being too low.

iget-master:
Someone on this? It's impossible to run cordova over android emulator with Android 8 AVD due this issue

filmaj:
I'd like to add a unit test here. I believe the Android SDK API levels / target listing (perhaps output from `android list target` or `sdkmanager`) is different, and the parsing fails. I need a repro case first to make sure that my assumption is right.

tensberg:
I have the same problem, hope that the following info helps to write a unit test.

With Android SDK Tools 26.0.2 the output of avdmanager is:

```
$ avdmanager list avd
Available Android Virtual Devices:
    Name: Nexus_5X_API_26_x86
  Device: Nexus 5X (Google)
    Path: /Users/koch/.android/avd/Nexus_5X_API_26_x86.avd
  Target: Google APIs (Google Inc.)
          Based on: Android API 26 Tag/ABI: google_apis/x86
    Skin: nexus_5x
  Sdcard: 800M
Snapshot: no
```

(The output of `android list avd` is the same since it just calls avdmanager.)

Based on this output, `list_images_using_avdmanager` sets `img_obj.target` to `Android API 26`. (And `list_images_using_android` leaves it as `undefined`.)

Perhaps it would be safer to change the parsing of the target property so that it produces the same output for different avdmanager/android versions instead of relaxing the regular expression?

infil00p:
@AndyClausen Can you send an empty commit to bump the Travis to make sure everything is working? I think there was confusion about @filmaj's comment on this PR, and if so, I would like to make sure this passes Travis before merging it in.

AndyClausen:
I'm on phone right now, maybe tomorrow.
This is such a simple fix though, I don't get the hassle.

filmaj:
@AndyClausen you may have submitted a simple fix, but we need a good definition of the problem. I'll repeat myself: we need a reproduction case, which, I think based on @tensberg helpful comment, we can put together. At that point we can write a unit test on this to ensure we support the output of _both_ the old Android AVD management tool (`android`) as well as the new one (`avdmanager`).

filmaj:
Just want to confirm @tensberg's conclusion, I can repro the same thing.

We'll need to change the code to be able to deal with both ways that the Android SDK describes  the API level of AVDs, using either the numeric API ID (i.e. Android API 26) or the Android version string (i.e. Android 8.0).

Since `list_images_using_xxx` is only used 'internally' in `emulator.js`, and all other scripts use `list_images` (which itself calls into either leveraging the `android` or the `avdmanager` binaries, depending on what's available), I suggest having `list_images` massage the parsed list of AVDs and fill in any missing info re: API level / Android version. I'll whip up a proof of concept and issue a new PR for that, and we can move discussion there.

I'll also add @tensberg's updated output above as a unit test case to ensure the new code works with both kinds of output.

filmaj:
FYI I've issued a new pull request in #406 for this. Closing this PR.

PR #394, opened 2017-07-11
CB-13006: replace create end-to-end tests with unit tests

filmaj:
There were create end-to-end tests that validated use of unicode, spaces and ascii characters in project names and destination paths. These tests were HEAVY - they would shell out to create _and_ compile each one, and there were 6 of these tests. Additionally, these tests would regularly time out on Travis.

Instead, this PR replaces those tests with lighter weight unit tests exercising the same logic (hopefully).

*NOTE*: by removing the create end-to-end tests, we now no longer test that created projects can compile using the Android SDK. We may want to consider adding a basic smoke-test for this back into the end-to-end tests.

Please review / discuss / FYI @infil00p, @stevengill, @dpogue, @alsorokin, @matrosov-nikita, @jcesarmobile


filmaj:
It seems like removing the create e2e tests, which regularly time out, just pushed the timeouts to the upgrade e2e tests. I believe it has something to do with the first test that compiles the project needing to kick up the gradle daemon and warm up the caches.

Will remove those tests as well - we've talked about doing that regardless.

filmaj:
This also sped up CI runtimes:

- AppVeyor: ~9 mins to ~4 mins
- Travis: ~12 mins to ~8 mins

PR #395, opened 2017-07-14
CB-12423: make explicit JDK 1.8 or greater is needed in the README.

filmaj:
Lining it up with https://cordova.apache.org/docs/en/latest/guide/platforms/android/index.html#installing-the-requirements

As per https://issues.apache.org/jira/browse/CB-12423


filmaj:
You cool with this change @infil00p ? Just making sure.

infil00p:
LGTM

PR #387, opened 2017-06-20
CB-12605 In Windows get Android studio path from the registry

purplecabbage:
### Platforms affected
Android on windows

### What does this PR do?
Allows installs of Android Studio in places other than C://

### What testing has been done on this change?
Tested with and without registry value present

### Checklist
- [x] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [x] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

PR #396, opened 2017-07-19
CB-12677: (android) added option for explicitly specifying path to An…

russaa:
### Platforms affected
android

### What does this PR do?
Adds the option for explicitly specifying the path to Android Studio via env variable.

Automatic detection of the _Android Studio_ path currently only works, if it is installed in the default location.

The option to specify the path via an env variable offers a simple solution, in cases where automatic detection does not work.

This solution relates to [issue & discussion CB-12677](https://issues.apache.org/jira/browse/CB-12677).

manually tested

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/check_reqs.js
+++ b/bin/templates/cordova/lib/check_reqs.js
@@ -99,7 +99,12 @@ module.exports.get_gradle_wrapper = function () {
     var i = 0;
     var foundStudio = false;
     var program_dir;
-    if (module.exports.isDarwin()) {
+    // CB-12677: check if Android Studio path was explicitly set via environment variable
+    if (process.env['ANDROID_STUDIO_HOME']) {
+
+        androidStudioPath = path.join(process.env['ANDROID_STUDIO_HOME'], 'gradle');
+
+    } else if (module.exports.isDarwin()) {
         program_dir = fs.readdirSync('/Applications');
         while (i < program_dir.length && !foundStudio) {
             if (program_dir[i].startsWith('Android Studio')) {
@@ -157,7 +162,8 @@ module.exports.check_gradle = function () {
     if (gradlePath.length !== 0) { d.resolve(gradlePath); } else {
         d.reject(new CordovaError('Could not find an installed version of Gradle either in Android Studio,\n' +
                                 'or on your system to install the gradle wrapper. Please include gradle \n' +
-                                'in your path, or install Android Studio'));
+                                'in your path, or install Android Studio, or set up \'ANDROID_STUDIO_HOME\'\n' +
+                                'env variable.'));
     }
     return d.promise;
 };


filmaj:
I'm not sure this is needed as we already do plenty of checking for the `ANDROID_HOME` environment variable. If you set that env var to the location of your Android SDK (whether or not it is the one bundled with Android Studio does not matter), that should be enough to help cordova figure out where the relevant Android tools are.

russaa:
it is for detecting `Android Studio`, i.e. for detecting `gradle` in cases where the already-built-in-detection does not work -- so `ANDROID_HOME` is not necessarily a help here, as it points to the SDK which could be at an entirely different location.

For example, when `Android Studio` is installed in some custom location, the already-built-in-detection fails, as described in the linked issue [CB-12677](https://issues.apache.org/jira/browse/CB-12677).

infil00p:
We are going to make Gradle a requirement.  Using Android Studio's copy of Gradle was a stop-gap solution to get builds working again, and we should really stop doing this.

russaa:
I have to say, I like the current solution better (using `gradle` from Android Studio):

 1. you are not required to install something, that you may already have installed
 2. if you are building/debugging the Cordova app with Android Studio, you can be sure that builds are same (as you are using the same version of `gradle`)
 3. users who prefer to install `gradle` separately still have this option

If you want to get rid of the automatic-detection (because it may become cumbersome to maintain) I would still like to have the option of using Android Studio's `gradle` by pointing to its place via the env variable.   
Its code should be fairly easy to maintain, since the user is responsible to specify the correct path, i.e. the detection routine only needs to look into the env variable and does not need to be adjusted to changing path structures etc. itself.  
Although the mechanism then should maybe be changed, so that the env variable needs to point the `gradle` directory directly, instead of the Android Studio directory (which it does now), e.g. change the env variable to something like `GRADLE_HOME` and the path would need to point, e.g. to the `gradle` sub-directory of the Android Studio directory.


As an additional reason:  
I assume the _"install `gradle` separately/as requirement"_ option also means, that you must have `gradle` in your PATH env variable.

On Windows, this is a concern for me, as the PATH variable has a maximum length -- which I, for instance, are always about to exceed. So I try to not have paths there, if I can avoid it.

infil00p:
Did you see the fix @purplecabbage did to pull this info from the Windows registry for Windows users? We're probably going to go with that solution for Windows because of the whole PATH issue.  Mac OS and Linux users are able to install Gradle without a problem.  There are already other issues with using or requiring Android Studio as a dependency, such as the fact that some people don't ever use the IDE

aminul-technext:
Guys I need this pr to be merged. I am developing on Ubuntu 16.04 and installed Android Studio. 
The problem is `check_reqs.js` never found my Android Studio.
And another thing, we think the Android will be the 'Application' directory, so if not then no body would understand whats the actual problem.
And about canary (Two Android Studio in Application directory Mac OS), I think we should inform users about that in the Documentation. In this case its throwing `EACCESS` which is just point user to different task.

russaa:
@infil00p I was not aware of this patch (I believe it is CB-12065 #387).

I guess it would solve the issue for Windows right now -- but does not address the PATH problem (for Windows) in future versions of `cordova-android`, when `gradle` becomes a requirement.

So in view of this future change, I would argue for the (additional) use of env variables (i.e. offer users to specify an env variable), because
 * it's basically platform independent, in difference to #387, which is fully automatic, but only works on Windows -- e.g. with env variables, if you have `gradle` and Android Studio installed on a *nix system, you can configure Cordova to use Android Studio's `gradle` quite easily
 * it's a low maintenance mechanism (i.e. implementation requires mostly no adjustments, when software paths/structures change), since users are responsible for providing the correct path

So I would very much like to have the option, to specify the `gradle` location via an env variable in the future (i.e. when `gradle` will be declared a requirement), instead of having to install `gradle` separately, and put it into the PATH variable.

infil00p:
Since 6.4.0 has been released, I'm going to close out this PR.  We can't rely on the permissions or the location of gradle to be present across major versions of Android Studio, so we can't rely on it.

PR #397, opened 2017-07-24
CB-13031: Fix bug with case-sensitivity of android-packageName

matrosov-nikita:
### Platforms affected
self

### What does this PR do?
Steps to reproduce:
1) cordova create testapp com.example.testapp testapp
2) cd testapp
3) edit config.xml, add android-packageName="com.example.Testapp" to <widget>
4) cordova platform add android

Since `windows` and `osx` file systems are case-insensitive, [this line](https://github.com/apache/cordova-android/blob/893356abcd1b005a13bb09e7467f227b52de3ae5/bin/templates/cordova/lib/prepare.js#L213) doesn't create a directory with a name from `android-packageName` attribute if it differs in casing from original package name. 

This PR makes case-sensitive check before removal of an old package.

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/prepare.js
+++ b/bin/templates/cordova/lib/prepare.js
@@ -24,6 +24,7 @@ var path = require('path');
 var shell = require('shelljs');
 var events = require('cordova-common').events;
 var AndroidManifest = require('./AndroidManifest');
+var checkReqs = require('./check_reqs');
 var xmlHelpers = require('cordova-common').xmlHelpers;
 var CordovaError = require('cordova-common').CordovaError;
 var ConfigParser = require('cordova-common').ConfigParser;
@@ -181,10 +182,10 @@ function updateProjectAccordingTo (platformConfig, locations) {
     events.emit('verbose', 'Wrote out android application name "' + name + '" to ' + locations.strings);
 
     // Java packages cannot support dashes
-    var pkg = (platformConfig.android_packageName() || platformConfig.packageName()).replace(/-/g, '_');
+    var androidPkgName = (platformConfig.android_packageName() || platformConfig.packageName()).replace(/-/g, '_');
 
     var manifest = new AndroidManifest(locations.manifest);
-    var orig_pkg = manifest.getPackageId();
+    var manifestId = manifest.getPackageId();
 
     manifest.getActivity()
         .setOrientation(platformConfig.getPreference('orientation'))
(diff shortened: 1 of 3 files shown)


filmaj:
Just for the record, here's the full stack trace I get when I try these steps:

```
~/src
➔ cordova create testapp com.example.testapp testapp
Warning: using prerelease version 7.0.2-dev (cordova-lib@7.0.1)

~/src
➔ cd testapp

~/src/testapp
➔ vim config.xml

~/src/testapp
➔ cordova platform add android
Warning: using prerelease version 7.0.2-dev (cordova-lib@7.0.1)
Using cordova-fetch for cordova-android@~6.2.2
Adding android project...
	Path: platforms/android
	Package: com.example.testapp
	Name: testapp
	Activity: MainActivity
	Android target: android-25
Subproject Path: CordovaLib
Android project created with cordova-android@6.2.3
Discovered plugin "cordova-plugin-whitelist" in config.xml. Adding it to the project

               This plugin is only applicable for versions of cordova-android greater than 4.0. If you have a previous platform version, you do *not* need this plugin since the whitelist will be built in.

Saved plugin info for "cordova-plugin-whitelist" to config.xml
ls: no such file or directory: /Users/maj/src/testapp/platforms/android/src/com/example/Testapp/*.java

Error: No Java files found that extend CordovaActivity.
```

This patch seems to fix that. However, I noticed that the output even using this patch still says:

```
	Package: com.example.testapp
```

Can we update the PR to also print out the correct, overridden package name, if provided?

bin/templates/cordova/lib/prepare.js
```diff
@@ -214,7 +215,11 @@ function updateProjectAccordingTo (platformConfig, locations) {
     shell.sed(/package [\w\.]*;/, 'package ' + pkg + ';', java_files[0]).to(destFile);
     events.emit('verbose', 'Wrote out Android package name "' + pkg + '" to ' + destFile);
 
-    if (orig_pkg !== pkg) {
+    var removeOrigPkg = checkReqs.isWindows() || checkReqs.isDarwin() ?
+        orig_pkg.toUpperCase() !== pkg.toUpperCase() :
+        orig_pkg !== pkg;
```
filmaj: I know the naming of these variables didn't originate from you, but while we're here, can we also update the variable names used here? I found `orig_pkg` and `pkg` to be not very descriptive. Maybe `manifest_id` instead of `orig_pkg`?

matrosov-nikita:
@filmaj, updated. But I little doubt about folders names after prepare, this PR just skips the removal of an original package if names (manifestId and androidPkgName) differ in casing only. So, `platforms\android\src\com\example\testapp` is still in the lower case despite appropriate changes in the manifest.

matrosov-nikita:
@filmaj, thanks for the information, I'm planning to merge in once Appveyor is fixed.

filmaj:
Hey @matrosov-nikita, rebase w/ latest master, I should have a fix (workaround) in for appveyor troubles.

PR #404, opened 2017-08-29
Update maven repo to include most recent lib versions

kanongil:
### What does this PR do?

Fix install issues with 25.4.0+ google support libraries. As it is, they are not installable, due to move to new repo.

See note at https://developer.android.com/topic/libraries/support-library/revisions.html#25-4-0 for details:

> Important: The support libraries are now available through Google's Maven repository. You do not need to download the support repository from the SDK Manager.

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/plugin-build.gradle
+++ b/bin/templates/cordova/lib/plugin-build.gradle
@@ -20,8 +20,10 @@
 
 buildscript {
     repositories {
-        mavenCentral()
         jcenter()
+        maven {
+            url "https://maven.google.com"
+        }
     }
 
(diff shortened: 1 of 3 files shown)


infil00p:
It doesn't look like it's blowing anything up.  Is there any good reason to use maven and not just use jCenter though?

kanongil:
This is the recommended setup from google. Since the latest versions of their support libraries are only published to the google maven repo, is is required.

Afaik, it is safe to remove the old mavenCentral source.

infil00p:
@macdonst What are your thoughts on this?

macdonst:
@infil00p it looks okay to me.

PR #405, opened 2017-09-13
Running Java unit tests more easily

filmaj:
### What does this PR do?

Trying to make it easy for both humans and our CI systems to run the Java unit tests. In particular, I wanted to seamlessly handle building the gradle wrapper and invoking `gradlew test` on both *nix and Windows systems.

Changes include:
- added a node script in the `test/` project directory, `run_java_unit_tests.js`, that wraps up building the gradle wrapper for the test project and invoking the unit tests (`gradlew test`)
- wrapped up running that script in the `npm run java-unit-tests` helper
- include `npm run java-unit-tests` in the top-level `npm test` script
- now runs java unit tests in travis and appveyor
- small refactor in `gradlebuilder.js` to support building the gradle wrapper more easily - kept the defaults and behaviour of gradlebuilder functions as before, just parameterized the .gradle file as part of the `runGradleWrapper` helper method.
- don't explicitly build gradlew on appveyor CI as now `npm test` will do it for you

I tested this on both Windows 10 and Mac OS 10.12, in both the fresh-clone and fully-loaded cordova-android repo cases. That is, in the fresh-clone situation, `npm run java-unit-tests` will a) download the proper gradle version, b) build the gradle wrapper and finally c) run the java unit tests. The fully-loaded case will skip right to c).

# Worth Noting

I dropped a TODO regarding the flow of promises from `GradleBuilder` - the helper method `runGradleWrapper` returns a promise in the case that it needs to build the gradle wrapper, but if the gradle wrapper is already built, it returns nothing. So, I had to handle this discrepancy in the `run_java_unit_tests.js` file. For cleanliness, it may be worth having the `runGradleWrapper` method always return a promise. I'm happy to make those changes if we deem it appropriate after a review.

Please review (or at least, FYI): @infil00p, @alsorokin, @stevengill, @dpogue, @macdonst, @purplecabbage, @jcesarmobile

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,17 +2,25 @@ language: android
 sudo: false
 jdk:
     - oraclejdk8
+env:
+    global:
+      - ANDROID_TOOLS=${ANDROID_HOME}/tools
 before_install:
     - nvm install 6
+    # ensure at least gradle 3.3 is in place.
+    - wget http://services.gradle.org/distributions/gradle-3.3-bin.zip
+    - unzip gradle-3.3-bin.zip
+    - export GRADLE_HOME=$PWD/gradle-3.3
+    - export PATH=${GRADLE_HOME}/bin:${ANDROID_HOME}:${ANDROID_HOME}/emulator:${ANDROID_TOOLS}:${ANDROID_TOOLS}/bin:${ANDROID_HOME}/platform-tools:$PATH
     - node --version
     - gradle --version
-install:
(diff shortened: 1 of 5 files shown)


filmaj:
Welp looks like Travis pooped itself. But `pushd` not found? wat

```
> pushd test && node run_java_unit_tests.js && popd;
sh: 1: pushd: not found
```

PR #406, opened 2017-09-13
CB-12981: handle slightly different AVD list output in Android SDK 26.0.2

filmaj:
Would cause "cannot read property replace of undefined" errors when trying to deploy an Android 8 emulator.

This supersedes #393.


filmaj:
The appveyor failure here is expected until the fix present in #405 lands in master. Once that lands, I can rebase this branch and get CI passing for this PR.

PR #423, opened 2017-12-29
Strip starting dot from ActivityName

Mitko-Kerezov:
### Platforms affected
Android

### What does this PR do?
Strip starting dot from ActivityName. Activity's name may already contain a leading dot symbol in the AndroidManifest.xml. If this is the case it has to be stripped.

1. `cordova create myApp`
1. `cd myApp`
1. `cordova platform add <this branch>`
1. `sed -i s/MainActivity/.MainActivity/ platforms/android/AndroidManifest.xml`
1. `cordova run android`
With the `master` branch `cordova` fails to launch the application and with this branch it succeeds. 

### Checklist
- [x] [Reported the issue](https://issues.apache.org/jira/browse/CB-13720) in the JIRA database


infil00p:
This breaks compatibility and may break every other application, not going to accept this PR at this time.

PR #434, opened 2018-03-19
CB-13830: Add handlers for plugins that use non-Java source files

infil00p:
### Platforms affected
Android

### What does this PR do?
This PR adds code that handles the uninstallation of plugins that use XML files for provider paths.  This further segments the re-mapper code, that we eventually will delete.

I also fixed some tabs, because my eslint was puking.

- Ran through unit tests
- Tested on project.

### Checklist


jcesarmobile:
Not strictly related to this PR, but for the whole Eclipse/Android Studio detection that is causing so much trouble with some plugin installs.
Why do we check for eclipse project? If we create the project with latest version, it creates an Android Studio project, so why do we have that check? Is it possible to create an Eclipse project somehow?
Can't we just use the new location for all the source-file tags and remove the Eclipse structure checking?

infil00p:
@jcesarmobile We check because we haven't migrated the core plugins to work with the new project yet and we need legacy plugins to still be able to be installed.  If we don't have this check, all the plugins will fail.

jcesarmobile:
What I mean is the first check where we look for AndroidManifest.xml, res and lib folder on the root of the project to see if it's an Eclipse project.
Can't we remove that check, but keep the mapping to the app/src/main if the target doesn't include it? and also do this mapping for other file types, not just .java and .xml.

As we don't have that mapping, what is happening is that old plugins are creating res and/or lib folders on the root, so the tooling thinks it's an Eclipse project and when it tries to install the next plugin it fails to install

infil00p:
I just made sure that this worked for install on our core plugins.  If people are moving files beyond the core plugins, I'm not sure what to say about that.  We could remove it, but I think that would be another breaking change.

jcesarmobile:
To be clear, I'm talking about this
https://github.com/apache/cordova-android/blob/master/bin/templates/cordova/lib/AndroidStudio.js#L13

If we are creating Android Studio projects, and it's not possible to create Eclipse projects, why do we do that check? That is just doing false positives when a plugin creates a libs or res folder.

I'm ok with this mapping https://github.com/apache/cordova-android/pull/434/files#diff-4f73e5f70cec312ae82aa7d023e3c88bR322 where you check if the target dir includes app/src/main and if not remap it to the new Android Studio structure.

infil00p:
Yeah, we could kill that and make it always return true.  It's in there because originally I was going to make this upgrade old projects without messing with their project structure until I was told that iOS doesn't do this.

infil00p:
@jcesarmobile But back to the PR, is this cool to add?

jcesarmobile:
Yeah, sorry, looks good, but I wanted to test it this weekend

PR #435, opened 2018-03-23
CB-13975: (android) Fix to fire pause event when cdvStartInBackground=true

richturner:
### Platforms affected
Android

### What does this PR do?
When the app is started with cdvStartInBackground=true then a pendingPause event is cached until the messageChannel is initialised (before the pause event never reached the webview and it wasn't possible to determine in JS code that the app was started in the background), this change results in the following JS events inside the app:

- pause
- deviceready

As a user you can then store a flag (e.g. app.inBackground) and this can be checked in 'deviceready' handler.

Only manually tested on several Android phones.

### Checklist


infil00p:
LGTM

PR #437, opened 2018-04-18
CB-14038 (android): fix false positive detecting project type

jcesarmobile:
### Platforms affected
Android

### What does this PR do?
Make the Android Studio check return true
Removed old android project structure tests and files

### Checklist


infil00p:
It's probably safe to delete GradleBuilder.js since it should never run now that the isAndroidStudio is always true.

jcesarmobile:
Yeah, but if I remove the file I should also remove the tests related to the builder, and as it's always true I can remove all the checks, but wanted to keep it simple as a fast fix and do the major refactor in the future

macdonst:
Wouldn't it make more sense to delete `AndroidStudio.js` as it is only used in a couple of places.

Then remove the if statement from here:

https://github.com/apache/cordova-android/blob/55d7cf38654157187c4a4c2b8784191acc97c8ee/bin/templates/cordova/lib/AndroidProject.js#L66

and only set the www directory once to the AndroidStudio path.

Then move all the assignments in this if statement:

https://github.com/apache/cordova-android/blob/83686542b2bbcf481b6b3ee709a84802d48df670/bin/templates/cordova/Api.js#L82

to the lines above so they are the defaults.

raphinesse:
@jcesarmobile Is there any WIP for the aforementioned refactor? Losing some dead weight would definitely be good. Furthermore, before I discovered this PR by accident, I already thought about starting to DRY the two very similar Builders, not knowing that removal of one of them is due.

BTW: GradleBuilder is used by the Java unit test runner. Just to create a Gradle wrapper, so the dependency could be easily removed.

jcesarmobile:
No, there is no work about it because the idea was to do a patch or minor release with this changes to unblock people with plugin problems, and the refactor was supposed to be done for next major release. But right now it's not clear to me if we are going to do a patch, minor or major release.

Anyway, the refactor is not a breaking change, so it can be done at any time.

raphinesse:
Thanks for your answer.

PR #438, opened 2018-04-25
Add support for EnableGoogleServicesPlugin

chemerisuk:
### Platforms affected
Android

### What does this PR do?
@macdonst [proposed](https://lists.apache.org/thread.html/264f66968b5b3a02533b0de1fbf887cba82efd8601e502476e5c8eee@%3Cdev.cordova.apache.org%3E) to add new preference to the platform, that runs google services plugin.

Checked locally.

### Checklist

### DIFF
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -30,6 +30,7 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:3.1.0'
+        classpath 'com.google.gms:google-services:3.2.1'
     }
 }
 
@@ -321,4 +322,8 @@ for (def func : cdvPluginPostBuildExtras) {
 //     ext.postBuildExtras = { ... code here ... }
 if (hasProperty('postBuildExtras')) {
     postBuildExtras()
+}
+
+if (cdvHelpers.getConfigPreference('EnableGoogleServicesPlugin', 'false').toBoolean()) {
+    apply plugin: 'com.google.gms.google-services'
 }


bin/templates/project/app/build.gradle
```diff
@@ -322,3 +323,7 @@ for (def func : cdvPluginPostBuildExtras) {
 if (hasProperty('postBuildExtras')) {
     postBuildExtras()
 }
+
+if (cdvHelpers.getConfigPreference('EnableGoogleServicesPlugin', 'false').toBoolean()) {
+    apply plugin: 'com.google.gms.google-services'
```
macdonst: @chemerisuk I've had to make this line:

```
apply plugin: 'com.google.gms.googleservices.GoogleServicesPlugin'
```

because the class must be used instead of id(string) to be able to apply plugin from non-root gradle file. I haven't tested the PR yet but I'm assuming the reason you are using the ID is because it is a root gradle file?

chemerisuk: Yes, because this is the root Gradle file. Worked fine on my side.

For testing I commented changes from cordova-support-google-services and inserted PR code manually. Then rebuilt the project and tried to start (btw google services v15 trigger error at runtime).

macdonst: 👍

macdonst:
@infil00p wanna take a look at this?

bin/templates/project/app/build.gradle
```diff
@@ -30,6 +30,7 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:3.1.0'
+        classpath 'com.google.gms:google-services:3.2.1'
```
erisu: Would it be possible to wrap this `classpath` in an if condition, like how the apply is done?
It would be nice to minimize the amount of downloading of build related libraries if they are not needed.

```
if (cdvHelpers.getConfigPreference('EnableGoogleServicesPlugin', 'false').toBoolean()) {
  classpath 'com.google.gms:google-services:3.2.1'
}
```

chemerisuk: It looks like the object `cdvHelpers` is not available at the dependency scope. Project build failed with the error:
```
Could not get unknown property 'cdvHelpers' for object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

erisu: @chemerisuk Yes, it seems `cordova.gradle` would need to be applied within the scope so it becomes available.

```
    dependencies {
        apply from: '../CordovaLib/cordova.gradle'

        classpath 'com.android.tools.build:gradle:3.1.0'

        if(cdvHelpers.getConfigPreference('EnableGoogleServicesPlugin', 'false').toBoolean()) {
            println 'Adding classpath "com.google.gms:google-services:3.2.1"'

            classpath 'com.google.gms:google-services:3.2.1'
        }
    }
```

@dpogue Would you know off-hand if something like this would be problematic? The `cordova.gradle` content seems to be only helper methods. Adding it seems to solve the desire to make the classpaths conditional.

dpogue: I don't know much about Gradle, and my experience consists mostly of swearing at it.

I believe you're correct that `cordova.gradle` is helper methods, but I don't know anything about scoping or when it's safe to import (apply?) stuff.

erisu:
@chemerisuk I would like to try and get this PR updated and merged before the upcoming major release.

If possible, could you rebase the PR with master and fix the conflicting files?

Also, as I proposed in a previous comment, could you wrap the `classpath 'com.google.gms:google-services:3.2.1'` with the `EnableGoogleServicesPlugin`?

The code snippet that I provided in a previous comment will work as intended. 

You will need to add the `apply from: '../CordovaLib/cordova.gradle'` to the `dependencies` scope to have access to the helper methods. I had tested this locally and it should not cause any problems.

Lastly, since this PR was created awhile back, I had some other nice to have that could be applied with this PR, if you would like to do this, I can do this in a second PR if needed.

1. Bump the Google Services version to a newer version. 

I believe the version must still be below `4.2.0` as the `phonegap-plugin-push` had build issues with `4.2.0`.

Actually with this PR and `phonegap-plugin-push` has a build issue and I am suspecting it might be related that the `phonegap-plugin-push` depends on the `cordova-support-google-services`

2. Could we introduce a second preference config options: `GoogleServicesPluginVersion` which allows users to change the version of the plugin that they want to use... Have it default to the version below `4.2.0`.

Please let me know if you able to do this or if I should create a new PR based from your changes and apply the additional changes.

erisu:
@chemerisuk I went ahead and created a second PR to be able to quickly fix the merge conflicts as well as apply the additional change request that I posted.

This PR will close as soon as my PR is merged. Since I cherry-picked your commit in, it will still contain your contribution.

Thank you for your contribution and sorry that it took this long to merge in your commit.

chemerisuk:
@erisu thanks for the update. Unfortunately my current environment is a bit different now, so it will be hard to test those changes properly. So please go ahead with the PR, it's a long waited one.

If you have time could you post a link to a compilation problem at `phonegap-plugin-push` you've mentioned.

erisu:
@chemerisuk I do not have an issue ticket on the `phonegap-plugin-push` issues I mentioned. But I pushed a PR that briefly explains some changes to the plugin that will use these new preference option over the previously added dependencies. The issue was around added the enable google services plugin as a dependency with these new changes. It created conflicts.

Here is the PR: https://github.com/phonegap/phonegap-plugin-push/pull/2873

It is currently labeled as `DRAFT` because we will need to first release the major.

PR #440, opened 2018-05-12
Basic support for annotation CordovaMethod

chemerisuk:
### Platforms affected
Android

### What does this PR do?
The PR provides basic support for new annotation `@CordovaMethod` that will reduce boilerplate code for plugin authors.

At present typical Cordova plugin looks like:
```java
public class MyPlugin extends CordovaPlugin {
    private static final String METHOD_1 = "method1";
    private static final String METHOD_2 = "method2";
 
    @Override
    public boolean execute(String action, CordovaArgs args, CallbackContext callbackContext) throws JSONException {
        if (METHOD_1.equals(action)) {
            method1(args, callbackContext);
        } else if (METHOD_2.equals(action)) {
            method2(args, callbackContext);
        /// more methods...
        } else {
            return false;
        }
        return true;
    }

    protected void method1(CordovaArgs args, CallbackContext callbackContext) {
        // method1 implementation
    }

    protected void method2(CordovaArgs args, CallbackContext callbackContext) {
        // method2 implementation
    }
    ...
}
```
The proposal is to have update default implementation of method `CordovaPlugin#execute` and to mark action methods with a new annotation `@CordovaMethod`:
```java
public class MyPlugin extends CordovaPlugin {
    @CordovaMethod
    protected void method1(CordovaArgs args, CallbackContext callbackContext) {
        // method1 implementation
    }
    
    @CordovaMethod
    protected void method2(CordovaArgs args, CallbackContext callbackContext) {
        // method2 implementation
    }
    ...
}
```

This is very similar to how `CDVPlugin` from `cordova-ios` works.

I've previously created a [thread](https://lists.apache.org/thread.html/b6746c7dcd12d5685bcd7bea813651dcc349f3be31f864dd4a27080d@%3Cdev.cordova.apache.org%3E) in the mail list. There were different opinions.

The most popular concern was related to **performance**. So I created [gist](https://gist.github.com/chemerisuk/287e6cd6e1aa61a5680027967008b702) that measures performance impact of different implementations. On my machine an extra time for the `MethodCordovaPlugin` strategy was an approx 0.03ms on the first call and 0.000051ms later. It's on a desktop machine and proper testing should be done still on mobile devices, but wait, cordova plugins should not be used for performance critical apps.

A note about **backward compatibility**. Changes in this PR done in `CordovaPlugin#execute` that currently just `return false`. Plugin authors must override it in order to invoke any logic. Therefore they won't notice any change. Unlike them new plugins with target future version of `cordova-android` might use annotation `@CordovaMethod` and reduce boilerplate code.

### Checklist

FYI @wtrocki

### DIFF
--- a/framework/src/org/apache/cordova/CordovaPlugin.java
+++ b/framework/src/org/apache/cordova/CordovaPlugin.java
@@ -19,6 +19,7 @@
 package org.apache.cordova;
 
 import org.apache.cordova.CordovaArgs;
+import org.apache.cordova.CordovaMethod;
 import org.apache.cordova.CordovaWebView;
 import org.apache.cordova.CordovaInterface;
 import org.apache.cordova.CallbackContext;
@@ -34,6 +35,9 @@ import android.os.Bundle;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.lang.reflect.Method;
+import java.util.Map;
+import java.util.HashMap;
 
 /**
  * Plugins must extend this class and override one of the execute methods.
@@ -43,6 +47,7 @@ public class CordovaPlugin {
     public CordovaInterface cordova;
     protected CordovaPreferences preferences;
     private String serviceName;
+    private Map<String, Method> methodsMap;
 
     /**
      * Call this after constructing to initialize the plugin.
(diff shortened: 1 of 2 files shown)


framework/src/org/apache/cordova/CordovaPlugin.java
```diff
@@ -132,6 +137,35 @@ public boolean execute(String action, JSONArray args, CallbackContext callbackCo
...
+                if (cordovaMethod != null) {
+                    try {
+                        method.setAccessible(true);
+                        if (cordovaMethod.action().isEmpty()) {
+                            methodsMap.put(method.getName(), method);
+                        } else {
+                            methodsMap.put(cordovaMethod.action(), method);
+                        }
+                    } catch (SecurityException e) {
+                        e.printStackTrace();
```
wtrocki: Worth to add `LOG.d(TAG, "Problem when building method map");` etc.

framework/src/org/apache/cordova/CordovaPlugin.java
```diff
@@ -132,6 +137,35 @@ public boolean execute(String action, JSONArray args, CallbackContext callbackCo
      * @return                Whether the action was valid.
      */
     public boolean execute(String action, CordovaArgs args, CallbackContext callbackContext) throws JSONException {
+        if (methodsMap == null) {
+            methodsMap = new HashMap<String, Method>();
+            for (Method method : this.getClass().getDeclaredMethods()) {
+                CordovaMethod cordovaMethod = method.getAnnotation(CordovaMethod.class);
+                if (cordovaMethod != null) {
+                    try {
```
wtrocki: IMHO `try should be outside for for 2 reasons:
- if one the method fails then rest of them will as well.

wtrocki: IMHO `try` should be outside `for` loop for couple reasons:
- if one the method fails then rest of them will as well. This will cause lots of noise in logs
- if one of the methods fails then entire plugin will be in inconsistent state. Proper handling will be to just stop processing if any error will happen (probably on single one).
- opening many try/catch blocks in loop is less performant than single one.

chemerisuk:
After some thinking I created a separate plugin [cordova-support-android-plugin](https://github.com/chemerisuk/cordova-support-android-plugin) instead of submitting changes to the platform. The important advantage of such approach is that it's possible to use this plugin for existing crodova-android versions without waiting for the next release.

@wtrocki I've addressed your comments in the latest commits to [`master branch`](https://github.com/chemerisuk/cordova-support-android-plugin/commits/master). Please take a look at the source code for now. README will be updated shortly.

PR #441, opened 2018-05-14
CB-14089: (android) Add Kotlin support

joshchandler:
### Platforms affected
Android

### What does this PR do?
This adds support for the Kotlin programming language.  Plugins can start being written in or converted to the language with this added.

I haven't added any testing yet, but I created an example project by updating the `./bin/create` command to use `Activity.kt` as the main activity instead of `Activity.java`.  Running `./cordova/run` compiles the app and the app runs without error.

### Checklist


bin/templates/project/Activity.kt
```diff
@@ -0,0 +1,38 @@
+/*
```
joshchandler: I'm not entirely sure if this file should make it in and it's currently not hook up to the create method. I had just created it to do some functional testing.  They claim that Kotlin is 100% interchangeable with Java as long as the stdlibs are imported.. Since that's now a thing with this change we could actually use this in favor of `Activity.java`, if you all are open to it.

dpogue:
These questions might be obvious to someone more familiar with gradle, but I have only limited experience with it. Does this cause additional libraries to be added unconditionally to non-Kotlin projects? Is there an increase in output binary size for existing projects?

joshchandler:
@dpogue Yes this adds the Kotlin library to all projects in all instances.  Initially I thought that it'd make sense to add it anyways since it's an official language for Android development..

I'd be open to only adding it if it's explicitly defined in config.xml or something similar to that.

bin/templates/project/app/build.gradle
```diff
@@ -30,13 +32,14 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:3.1.0'
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
```
erisu: Can we wrap all of the Kotlin code and make it conditional by a flag? This is regarding to the same concerns that @dpogue mentioned.

It might be possible that not everyone would use Kotlin.

For example:
```
if (cdvHelpers.getConfigPreference('EnableKotlin', 'false').toBoolean()) {
  classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
}
```

erisu: And this conditional statement would be applied everywhere else relating to Kotlin.

bin/templates/project/Activity.kt
erisu: If we make Kotlin an optional feature, this file should not be used and defaults to `Activity.java`. Only use this file if the EnableKotlin flag is set to `true`.

test/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Wed Oct 25 11:17:25 PDT 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.4-bin.zip
```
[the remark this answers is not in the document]
dpogue: Yeah... I think this might already have been updated on master (and maybe the source of a conflict now)

bin/templates/project/Activity.kt
joshchandler: This file is actually not unnecessary.. I'm taking it out.

bin/templates/project/app/build.gradle
joshchandler: Awesome, did not know this was available.  Thanks

erisu: @joshchandler Sorry, the `cdvHelpers` is currently not available in the `dependencies` scope, as I mentioned in this example. 

Wrapping the apply plugin should be OK. Example:
```
if (cdvHelpers.getConfigPreference('EnableKotlin', 'false').toBoolean()) {
    apply plugin: 'kotlin-android'
}
```

There is a discussion on a PR that also modifies the the same `build.gradle` file for other features. It also had the same issue where `cdvHelpers` is not available in the dependency scope. Adding it is as easy as adding the `apply from: '../CordovaLib/cordova.gradle'` in the scope BUT I am waiting for any feedback about it.  
You can see the comment here: https://github.com/apache/cordova-android/pull/438#discussion_r216195552

joshchandler: Oh I see.. Ok well then I'll wait until feedback is received on that google services PR.  I'm not familiar enough with gradle to confidently say that importing into the dependencies scope wouldn't cause any problems.  `cordova.gradle` only consists of helper methods, so I don't see any reason why it would be an issue though.

If it doesn't cause any issues then this would just be another reason to pull it in here and the other dependencies scope further below this file.

bin/templates/project/app/build.gradle
```diff
@@ -30,13 +34,16 @@ buildscript {
...
+            classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
+        }
     }
 }
 
 // Allow plugins to declare Maven dependencies via build-extras.gradle.
 allprojects {
     repositories {
-        mavenCentral();
+        mavenCentral()
```
videmort: I think that shoud be google() instead mavenCentral()

brodybits: I think this should be proposed in a separate PR (this may be done in PR #507).

Also I think removing semicolons should be proposed in a separate PR.

brodybits:
What is the status of this proposal?

If we merge PR #507 (Gradle updates) I think this one should be rebased.

dpogue:
The status is that we're trying to find an option that doesn't unconditionally pull kotlin support libraries into all Cordova projects, although it's worth more investigation there to see whether that actually impacts the resulting APK filesize.

joshchandler:
I had some time to come back to this..  

This now successfully ensures that the Kotlin binaries are not added if the config.xml preference `EnableKotlin` is false or does not exist.  Building the example app in the project, without this preference, the compiled apk `app-debug.apk` is 1.8MB.  When set to true, we are adding the binaries for Kotlin and `app-debug.apk` grows to a size of 2.6MB.  So, this change will not affect android projects unless that preference is explicitly set to true.

Given that it adds 0.8MB to the apk, is that large enough to keep this preference set to false by default?

test/build.gradle
```diff
@@ -19,6 +19,7 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext.kotlin_version = '1.3.31'
```
timbru31: Given the PR is already one year old, this should be updated (https://mvnrepository.com/artifact/org.jetbrains.kotlin/kotlin-stdlib)

bin/templates/project/app/build.gradle
```diff
@@ -291,6 +300,11 @@ android {
 
 dependencies {
     implementation fileTree(dir: 'libs', include: '*.jar')
+
+    if (cdvHelpers.getConfigPreference('EnableKotlin', 'false').toBoolean()) {
+        compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
```
timbru31: Why do we require the `jdk7` stdlib here? This should be changed to `kotlin-stdlib` instead.

```suggestion
        compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```

bin/templates/project/app/build.gradle
```diff
@@ -18,16 +18,25 @@
 */
 
 apply plugin: 'com.android.application'
+if (cdvHelpers.getConfigPreference('EnableKotlin', 'false').toBoolean()) {
+    apply plugin: 'kotlin-android'
+}
 
 buildscript {
+    ext.kotlin_version = '1.3.31'
```
timbru31:
```suggestion
    ext.kotlin_version = '1.3.61'
```

test/app/build.gradle
```diff
@@ -51,4 +52,8 @@ dependencies {
     implementation 'com.android.support:appcompat-v7:26.1.0'
     testImplementation 'junit:junit:4.12'
     testImplementation 'org.json:json:20140107'
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
```
timbru31: see above

```suggestion
    compile "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
```

test/app/build.gradle
```diff
@@ -51,4 +52,8 @@ dependencies {
...
+}
+repositories {
+    mavenCentral()
```
timbru31: See https://blog.bintray.com/2015/02/09/android-studio-migration-from-maven-central-to-jcenter/ why jcenter should be used

```suggestion
    jcenter()
```

test/build.gradle
timbru31:
```suggestion
    ext.kotlin_version = '1.3.61'
```

erisu:
Hi @joshchandler, I noticed that there was no activity after the "request change" that was posted a week ago. I went ahead and created a second PR to quickly prepare and get ready to release Kotlin support.

The new PR uses this PR as its base. (New PR: https://github.com/apache/cordova-android/pull/896)

The PR changes consist of:
* Rebased with the latest master branch
* Fixes the merge conflicts
* Applies suggested changes
* Additionally improves

I expect to merge my PR with in the next day or two which will automatically close this PR.

Since your commits were cherry-picked into my new PR, it will still retain your contributions.

Thank you for your contribution and sorry that it took this long to merge in your commit.

joshchandler:
@erisu No problem! Yea, I had seen the feedback and was planning on addressing them soon.  Thanks for picking this up!

PR #446, opened 2018-06-03
CB-14101 Fix Java version check for Java >= 9

raphinesse:
### Platforms affected
Android

### What does this PR do?
- Fix Java version check for Java >= 9
- Checks that we have exactly Java 1.8 since nothing else works with the Android SDK
- Tell the user that we need exactly Java 1.8 if we can't find it

The relevant function was not covered by unit tests to begin with, so I only tested this manually with the following code:
```js
Promise.resolve(
  require('./bin/templates/cordova/lib/check_reqs').check_java()
).then(console.log)
```
Output for Java 8 is `1.8.0`, for Java 10 is `10.0.1`.

### Checklist

~~Added automated test coverage as appropriate for this change~~

### DIFF
--- a/bin/templates/cordova/lib/check_reqs.js
+++ b/bin/templates/cordova/lib/check_reqs.js
@@ -39,17 +39,6 @@ function forgivingWhichSync (cmd) {
     }
 }
 
-function tryCommand (cmd, errMsg, catchStderr) {
-    var d = Q.defer();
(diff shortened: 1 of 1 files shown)


raphinesse:
Added node 4 compatibility so we can release this in a patch release

raphinesse:
Now it should be fine

PR #449, opened 2018-06-10
CB-14125:(android) Increase old plugin compatibility

jcesarmobile:
### Platforms affected
Android

### What does this PR do?
Increases old plugin compatibility by adding 'app/src/main' to all plugin source-file tags that didn't include it, so the files are copied to the new Android Studio structure

Tested with `cordova plugin add https://github.com/OutSystems/cordova-plugin-audiorecorder.git` as it uses source-file for images and xml

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/pluginHandlers.js
+++ b/bin/templates/cordova/lib/pluginHandlers.js
@@ -320,8 +320,8 @@ function studioPathRemap (obj) {
     if (!obj.targetDir.includes('app/src/main')) {
         if (obj.src.endsWith('.java')) {
             return path.join('app/src/main/java', obj.targetDir.substring(4), path.basename(obj.src));
-        } else if (obj.src.endsWith('.xml')) {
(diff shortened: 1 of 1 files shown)


brodybits:
Really nice!

Wondering if this should be considered a breaking change or not?

purplecabbage:
LGTM!  
I think this is just a minor.bump as we are not changing apis, just how some missing input is treated.

jcesarmobile:
Yeah, I don't think it's a breaking change.
If plugins were updated to use the new path, then it won't affect them
If plugins were updated to use resource-file tag instead of source-file tag, then it won't affect them
Plugins with old paths and using source-file tag instead of resource-file tag for resources are currently broken on cordova-android 7.1.0 and this change will make them work as they used to, so shouldn't be breaking change.

jcesarmobile:
Agree, but we should start by updating core plugins and docs

PR #450, opened 2018-06-11
CB-14127: (android) Move google maven repo ahead of jcenter

jcesarmobile:
### Platforms affected
Android

### What does this PR do?
Move google maven repo ahead of jcenter because of some problem to get Google libs from jcenter

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/plugin-build.gradle
+++ b/bin/templates/cordova/lib/plugin-build.gradle
@@ -20,10 +20,10 @@
 
 buildscript {
     repositories {
-        jcenter()
         maven {
             url "https://maven.google.com"
         }
+        jcenter()
     }
 
     // Switch the Android Gradle plugin version requirement depending on the
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -22,10 +22,10 @@ apply plugin: 'com.android.application'
 buildscript {
     repositories {
         mavenCentral()
-        jcenter()
         maven {
             url "https://maven.google.com"
         }
+        jcenter()
     }
 
     dependencies {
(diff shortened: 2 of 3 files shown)


josenobile:
When this commit already merged in the master branch can be merged into a release? 7.1.1 or 7.2.0 ?

jcesarmobile:
Current master is 7.2.0-dev, so this should be on next 7.2.0 release

garyjohnson:
Seeing this issue in 6.4.0 -- are all versions of cordova-android broken due to this? Will they be patched? Is there a version I can upgrade to now without installing the plugin from git?

jcesarmobile:
Only security fixes are back ported to previous versions, and only for 6 months since latest major version bump, so 6.4.0 won't get this because it's not security related.

This is not released yet, so you have to install from git, or probably using @nightly as version (didn't try, might also need nightly CLI).

You can also manually move it up in your project as workaround

garyjohnson:
Understandable, but aren't all released versions of cordova-android currently broken because of this? Or does this only affect a subset of projects somehow?

jcesarmobile:
It doesn't directly affect cordova-android, it affects some plugins that can't fetch the dependencies because Google didn't publish their libraries correctly in jcenter. This fix is more a workaround than a real fix, the real fix is that Google republish the libraries again and do it right.

And even if it affected all cordova-android versions and all Cordova apps, we only backport security fixes.

smart-union:
Makes sense, thanks

davidsidlinger:
After this change, we're still getting the "old" repository order in `platforms/android/CordovaLib/build.gradle`, though the `build.gradle` in `platforms/android/` is 👌.

dpogue:
@davidsidlinger That was just fixed in https://github.com/apache/cordova-android/pull/523

PR #451, opened 2018-06-13
CB-14145: cordova-common update to resolve npm audit & other updates in patch release

brodybits:
### Platforms affected

Android

### What does this PR do?

- cordova-common _2.2.5_ update _(pinned)_ to resolve ugly `npm audit` _and engine warning messages_
- _pin other dependencies in this patch fix_
- _completely reinstalled dependencies in `node_modules` using npm@6.1.0_
- _update `bundledDependencies`, needed to support deprecated Node.js 4 in this patch fix_
- _update cordova.js from cordova-js@4.2.4 with the following change in this patch fix  (using local coho with apache/cordova-coho#176 for patch release support):_
  - _CB-9366 log error.stack_
- _other changes requested by @raphinesse:_
  - _CB-13923 (android) fix -1 length for compressed files_
  - _CB-14127: (android) Move google maven repo ahead of jcenter_
  - _CB-14101 Fix Java version check for Java >= 9 (#446)_
  - _Emit log event instead of logging directly (#452)_
  - _Fix unsafe property access in run.js (#445)_
- ~~enable node 4 (again), needed to support minor release~~ _(no need: deprecated Node.js 4 is still allowed by engines rule in package.json; `npm install` & `npm test` are still covered by AppVeyor CI on deprecated Node.js 4)_

_resolves `npm audit` warnings & other issues in patch release as needed asap (before the next major release is ready)_

- _`npm audit` with npm@6.1.0 (latest version) shows 0 vulnerabilities_
- _using `cordova platform add brodybits/cordova-android#cordova-common-2-update` to add platform to new Cordova project and test on Android device using `cordova run android` succeeds on the following Node.js versions:_
  - _deprecated Node.js 4 (npm 2.15.11)_
  - _Node.js 6 (npm 3.10.10)_
  - _Node.js 8 (npm 5.6.0)_
  - _Node.js 10 (npm 6.1.0)_
- _`npm test` succeeds (along with some other tasks) on AppVeyor CI & Travis CI_

### Checklist

- [x] _[Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database_
- [x] _Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected (with some exceptions)._


raphinesse:
Why do we need the dependency update before unbundling the dependencies (and thus the next major release)?

brodybits:
Now tested as follows:
- passes `npm test` on Travis CI
- Using `cordova platform add https://github.com/brodybits/cordova-android#cordova-common-2-update` to add cordova-android with the proposed changes to new Cordova app results in an app that starts properly on Android (device)
- Using `cordova platform add https://github.com/brodybits/cordova-android#cordova-common-2-update` to add cordova-android with proposed changes to `cordova-sqlite-storage` test suite results in app that passes `cordova-sqlite-storage` test suite on Android (device)

The proposed changes resolves the `npm audit` warning messages on a minor release.

Menardi:
@brodybits Why does this PR require re-enabling support for Node 4? This would break a number of waiting PRs which depend on Node 6+.

dpogue:
We can't drop support for node 4 in a minor version update, but I agree that we should keep master moving forward to the next major version.

@brodybits Can we do these updates for the next minor on top of the 7.1.x branch? That way we can simply delete node_modules entirely on master and not worry about introducing more conflicts into the pending PRs

brodybits:
> @brodybits Can we do these updates for the next minor on top of the 7.1.x branch?

Will do.

raphinesse:
I've cherry-picked all code fix commits from master that I could identify.

brodybits:
> I've cherry-picked all code fix commits from master that I could identify.

Thanks @raphinesse, updated title yet again to reflect what we actually want to do in the patch release.

TBH I have some mixed feelings, though not major. In general I would rather avoid including other fixes when making a security related patch. For a security patch we want the least risk possible that something goes wrong and the "end" user decides to roll back.

I think the actual security risk is very low. In general I would rather keep it this way.

Another really strange thing is that f05e61db07ff7b7acbcadaa6427b7b12eaac8950 seems to have a MacBook-Pro.local address, not linked to any user on GitHub.

What do you think, any comments?

brodybits:
FYI the delay is because it is taking me much longer than I expected to resolve the npm audit issues on cordova-js which I would consider to be an upstream dependency. My apologies for the confusion.

/cc @jcesarmobile

raphinesse:
As a user, I'd be annoyed by a patch release not including fixes to my problems even though they are ready to ship.

I see your point however. Since you do the release, decide at your own discretion.

brodybits:
@raphinesse I just updated the last 2 commits to reference the PRs they originated from and show what kind of scripts are affected.

(GitHub links to the PRs from the commits in master, would be nice to have the same kind of thing in the 7.1.x branch. To show what kind of scripts are affected would also be good for the patch release notes.)

I may just include all the fixes in the same patch release to streamline things. I think the npm audit issues will not lead to any real vulnerabilities, at least not yet:)

raphinesse:
I'd be happy to see these fixes in the upcoming release. I don't see why these changes should break anything. If things break, we'll fix them. I think you need to worry a bit less :blush:

brodybits:
I just pushed a complete rebase, with the following updates:
- Update cordova.js from cordova-js@4.2.4 with the following:
  - CB-9366 log error.stack
- update to cordova-common@2.2.5
- pin other dependencies in package.json
- completely reinstalled dependencies in `node_modules` using npm@6.1.0
- update bundledDependencies in package.json, needed by the deprecated node 4 version
- include changes that were added by @raphinesse (with PR number added to the commit messages in the last 3 changes)

tested in Cordova project using node 4 (deprecated) and node 8.

brodybits:
> I don't see why we need separate commits for removing an adding node_modules contents.

As I explained in <https://github.com/apache/cordova-windows/pull/281#issuecomment-403521072>: a combination of updated dependencies and npm from non-deprecated version of Node.js results in such a massive change to node_modules that it seems cleanest to remove old node_modules before making the update.

> It seems you listed each transitive dependency in bundledDependencies. According to this npm test we should not need to do that. Was there something that made this necessary?

Yes, also explained in <https://github.com/apache/cordova-windows/pull/281#issuecomment-403521072>: With node_modules installed by newer version of npm (comes with non-deprecated version of Node.js), additional libraries need to be listed to work on Node.js 4. We know that Node.js 4 is deprecated but should not be dropped in a patch release:-(

Thanks for checking, will probably merge this really soon.

PR #455, opened 2018-06-20
Start 8.0.0-dev in master - WIP

brodybits:
### Platforms affected

Android

### What does this PR do?

Set version to 8.0.0-dev via coho

This is definitely needed since node 4 support was dropped from CI and we would like to avoid double revert that would be needed to continue with a minor release in master.

I would like to go ahead with this asap, will wait for an approval for up to 12 hours. I would prefer to merge this myself, no squash merge or extra merge commit please.

I would be happy to start 7.2.x branch if wanted to ship the existing changes in a minor release, do not personally see the value of this. We could always ship a patch release with cherry-pick of existing bug fixes if really needed for any reason.

Just like #454 I had to use a custom version of cordova-coho with a local patches to get it to work in my workarea, hoping to fix cordova-coho someday in the future.

-

### Checklist

- ~~[Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database~~
- ~~Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.~~


jcesarmobile:
I don't understand it yet.

You said you wanted to do a patch release, that it has not started yet. 
Here you talk about creating a branch for a minor release in case we want to do it with current changes.
But this PR is about a major future release.

I think we should do the patch or minor release now, and then bump the major version after we are done with the release.

brodybits:
> You said you wanted to do a patch release, that it has not started yet.

I just addressed that point on the email thread, would like to keep patch release discussion there if possible.

> Here you talk about creating a branch for a minor release in case we want to do it with current changes.

I would actually prefer **not** to create a branch for a minor release unless absolutely necessary. My apologies for the confusion.

> and then bump the major version after we are done with the release.

my response coming later

P.S. Travis CI went red since cordova-coho stamps VERSION with double-quotes in `bin/templates/cordova/version`. Marking this PR as WIP for now, will try to fix in cordova-coho.

dpogue:
> I think we should do the patch or minor release now, and then bump the major version after we are done with the release.

There are already breaking changes on master, so any patch release needs to be based off the previous release branch.

Honestly, I'd kinda prefer we all focus on the major update and not leave PRs stagnating while we debate patch versions. There's a lot of uncertainty right now about whether it's safe to merge stuff or whether that will break patch releases, there's a bunch of conflicting work going into updating committed dependencies for patch releases when we're planning to drop those committed dependencies in the major, etc.

brodybits:
> There are already breaking changes on master, so any patch release needs to be based off the previous release branch.

Agreed

"previous release branch" is `7.1.x` to be absolutely clear

> Honestly, I'd kinda prefer we all focus on the major update and not leave PRs stagnating while we debate patch versions.

Agreed

It is because of these points that I wanted to update the current working version to `8.0.0-dev`.

raphinesse:
Sorry, I can't review this. I've got no clue about the release cycle or when versions are bumped to what *-dev marker. I have to say that the ongoing preparation of the coming patch releases seemed a bit chaotic to me.

brodybits:
Closing for now

janpio:
(Forgot actually closing? Did it for you @brodybits)

PR #456, opened 2018-06-21
More android tests

Menardi:
### Platforms affected
Android

### What does this PR do?
Adds more unit tests for Android.

Since my last PR adding tests went well, I decided to add some more. I have basically gone alphabetically and added tests to give full coverage for `Adb.js` and `AndroidManifest.js`. I may have time later in the next few days to add more tests, but I didn't want to make a huge PR which would take a long time to review.

I also made a small change to the tests for `Api.js`. These tests previously logged `Android Studio project detected`, polluting the test reporter's output. They now no longer do this.

I know the contribution guidelines say to squash PRs into one commit, but in this case, I believe the separate commits make sense. I suppose this could be two (or three) separate PRs. If I should do that next time, please let me know.

Run `npm run unit-tests` locally

### Checklist

### DIFF
--- a/spec/unit/Adb.spec.js
+++ b/spec/unit/Adb.spec.js
@@ -1,0 +1,229 @@
+/**
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+const CordovaError = require('cordova-common').CordovaError;
+const rewire = require('rewire');
+
+describe('Adb', () => {
+    const adbOutput = `List of devices attached
+emulator-5554\tdevice
+123a76565509e124\tdevice`;
+    const [, emulatorLine, deviceLine] = adbOutput.split('\n');
+    const emulatorId = emulatorLine.split('\t')[0];
+    const deviceId = deviceLine.split('\t')[0];
+
+    const alreadyExistsError = 'adb: failed to install app.apk: Failure[INSTALL_FAILED_ALREADY_EXISTS]';
+    const certificateError = 'adb: failed to install app.apk: Failure[INSTALL_PARSE_FAILED_NO_CERTIFICATES]';
+    const downgradeError = 'adb: failed to install app.apk: Failure[INSTALL_FAILED_VERSION_DOWNGRADE]';
+
+    let Adb;
+    let spawnSpy;
+
+    beforeEach(() => {
+        Adb = rewire('../../bin/templates/cordova/lib/Adb');
+        spawnSpy = jasmine.createSpy('spawn');
+        Adb.__set__('spawn', spawnSpy);
+    });
+
+    describe('isDevice', () => {
+        it('should return true for a real device', () => {
+            const isDevice = Adb.__get__('isDevice');
+
+            expect(isDevice(deviceLine)).toBeTruthy();
+            expect(isDevice(emulatorLine)).toBeFalsy();
+        });
+    });
+
+    describe('isEmulator', () => {
+        it('should return true for an emulator', () => {
+            const isEmulator = Adb.__get__('isEmulator');
+
+            expect(isEmulator(emulatorLine)).toBeTruthy();
+            expect(isEmulator(deviceLine)).toBeFalsy();
+        });
+    });
+
+    describe('devices', () => {
+        beforeEach(() => {
+            spawnSpy.and.returnValue(Promise.resolve(adbOutput));
+        });
+
+        it('should return only devices if no options are specified', () => {
+            return Adb.devices().then(devices => {
+                expect(devices.length).toBe(1);
+                expect(devices[0]).toBe(deviceId);
+            });
+        });
+
+        it('should return only emulators if opts.emulators is true', () => {
+            return Adb.devices({emulators: true}).then(devices => {
+                expect(devices.length).toBe(1);
+                expect(devices[0]).toBe(emulatorId);
+            });
+        });
+    });
+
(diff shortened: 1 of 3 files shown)


spec/unit/Adb.spec.js
```diff
@@ -0,0 +1,222 @@
...
+
+        it('should reject with a CordovaError on a shell error', () => {
+            const errorMessage = 'Test Start error';
+            spyOn(Adb, 'shell').and.returnValue(Promise.reject(errorMessage));
+
+            return Adb.start(deviceId, activityName)
+                .then(() => fail('Unexpectedly resolved'))
+                .catch(err => {
+                    expect(err instanceof Adb.__get__('CordovaError')).toBe(true);
+                    expect(err).toMatch(errorMessage);
```
raphinesse: I usually explicitly check the message property to make sure it's OK. Coercing an error to string can give you all kinds of junk in addition to the message. Apply elsewhere accordingly.

janpio:
Wheee, additional tests are always nice. Great work @Menardi !

spec/unit/AndroidManifest.spec.js
```diff
@@ -0,0 +1,311 @@
...
+        it('should throw an error if not a valid manifest', () => {
+            createTempManifestFile(`<?xml version='1.0' encoding='utf-8'?><notamanifest></notamanifest>`);
+
+            try {
+                manifest = new AndroidManifest(manifestPath);
+                fail('Expected error to be thrown');
+            } catch (e) {
+                // Pass
+            }
+        });
```
raphinesse: Replace with
```js
expect(_ => new AndroidManifest(manifestPath)).toThrowError(CordovaError);
// or if it throws another Error
expect(_ => new AndroidManifest(manifestPath)).toThrowError();
```

spec/unit/AndroidManifest.spec.js
```diff
@@ -0,0 +1,311 @@
...
+
+    describe('versionName', () => {
+        it('should get the version name', () => {
+            expect(manifest.getVersionName()).toBe(VERSION_NAME);
+        });
+
+        it('should set the version name', () => {
+            manifest.setVersionName('6.1.0');
+            expect(manifest.getVersionName()).toBe('6.1.0');
+        });
```
raphinesse: For all the setter tests, it would be nice if it were immediately obvious, that the value you are setting is *not* equal to the original value in the Manifest. Maybe something along the lines of:
```js
const newVersion = VERSION_NAME + '-new';
manifest.setVersionName(newVersion);
expect(manifest.getVersionName()).toBe(newVersion);
```
What do you think?

spec/unit/AndroidManifest.spec.js
```diff
@@ -0,0 +1,311 @@
...
+            expect(manifest.getTargetSdkVersion()).toBe(TARGET_SDK_VERSION);
+        });
+
+        it('should set targetSdkVersion', () => {
+            manifest.setTargetSdkVersion(1);
+            expect(manifest.getTargetSdkVersion()).toBe(1);
+        });
+
+        it('should create the uses-sdk node if it does not exist when setting targetSdkVersion', () => {
+            const NO_SDK_MANIFEST = DEFAULT_MANIFEST.replace(/<uses-sdk.*\/>/, '');
```
raphinesse: `/<uses-sdk.*?\/>/` or `/<uses-sdk[^>]*\/>/` will be safer if someone decides to add something after the last element.

Or you could not use a RegeExp to parse XML (kidding :wink:) and manipulate your doc after parsing:
```js
manifest.doc.getroot().remove(manifest.doc.getroot().find('./uses-sdk'))
```

spec/unit/AndroidManifest.spec.js
```diff
@@ -0,0 +1,311 @@
...
+            manifest.setDebuggable(true);
+            expect(manifest.getDebuggable()).toBe(true);
+        });
+    });
+
+    describe('write', () => {
+        let fsSpy;
+
+        beforeEach(() => {
+            fsSpy = jasmine.createSpyObj(['writeFileSync']);
```
raphinesse: Use `jasmine.createSpyObj('fs', ['writeFileSync'])` or any Jasmine error messages will refer to the spy as `undefined.writeFileSync`

spec/unit/AndroidManifest.spec.js
Menardi: The reason I didn't do that is that it wouldn't technically be a valid value for the manifest. I suppose it doesn't really matter though since we are really only testing text manipulation, not the validity of the manifest. This would make it more readable, so I will change it.

spec/unit/AndroidManifest.spec.js
Menardi: Ah, that's much better! I couldn't quite figure out the XML lib, and the docs didn't describe how to remove a node. I'll do this, thanks!

spec/unit/AndroidManifest.spec.js
Menardi: That is much better. I should have realised Jasmine could cover this case.

spec/unit/Adb.spec.js
```diff
@@ -0,0 +1,222 @@
...
+        // The following two tests are somewhat brittle as they are dependent on the
+        // exact message returned. But it is better to have them tested than not at all.
+        it('should give a more specific error message if there is a certificate failure', () => {
+            spawnSpy.and.returnValue(Promise.resolve(certificateError));
+
+            return Adb.install(deviceId, '')
+                .then(() => fail('Unexpectedly resolved'))
+                .catch(err => {
+                    expect(err).toMatch('Sign the build');
+                });
```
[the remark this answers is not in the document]
raphinesse: I asked because I would assert that err is an error whenever possible. Debugging thrown primitives can be really nasty, due to the lack of a stack trace.

So if we actually get a CordovaError here, I would add
```js
expect(err).toEqual(jasmine.any(CordovaError));
```

spec/unit/Adb.spec.js
```diff
@@ -0,0 +1,222 @@
...
+
+            return Adb.shell(deviceId, shellCommand).then(() => {
+                const args = spawnSpy.calls.argsFor(0);
+                expect(args[0]).toBe('adb');
+
+                const adbArgs = args[1].join(' ');
+                expect(adbArgs).toContain('shell');
+                expect(adbArgs).toMatch(`-s ${deviceId}`);
+                expect(adbArgs).toMatch(new RegExp(`${shellCommand}$`));
+            });
```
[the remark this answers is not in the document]
raphinesse: Yes, I completely agree then. Thanks for the explanation!

raphinesse:
Thanks again for these wonderful tests. :1st_place_medal: 

Thanks to you we have now over 50% coverage! Time to get us some badges :grin:

PR #458, opened 2018-06-27
Unit tests for AndroidProject and Builders

Menardi:
### Platforms affected
Android unit tests

### What does this PR do?
I have written some more unit tests to continue increasing the coverage. This brings coverage to 54.94%. The coverage didn't increase as much as expected, as one of the files I have now written tests for (`GenericBuilder.js`) was not included at all in the previous coverage reports.

Unlike my previous tests, none of the files I have covered this time have reached 100%. This is because they are processing configuration files which I do not have good example of for testing. The main file I need a good example of is `project.properties`, with various entries. If there is one I could use, I am happy to complete the test coverage for these files.

Another thing to note is that `GradleBuilder.js` and `StudioBuilder.js` are very similar files, so the tests are pretty much copy-pasted. `GenericBuilder.js` appears to exist to pull out some functionality shared between the two, but it could definitely contain more. In a future PR, once I have complete test coverage of these files, I would like to break out the common code to make it more maintainable.

Locally run `npm run unit-tests`

### Checklist


raphinesse:
Thanks for the tests! I'll review them as soon as I find time.

Please take a look at #437. These two Builders also bothered me to the point that I wanted to refactor them but thankfully we might be able to get rid of one. Maybe @jcesarmobile can provide you with more details in case you want to tackle this.

raphinesse:
Also note the test failures on windows due to unexpected path separators.

Menardi:
Whoops, those failing Windows tests should be fixed now. I was wondering about the different builders and how they are used. I'll dig deeper and see what I can do.

jcesarmobile:
I'm not really sure what the GradleBuilder is for, it was probably used before we had the StudioBuilder to build Eclipse Gradle projects, as Eclipse projects could be built with Ant or Gradle.

According to a Joe's comment on this PR, it's not used now that we only have Android Studio projects https://github.com/apache/cordova-android/pull/437, so should be safe to delete.

All Android Studio projects are Gradle projects, that's probably why the GradleBuilder and StudioBuilder are similar.

raphinesse:
> All Android Studio projects are Gradle projects

@jcesarmobile Because of this fact I thought that we should name the remaining builder GradleBuilder. Especially if it might also be used to build non-Studio projects. Does that make sense?

spec/unit/AndroidProject.spec.js
```diff
@@ -0,0 +1,134 @@
...
+
+    beforeEach(() => {
+        AndroidProject = rewire('../../bin/templates/cordova/lib/AndroidProject');
+
+        AndroidStudioSpy = jasmine.createSpyObj('AndroidStudio', ['isAndroidStudioProject']);
+        AndroidProject.__set__('AndroidStudio', AndroidStudioSpy);
+    });
+
+    describe('constructor', () => {
+        it('should set the project directory and www folder', () => {
```
raphinesse: This tests only the first part, right?

spec/unit/AndroidProject.spec.js
```diff
@@ -0,0 +1,134 @@
...
+    });
+
+    describe('getProjectFile', () => {
+        it('should create and return a new project if one does not exist', () => {
+            const newProject = AndroidProject.getProjectFile(PROJECT_DIR);
+
+            expect(newProject).toEqual(jasmine.any(AndroidProject));
+        });
+
+        it('should return the existing project', () => {
```
raphinesse: Maybe "should cache created projects"?

spec/unit/AndroidProject.spec.js
```diff
@@ -0,0 +1,134 @@
...
+            });
+        });
+
+        it('should only remove the specified project from the cache', () => {
+            const projectToRemove = 'project2';
+            AndroidProject.purgeCache(projectToRemove);
+
+            const cache = AndroidProject.__get__('projectFileCache');
+            expect(Object.keys(cache).length).toBe(2);
+            expect(cache[projectToRemove]).toBe(undefined);
```
raphinesse: I find `in` to be vastly underused in the JS community :wink:

spec/unit/AndroidProject.spec.js
```diff
@@ -113,15 +113,15 @@ describe('AndroidProject', () => {
 
             androidProject.getPackageName();
 
-            expect(AndroidManifestSpy).toHaveBeenCalledWith(`${PROJECT_DIR}/AndroidManifest.xml`);
+            expect(AndroidManifestSpy).toHaveBeenCalledWith(path.join(PROJECT_DIR, 'AndroidManifest.xml'));
```
raphinesse: That went into the wrong commit

jcesarmobile:
All Cordova projects are Android Studio projects. That just means that you can open them in Android Studio without problems and they will work if you run from there.

We are not going to build projects that are not compatible with Android Studio.

But as Android Studio projects are also Gradle projects, you don't need to open Android Studio to build them.

About the naming, I don't care much if we keep StudioBuilder or we keep the GradleBuilder and move Android Studio stuff there.

janpio:
So one could rename the AndroidBuilder.js to e.g. ProjectBuilder.js, integrate GenericBuilder.js and clean the code to match this?

raphinesse:
@janpio I like ProjectBuilder. Future proof naming for whatever Google throws at us :smile:

spec/unit/AndroidProject.spec.js
Menardi: Ah yes, you're right. It originally tested `project.www` but I ended up breaking those into separate tests and forgot to rename this one.

spec/unit/AndroidProject.spec.js
Menardi: Yeah, it wouldn't even cross my mind to use it!

Menardi:
Thanks for all your feedback. It sounds like all the three builders could be made in to one. At that point do we still need `builder.js`, which is the intermediary that returns the requested builder? I'm not sure if its use is only within `cordova-android`, or if `builders.getBuilder` is a more general Cordova spec.

If it's not needed, then all of these files could be brought into a single file (`ProjectBuilder.js`), which would certainly make it a lot easier to understand.

As for the testing, I mentioned that I would like to fully cover the builders, but I don't have a good `project.properties` file to write the tests around. Do any of you have an example file with many entries, which I could use?

PR #459, opened 2018-06-28
CB-14139 android: Add jvmargs flag for custom values

erisu:
### Platforms affected
Android

### What does this PR do?
Added an additional flag `--jvmargs` to allow users to customize the Gradle's JVM memory heap size settings.

Document updates have been submitted in [cordova-docs PR 839](https://github.com/apache/cordova-docs/pull/839).

JIRA Ticket: [CB-14139](https://issues.apache.org/jira/browse/CB-14139)

- cordova build and run with and without flag settings.

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/build.js
+++ b/bin/templates/cordova/lib/build.js
@@ -53,6 +53,7 @@ function parseOpts (options, resolvedTarget, projectRoot) {
         buildMethod: process.env.ANDROID_BUILD || 'studio',
         prepEnv: options.argv.prepenv,
         arch: resolvedTarget && resolvedTarget.arch,
+        jvmargs: (options.argv.jvmargs || '-Xmx2048m'),
         extraArgs: []
     };
 
--- a/bin/templates/cordova/lib/builders/GradleBuilder.js
+++ b/bin/templates/cordova/lib/builders/GradleBuilder.js
@@ -57,7 +57,7 @@ GradleBuilder.prototype.getArgs = function (cmd, opts) {
     // 10 seconds -> 6 seconds
     args.push('-Dorg.gradle.daemon=true');
     // to allow dex in process
-    args.push('-Dorg.gradle.jvmargs=-Xmx2048m');
+    args.push('-Dorg.gradle.jvmargs=' + opts.jvmargs);
     // allow NDK to be used - required by Gradle 1.5 plugin
     args.push('-Pandroid.useDeprecatedNdk=true');
     args.push.apply(args, opts.extraArgs);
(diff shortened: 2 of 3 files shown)


raphinesse:
Couldn't we instead remove the default jvmargs setting and use one of the various other transparent methods of setting this? As in environment variables (per command) or Gradle properties.

I'm just asking. New features are hard to get rid of, once added.

erisu:
@raphinesse 

I do not mind if the `org.gradle.jvmargs` is defined in the `gradle.properties` file. I also agree that once a new feature is added, it will be hard to remove.

It is referenced in [Gradle's Documentation](https://docs.gradle.org/current/userguide/build_environment.html#sec:configuring_jvm_memory) and [Cordova's Documentation](https://cordova.apache.org/docs/en/latest/guide/platforms/android/index.html#configuring-gradle).

The only **potential issue** is for users that have already created their own project's `gradle.properties` file.

If the `command-line flag` is removed from `GradleBuilder.js` and `StudioBuilder.js`, users with the `gradle.properties` file may need to manually add the property.

Gradle's default to maximum heap size is set to `Xmx1024m` while we had defined ours as `Xmx2048m`.

The `command-line flag` setting was committed for [enabling the dex process for large projects](https://github.com/apache/cordova-android/commit/3a33f4c0a62291ed4faf444cd4525952f90a0b57).

raphinesse:
Thanks for your Analysis.

Looking at the issue this is supposed to fix, I wonder if removing the default memory setting is the right thing here. The way I see it, we would be improving support for 32 bit JVMs while worsening the situation for large projects. That does not sound like a good deal to me.

Is 32bit Java support even relevant? If so, IMHO the best thing would be detecting it and limit the heap size accordingly. However, we probably need the check_reqs cleanup that I put on the roadmap for next major first.

Does that make sense or am I getting something wrong?

janpio:
Huh @erisu? What happened?

erisu:
@janpio, sorry for the missing comment.

The more I think about it, it's probably better to not add an additional flag to set the jvmargs. It is a quicker and simpler solution but as mentioned, it would be harder to remove flag in the future.

Using the gradle.properties would be better in the long run. 

Looking back at the potential issue mentioned above when using the properties file, I think we just need to create a GradlePropertiesConfigParser/Writer. 

If the file doesn't exist, we can either
- Create the file with our defaults
- Continue using the inline command

If the file already exists, parse the file to see if they have the default variables (excluding its value). If the default arguments are missing:
- Edit the file with our defaults
- Continue using the inline command

If they already have the arguments define, then just accept their values.

Also, I feel the amount of 32bit users could probably be very low and this would be an edge case. Since the user had a temporary workaround, we might be able to focus on removing all of the command-line usage in-favor of the gradle.properties instead of just jvmargs.

The user's temporary workaround did mention setting the environment variables but affecting the whole system. Though testing is needed, it might be possible to set the variable during execution which shouldn't affect the whole system. `_JAVA_OPTIONS=-Xmx512m cordova ...`. The only question is if the set variable is still set when gradle runs.

PR #465, opened 2018-07-11
CB-14201: (android) Removes Gradle property in-line command arguments…

erisu:
### Platforms affected
Android

### What does this PR do?
This PR removes the Gradle properties in-line command arguments in favor of setting the gradle.properties file.

This PR will give users more flexibility to override defaults that Cordova defines.

- Locally ran the unit test, eslint, and code coverage.
- Locally ran the Cordova prepare, run and build commands.
- https://travis-ci.org/erisu/cordova-android/builds/402490984
- https://ci.appveyor.com/project/erisu/cordova-android/build/1.0.41

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/builders/ProjectBuilder.js
+++ b/bin/templates/cordova/lib/builders/ProjectBuilder.js
@@ -46,20 +46,15 @@ class ProjectBuilder {
         } else if (cmd === 'debug') {
             cmd = 'cdvBuildDebug';
         }
-        var args = [cmd, '-b', path.join(this.root, 'build.gradle')];
+
+        let args = [cmd, '-b', path.join(this.root, 'build.gradle')];
+
         if (opts.arch) {
             args.push('-PcdvBuildArch=' + opts.arch);
         }
 
-        // 10 seconds -> 6 seconds
-        args.push('-Dorg.gradle.daemon=true');
-        // to allow dex in process
-        args.push('-Dorg.gradle.jvmargs=-Xmx2048m');
-        // allow NDK to be used - required by Gradle 1.5 plugin
-        // args.push('-Pandroid.useDeprecatedNdk=true');
         args.push.apply(args, opts.extraArgs);
-        // Shaves another 100ms, but produces a "try at own risk" warning. Not worth it (yet):
-        // args.push('-Dorg.gradle.parallel=true');
+
         return args;
     }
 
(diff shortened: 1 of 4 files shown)


erisu:
@raphinesse 

I created this PR to replace the previous PR #459.

As discusses in #459, it does feel better to remove the Gradle command line argument usage in-favor of the `gradle.properties` file. This also eliminates the idea or need for creating extra Cordova arguments for setting `jvmargs` or potential other Gradle property overrides.

The user can instead change the values in gradle.properties.

Please let me know what you think of this over the other PR.

Also, this PR is linked to a new ticket, https://issues.apache.org/jira/browse/CB-14201, as the previous ticket only focused on `jvmargs`. This PR will affect all Gradle properties that we current set. (`org.gradle.daemon`, `org.gradle.jvmargs`, `android.useDeprecatedNdk`)

janpio:
Did I get this right: You removed the previous harcoded default configs and created a `gradle.properties` parser that uses the old defaults to create the file if it is missing. Correct?

erisu:
Partial,

Yes, The hardcoded defaults were removed so they can now be overridden by the user from the `gradle.properties`.  

**Additional Note:** Overridden defaults are accepted, not reverted, but the user will see an info message emitted with the Cordova's recommended value.

Yes, the old defaults are used for creating the file. Since the first step for a new project is `cordova platform add`, it will naturally run `prepare` right after. This will create the new `gradle.properties` file with the original defaults and values that were previously hard-coded.

The parser can read other non-default properties but is only used for the original defaults.

As for supplying a lot of other configs, it was already possible by creating this file manually since  Gradle naturally detects and uses the properties file. This PR main goal was removing the three hardcoded command-line arguments so it won't be locked in.

__Here are also some additional use cases:__
**Use Case 1:** For existing projects that are missing the `gradle.properties` file, it will be created during the `prepare` step. Either triggered by `cordova (prepare|run|build)`.

**Use Case 2:** Existing projects with existing  `gradle.properties` file will check and add missing defaults. 

__And lastly, possible drawbacks, but not serious:__
**1.** If the user deletes a default, it will be re-appended. In this case, the user would just need to negate its value or take the default value listed by Gradle's docs if they don't want it or modify it.

**2.** (Possibly a Very Small Edge Case) For existing projects, with no `gradle.properties` file, if only running `cordova compile` without preparing, builds will not run with any defaults. During local testing, with no `gradle.properties` file builds continue to build successfully.

8bhsolutions:
@erisu 
Hi, I am looking to increase the JVM heap size by overwrite the default org.gradle.jvmargs value.

I understand that with this PR, the setting can now be changed in the gradle.properties file created as part of the cordova build process. And if it the defaults are to be changed, the user needs to create their own custom gradle.properties.

It is not clear to me where to put my custom gradle.properties file (in my source directory) so that the build process will pick up the file and overwrite the default file.

Issue #485, opened 2018-08-27
Cordova generates build.gradle with deprecated "compile" dependencies


jedrivisser:
When an android cordova plugin has a `framework` dependency, StudioBuilder.js generates a deprecated `compile` dependency instead of a `implementation` dependency:

https://github.com/apache/cordova-android/blob/7.1.x/bin/templates/cordova/lib/builders/StudioBuilder.js#L201

janpio:
This is probably happen in multiple places.

Are you up to creating a PR for that to fix all of those? We would really appreciate it 💯

jedrivisser:
Will do

jedrivisser:
Changes are done

PR #486, opened 2018-08-27
GH-485: (android) Replace deprecated "compile" with "implementation"

jedrivisser:
### Platforms affected
android

### What does this PR do?
generates gradle dependencies in the `build.gradle` from `<framework>` elements found in the `plugin.xml` as the new `implementation` dependencies instead of deprecated `compile` dependencies

ran `npm run test`

### Checklist
- [✘] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
Created a github Issue: https://github.com/apache/cordova-android/issues/485

- [✓] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.

- [✓ ] Added automated test coverage as appropriate for this change.
None were needed

There are some more references to `compile` dependencies in other files but I left them alone bacause I am not sure what they are used for, see:

https://github.com/apache/cordova-android/blob/7.1.x/bin/templates/cordova/lib/plugin-build.gradle#L41-L43
https://github.com/apache/cordova-android/blob/7.1.x/spec/fixtures/android_studio_project/app/build.gradle#L23-L25

### DIFF
--- a/bin/templates/cordova/lib/builders/ProjectBuilder.js
+++ b/bin/templates/cordova/lib/builders/ProjectBuilder.js
@@ -196,7 +196,7 @@ class ProjectBuilder {
                     throw new CordovaError('Unsupported system library (does not work with gradle): ' + p);
                 }
             }
-            depsList += '    compile "' + mavenRef + '"\n';
+            depsList += '    implementation "' + mavenRef + '"\n';
         });
 
         buildGradle = buildGradle.replace(/(SUB-PROJECT DEPENDENCIES START)[\s\S]*(\/\/ SUB-PROJECT DEPENDENCIES END)/, '$1\n' + depsList + '    $2');


janpio:
> There are some more references to compile dependencies in other files but I left them alone bacause I am not sure what they are used for, see:

Doesn't Android Studio complain now when one opens a project? Does it complain about these two files?

I would say they could be changed as well - but I don't know too much about Android ;)

erisu:
@janpio 
#### Initial Questions
- Do we have a Gradle versions requirement for Cordova projects? 
- How far back do we support? 

#### Overview
The later Android Studio should not complain.

`compile` had been deprecated from Gradle since 3.4, but still exists for backwards compatibility.

`implementation` and `api` was introduced to replace `compile`. The key differences is that `implementation` dependencies are identified as internal while `api` are external. 

```
implementation configuration should be used to declare dependencies which are internal to the component.
```
```
Dependencies appearing in the api configurations will be transitively exposed to consumers of the library, and as such will appear on the compile classpath of consumers. 

Dependencies found in the implementation configuration will, on the other hand, not be exposed to consumers, and therefore not leak into the consumers' compile classpath. 
```

https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_configurations_graph

#### Final Question
- Should we suppose the ability where the user can identify the framework library as `implementation` or `api`?

In most cases, `implementation` should be enough.

With this in mind, is there any concern for merging this PR or should we improve to support identify the type.

janpio:
(I have no idea @erisu - best create an issue with label `discussion` here for your first two questions and get some eyes on it [e.g. by linking to it in Slack or dev mailing list])

jedrivisser:
@erisu @janpio half the `compile` dependencies were already changes to `implementation` in this commit:
https://github.com/apache/cordova-android/commit/9fdb12671519e97e78c36f4d68025246334f7c04
and
https://github.com/apache/cordova-android/commit/18d68845220d0c29122432f4670a3484e47be7b7
otherwise I would have had the same questions

the current output is a mixed one now, for example:

image

janpio:
Oh, we should then probably change the target to `master` here. Whoever creates a patch release for 7.1.x (if at all) should cherry pick up this change over from there.

brodybits:
I would vote -1 on making this kind of change to `7.1.x` branch. Better use a "deprecated" thing than risk breaking something in a patch release, IMHO.

I would favor fixing this kind of an issue on the `master` branch then cherry-picking into a new minor release (7.2) if absolutely needed.

erisu:
I would also agree that cherry picking is the typical and preferred way, but the files being modifying does not exist in master anymore. Also, the file that would need to be modified in master, to be equivalent to this change, is new and does not exist in 7.1.x.

In this case, I don't really think cherry picking will work... Correct me if I am wrong...
@janpio @brodybits

erisu:
Sure, the file and line number in master:

https://github.com/apache/cordova-android/blob/23b24491c3e2ad5721a2813ec08f80985eca5e64/bin/templates/cordova/lib/builders/ProjectBuilder.js#L196

jedrivisser:
So... I assume I need to create a new pull request? because I cannot change the `from` location of the pull request. Do I just close/reject this one?

Changes are available here:
https://github.com/jedrivisser/cordova-android/commit/f4478aeffd269f72be7399aa6ac93a562909d524

jedrivisser:
ah, nevermind, I found a way, should be ready now

Issue #488, opened 2018-09-01
Run command fails if build is skipped


Menardi:
cordova-android version: v8.0.0-dev

As part of the builders cleanup PR (#461), I removed all the builders except for the StudioBuilder. It turns out that the `none` option is still invoked when the user uses the `cordova run` command in a way that doesn't trigger a build (like if a build has been done just before it). This causes the run command to fail.

cc @erisu

erisu:
To expand on the issue, it appears that `cordova run` actually triggers two builds. The first build uses `studio` and builds successfully. The second build is triggered after the message `deploying to emulator` and tries to use `none` which throws an exception.

[Build Log File](https://github.com/apache/cordova-android/files/2342207/build.log)

Additionally, when using `cordova build android` and installing the app with `adb install` the application closes immediately when trying to launch the app. It might be related to pathing. Using `adb logcat` you will see this error also.

[ADB Log File](https://github.com/apache/cordova-android/files/2342227/adb.log)

PR #490, opened 2018-09-04
Fixes build & run related bugs from builder refactor

erisu:
### Platforms affected
Android

### What does this PR do?
Fixes, refactors, and improves the build and run process that came from the builder refactor.

- General Code Refactor
- Removed builder type argument from getBuilder API
- Removed any reference of conditional statements around builder type
- Remove plugin handler install and uninstall option flag android_studio
- Remove --gradle flag references
- Fixed plugin handler install and uninstall pathing issues
- Use the nobuild flag option to control the run build.
- Updated test spec to reflect the changes.

#### General Tests
- [x] `npm test`
- [x] `npm run eslint`

#### Default Template
- [x] `cordova prepare android`
- [x] `cordova compile android`

#### Builds Tested with Device
- [x] `cordova build android`
- [x] `cordova build android --release --device`
- [x] `cordova build android --debug --device`
- [x] `cordova run android --device`

#### Builds Tested on Emulator
- [x] `cordova build android`
- [x] `cordova build android --release --emulator`
- [x] `cordova build android --debug --emulator`
- [x] `cordova run android --emulator`

### Checklist
- [x] Created GitHub Issue #488
- [x] Updated automated test coverage as appropriate for this change.

### DIFF
--- a/bin/lib/create.js
+++ b/bin/lib/create.js
@@ -141,9 +141,9 @@ function writeProjectProperties (projectPath, target_api) {
 }
 
 // This makes no sense, what if you're building with a different build system?
-function prepBuildFiles (projectPath, builder) {
+function prepBuildFiles (projectPath) {
     var buildModule = require(path.resolve(projectPath, 'cordova/lib/builders/builders'));
-    buildModule.getBuilder(builder).prepBuildFiles();
+    buildModule.getBuilder().prepBuildFiles();
 }
 
 function copyBuildRules (projectPath, isLegacy) {
@@ -329,7 +329,7 @@ exports.create = function (project_path, config, options, events) {
             });
             // Link it to local android install.
             exports.writeProjectProperties(project_path, target_api);
-            exports.prepBuildFiles(project_path, 'studio');
+            exports.prepBuildFiles(project_path);
             events.emit('log', generateDoneMessage('create', options.link));
         }).thenResolve(project_path);
 };
--- a/bin/templates/cordova/Api.js
+++ b/bin/templates/cordova/Api.js
@@ -55,23 +55,23 @@ function setupEvents (externalEventEmitter) {
 function Api (platform, platformRootDir, events) {
     this.platform = PLATFORM;
     this.root = path.resolve(__dirname, '..');
-    this.builder = 'studio';
 
     setupEvents(events);
 
-    var self = this;
+    const appMain = path.join(this.root, 'app', 'src', 'main');
+    const appRes = path.join(appMain, 'res');
 
     this.locations = {
-        root: self.root,
-        www: path.join(self.root, 'app/src/main/assets/www'),
-        res: path.join(self.root, 'app/src/main/res'),
-        platformWww: path.join(self.root, 'platform_www'),
-        configXml: path.join(self.root, 'app/src/main/res/xml/config.xml'),
-        defaultConfigXml: path.join(self.root, 'cordova/defaults.xml'),
-        strings: path.join(self.root, 'app/src/main/res/values/strings.xml'),
-        manifest: path.join(self.root, 'app/src/main/AndroidManifest.xml'),
-        build: path.join(self.root, 'build'),
-        javaSrc: path.join(self.root, 'app/src/main/java/'),
+        root: this.root,
+        www: path.join(appMain, 'assets', 'www'),
+        res: appRes,
+        platformWww: path.join(this.root, 'platform_www'),
+        configXml: path.join(appRes, 'xml', 'config.xml'),
+        defaultConfigXml: path.join(this.root, 'cordova', 'defaults.xml'),
+        strings: path.join(appRes, 'values', 'strings.xml'),
+        manifest: path.join(appMain, 'AndroidManifest.xml'),
+        build: path.join(this.root, 'build'),
+        javaSrc: path.join(appMain, 'java'),
         // NOTE: Due to platformApi spec we need to return relative paths here
         cordovaJs: 'bin/templates/project/assets/www/cordova.js',
         cordovaJsSrc: 'cordova-js-src'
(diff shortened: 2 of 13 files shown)


erisu:
### Remaining Tasks
1. Revet adding the parseBuildOptions export.
~~Added parseBuildOptions export so run can get build related options.~~

I originally added this so that the the second build process had all the necessary build options so that I could remove the first build process. Removing the first build process would have been a better workflow but would have affected all platforms.
I decided that it would be better to keep the first build process and remove the second build process instead to minimize the amount of refactoring at this time.

2. Run the remaining tests.

bin/templates/cordova/lib/build.js
```diff
@@ -276,12 +262,10 @@ module.exports.help = function () {
     console.log('Flags:');
     console.log('    \'--debug\': will build project in debug mode (default)');
     console.log('    \'--release\': will build project for release');
-    console.log('    \'--ant\': will build project with ant');
-    console.log('    \'--gradle\': will build project with gradle (default)');
     console.log('    \'--nobuild\': will skip build process (useful when using run command)');
     console.log('    \'--prepenv\': don\'t build, but copy in build scripts where necessary');
-    console.log('    \'--versionCode=#\': Override versionCode for this build. Useful for uploading multiple APKs. Requires --gradle.');
-    console.log('    \'--minSdkVersion=#\': Override minSdkVersion for this build. Useful for uploading multiple APKs. Requires --gradle.');
+    console.log('    \'--versionCode=#\': Override versionCode for this build. Useful for uploading multiple APKs..');
```
janpio: additional dot at the end

bin/lib/create.js
```diff
@@ -143,7 +143,7 @@ function writeProjectProperties (projectPath, target_api) {
 // This makes no sense, what if you're building with a different build system?
 function prepBuildFiles (projectPath, builder) {
```
janpio: second param probably not needed any more here

bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -295,12 +294,18 @@ class ProjectBuilder {
     }
 
     findOutputApks (build_type, arch) {
-        var self = this;
-        return Object.keys(this.binDirs).reduce(function (result, builderName) {
-            var binDir = self.binDirs[builderName];
-            return result.concat(findOutputApksHelper(binDir, build_type, builderName === 'ant' ? null : arch));
+        return Object.keys(this.binDirs).reduce((result, builderName) => {
```
janpio: couldn't `binDirs` and this code be simplified to just a string?

bin/templates/cordova/lib/builders/builders.js
```diff
@@ -17,29 +17,18 @@
...
- * @param   {String}  builderType   Builder name to construct and return. Must
- *   be one of gradle' or 'studio'
- *
- * @return  {Builder}               A builder instance for specified build type.
+ * @return {Builder} A builder instance for specified build type.
  */
-module.exports.getBuilder = function (builderType, projectRoot) {
-    if (!knownBuilders[builderType]) { throw new CordovaError('Builder ' + builderType + ' is not supported.'); }
-
+module.exports.getBuilder = function () {
```
janpio: Couldn't you completely get rid of this? Or is this used from externally?

bin/templates/cordova/lib/builders/ProjectBuilder.js
erisu: Yes, I also think this can be simplified. 

Originally I was going to but decided to wait and clean up the other other areas first. Thanks for reminding me this one. I will take a look at it again.

bin/templates/cordova/lib/builders/builders.js
erisu: I will scan across our repos to verify.

I believe this entire file could be safely removed and point directly to the ProjectBuilder. I am not sure though if anyone uses this in their own personal build system but shouldn't affect us from removing if we don't use.

It might also be possible to do this type of task in separate PR as a final cleanup PR.

erisu: I scanned across repos and it seems only Android uses it. I will say its safe to remove, but I believe a cleanup PR will be better for this so we do not delay the fix. 

I tried changing them and it ended up breaking around 70 tests. I would like to focus on that after.

bin/templates/cordova/Api.js
```diff
@@ -208,17 +208,13 @@ Api.prototype.addPlugin = function (plugin, installOptions) {
...
-    }
-
     return Q().then(function () {
         return PluginManager.get(self.platform, self.locations, project).addPlugin(plugin, installOptions);
     }).then(function () {
         if (plugin.getFrameworks(this.platform).length === 0) return;
         selfEvents.emit('verbose', 'Updating build files since android plugin contained <framework>');
         // This should pick the correct builder, not just get gradle
-        require('./lib/builders/builders').getBuilder(this.builder).prepBuildFiles();
+        require('./lib/builders/builders').getBuilder().prepBuildFiles();
```
dpogue: Probably for the next round of cleaning, but there's a bunch of stuff around here that could be simplified with arrow functions and could use native Promises.

erisu: I also agree. Some of the things I was thinking about in the clean up PR would be:
- arrow functions
- native promises
- `let` and `const`
- Use `fs-extra` to replace things like `shell.cp` etc.
- Remove `builders/builders.js`

bin/templates/cordova/lib/builders/builders.js
janpio: Sure, makes sense to do this individually as it is a major rip-out operation.

bin/templates/cordova/lib/build.js
```diff
@@ -276,12 +262,10 @@ module.exports.help = function () {
     console.log('Flags:');
     console.log('    \'--debug\': will build project in debug mode (default)');
     console.log('    \'--release\': will build project for release');
-    console.log('    \'--ant\': will build project with ant');
-    console.log('    \'--gradle\': will build project with gradle (default)');
     console.log('    \'--nobuild\': will skip build process (useful when using run command)');
     console.log('    \'--prepenv\': don\'t build, but copy in build scripts where necessary');
-    console.log('    \'--versionCode=#\': Override versionCode for this build. Useful for uploading multiple APKs. Requires --gradle.');
-    console.log('    \'--minSdkVersion=#\': Override minSdkVersion for this build. Useful for uploading multiple APKs. Requires --gradle.');
+    console.log('    \'--versionCode=#\': Override versionCode for this build. Useful for uploading multiple APKs.');
```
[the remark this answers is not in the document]
erisu: These flag should still work. 

It appears that there is an issue where the flags are not being accepted. I am not sure when this issue was introduced but it appears that it also existed in the latest release.

I will create a separate ticket and PR to fix CLI parameter issue.

bin/templates/cordova/Api.js
janpio: (Best create a cleanup issue with these tasks so they won't be forgotten)

erisu: Will do.

PR #495, opened 2018-09-12
Incorrect default sdk version issue fix

### LINKED ISSUE #489: Incorrect default min SDK version ternary in build.gradle
I've been trying to work out why the min SDK was always being set to `19` even when I updated it to `23` every where possible.

I believe I've found an issue here https://github.com/apache/cordova-android/blob/master/bin/templates/project/app/build.gradle#L120 where it's setting the `cdvMinSdkVersion` to `null` when it's already `null`, else if it's not `null` it's setting it to the default min SDK version, meaning if you ever set a min SDK it will always change it to the default.

I believe the correct line should read:

```
ext.cdvMinSdkVersion = cdvMinSdkVersion == null ? defaultMinSdkVersion : cdvMinSdkVersion
```

Darshan-Chauhan:
This PR fixes issue [#489](https://github.com/apache/cordova-android/issues/489)

### Checklist
- ~~Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.~~

Resolves #489

### DIFF
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -69,7 +69,7 @@ ext {
     if (!project.hasProperty('cdvBuildMultipleApks')) {
         cdvBuildMultipleApks = null
     }
-    // Whether to append a 0 "abi digit" to versionCode when only a single APK is build 
+    // Whether to append a 0 "abi digit" to versionCode when only a single APK is build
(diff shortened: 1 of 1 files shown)


brodybits:
The change looks right, thanks. I edited the description to update the checklist and automatically close #489 as resolved.

Issue #504, opened 2018-09-21
NDK 18 toolchains


bugsmafia:
From Android NDK v18 removed support:
mips64el-linux-android-4.9
mipsel-linux-android-4.9

cordova run android: fail
"No toolchains found in the NDK toolchains folder for ABI with prefix: mips64el-linux-android"

Loading the old archive NDK v17
https://developer.android.com/ndk/downloads/older_releases
and copy in local folder android-sdk > toolchains

janpio:
Thanks for creating this issue, seems like instructions how to fix things, right?

When exactly is this needed? I never got in contact with Android NDK when creating a Cordova app.

bugsmafia:
> Thanks for creating this issue, seems like instructions how to fix things, right?
> 
> When exactly is this needed? I never got in contact with Android NDK when creating a Cordova app.

Android studio updated my NIR and I had problems with the Cordova-Android

gradle version for Android studio should now be 3.1 and higher

cordova-android v 7.0.0 - creates "build.gradle" classpath 'com.android.tools.build:gradle:3.0.1'

bugsmafia:
https://android.googlesource.com/platform/ndk/+/ndk-release-r18/CHANGELOG.md

> This version of the NDK is incompatible with the Android Gradle plugin version 3.0 or older. If you see an error like No toolchains found in the NDK toolchains folder for ABI with prefix: mips64el-linux-android, update your project file to use plugin version 3.1 or newer. You will also need to upgrade to Android Studio 3.1 or newer.

ubergeek77:
Correct me if I'm wrong, but does this mean cordova-android can't currently build for aarch64? I have the toolchain "aarch64-linux-android-4.9" available on my system, however I receive the same error message, stating that "mips64el-linux-android" is not available. However, I don't want to build for a 64-bit MIPS based device. I want to build an app for my phone, which uses an AArch64 processor.

What do I have to do to get a build going for this architecutre? I've tried adding [all of these config options](https://github.com/MBuchalik/cordova-build-architecture/issues/3#issuecomment-359358657) to no avail.

ndarilek:
[...]

PR #507, opened 2018-09-26
Gradle updates

dpogue:
### Platforms affected
Android

### What does this PR do?
* Updates to latest version of the [Android gradle plugin](https://developer.android.com/studio/releases/gradle-plugin#3-2-0) (3.2.0)
* Updates gradle to the minimum version required by that plugin (4.6)
* Fixes a gradle warning about LeftShift Closure being deprecated
* Fixes a gradle warning about publishNonDefault being deprecated
* Uses the `google()` repo syntax to reference the Google maven repository
* Updates the Github/project URLs for maven artifacts
* Fixes a handful of deprecation warnings happening inside the CordovaLib classes

In theory, this should fix the NDK issues reported in #504 due to outdated gradle versions.

Nothing yet :upside_down_face: 

### Checklist
- [x] Reported an issue: #504

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,7 @@ environment:
     GRADLE_OPTS: -Dorg.gradle.daemon=false
 
     # URL for SDK Tools, Revision 26.1.1 (September 2017)
-    SDK_TOOLS_URL: https://dl.google.com/android/repository/sdk-tools-windows-3859397.zip
+    SDK_TOOLS_URL: https://dl.google.com/android/repository/sdk-tools-windows-4333796.zip
 
     matrix:
         - nodejs_version: 6
@@ -20,7 +20,7 @@ install:
     - set PATH=%PATH%;"%ANDROID_HOME%\tools\bin"
 
     - yes 2> nul | sdkmanager --licenses > nul
-    - sdkmanager "build-tools;26.0.2"
+    - sdkmanager "build-tools;28.0.3"
 
     - choco install gradle --version 3.4.1
 
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -22,33 +22,32 @@ apply plugin: 'com.android.application'
 buildscript {
     repositories {
         mavenCentral()
-        maven {
-            url "https://maven.google.com"
-        }
+        google()
         jcenter()
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.1.0'
+        classpath 'com.android.tools.build:gradle:3.2.0'
     }
 }
 
 // Allow plugins to declare Maven dependencies via build-extras.gradle.
 allprojects {
     repositories {
-        mavenCentral();
+        mavenCentral()
         jcenter()
     }
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '4.4.0'
+    gradleVersion = '4.6.0'
 }
 
 // Configuration properties. Set these via environment variables, build-extras.gradle, or gradle.properties.
 // Refer to: http://www.gradle.org/docs/current/userguide/tutorial_this_and_that.html
 ext {
     apply from: '../CordovaLib/cordova.gradle'
+
     // The value for android.compileSdkVersion.
     if (!project.hasProperty('cdvCompileSdkVersion')) {
         cdvCompileSdkVersion = null;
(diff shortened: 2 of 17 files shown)


appveyor.yml
```diff
@@ -5,7 +5,7 @@ environment:
     GRADLE_OPTS: -Dorg.gradle.daemon=false
 
     # URL for SDK Tools, Revision 26.1.1 (September 2017)
```
raphinesse: I guess the version here is no longer correct with this change

appveyor.yml
```diff
@@ -20,9 +20,9 @@ install:
     - set PATH=%PATH%;"%ANDROID_HOME%\tools\bin"
 
     - yes 2> nul | sdkmanager --licenses > nul
-    - sdkmanager "build-tools;26.0.2"
+    - sdkmanager "build-tools;28.0.3"
 
-    - choco install gradle --version 3.4.1
+    - choco install gradle --version 4.6.0
```
raphinesse: Actually, this should install the minimum version of Gradle needed to generate the wrapper with our actually required version. At least that was true before this change.

appveyor.yml
dpogue: Supposedly 26.1.1 is still the latest, but the download link doesn't match what was here before :confused:

oliversalzburg:
I think you missed to replace one location with `google()`: https://github.com/apache/cordova-android/blob/master/bin/templates/cordova/lib/plugin-build.gradle#L23

oliversalzburg:
I'm also seeing this in the sync log:
```
Configuration 'compile' in project ':app' is deprecated. Use 'implementation' instead.
```
https://stackoverflow.com/questions/47300679/configuration-compile-in-is-deprecated-but-all-configurations-are-implementa says to upgrade gradle to 3.2.0, which resolves the warning for me.

oliversalzburg:
I'm also seeing
```
publishNonDefault is deprecated and has no effect anymore. All variants are now published.
``
Removing https://github.com/apache/cordova-android/blob/master/bin/templates/cordova/lib/plugin-build.gradle#L49 resolves the warning.

**Comment by oliversalzburg** [comment]
The SDK version should be moved from https://github.com/apache/cordova-android/blob/master/bin/templates/project/AndroidManifest.xml#L48 to https://github.com/apache/cordova-android/blob/master/bin/templates/project/app/build.gradle#L165

**Comment by dpogue** [comment]
> The SDK version should be moved from https://github.com/apache/cordova-android/blob/master/bin/templates/project/AndroidManifest.xml#L48 to https://github.com/apache/cordova-android/blob/master/bin/templates/project/app/build.gradle#L165

See https://github.com/apache/cordova-android/issues/508 for additional discussion around that point. 
Short answer is that you're right, longer answer is that it's complicated.

File: test/wrapper.gradle [hunk 3]
```diff
@@ -1,3 +1,3 @@
 wrapper {
-    gradleVersion = '4.4.1'
+    gradleVersion = '4.6'
```
erisu:
```suggestion
    gradleVersion = '4.6.0'
```

Does this need to be `4.6.0`? In the previous file `bin/templates/project/app/build.gradle` you had written it as 

```
wrapper {
    gradleVersion = '4.6.0'
}
```
but this file was different.

dpogue: Ahh, right... so this is used to determine which version of gradle to download, and there is no `4.6.0` release, there is only a `4.6` release.

bin/templates/project/app/build.gradle
```diff
@@ -22,33 +22,32 @@ apply plugin: 'com.android.application'
...
 }
 
 // Configuration properties. Set these via environment variables, build-extras.gradle, or gradle.properties.
 // Refer to: http://www.gradle.org/docs/current/userguide/tutorial_this_and_that.html
 ext {
     apply from: '../CordovaLib/cordova.gradle'
+
     // The value for android.compileSdkVersion.
     if (!project.hasProperty('cdvCompileSdkVersion')) {
         cdvCompileSdkVersion = null;
```
brodybits: Remove semicolon here (as proposed in PR #441)?

bin/templates/project/app/build.gradle
```diff
@@ -22,33 +22,32 @@ apply plugin: 'com.android.application'
...
-        classpath 'com.android.tools.build:gradle:3.1.0'
+        classpath 'com.android.tools.build:gradle:3.2.0'
     }
 }
 
 // Allow plugins to declare Maven dependencies via build-extras.gradle.
 allprojects {
     repositories {
-        mavenCentral();
+        mavenCentral()
```
brodybits: Add google() or replace mavenCentral() with google() here, as suggested in <https://github.com/apache/cordova-android/pull/441/files/478450fe4fe6a7b06e15d2b68facf0aecd63a27e#r228693626>?

bin/templates/project/app/build.gradle
janpio: What does this have to do with this PR?

brodybits:
I raised a few comments based on PR #441. Should we address these?

Is this ready to be merged?

bin/templates/project/app/build.gradle
janpio: What has #441 to do with this?

dpogue:
I believe this is ready to be merged, but I'm not sure what you think it has to do with #441, which is proposing the addition of an entirely different feature.

**This** PR is updating gradle versions for compatibility with the latest Android SDK.  
Kotlin support is out of scope here.

bin/templates/project/app/build.gradle
brodybits: Updates and suggestions in #441 may be relevant here. This PR is to update Gradle in general. #441 was raised to introduce Kotlin support but it did include some general updates that I think we should consider outside of adding the Kotlin support.

brodybits:
I understand that Kotlin is out of scope. But I think some suggestions in #441 should be considered since they seem to be outside the scope of adding Kotlin.

brodybits:
Or should some of the suggestions be considered in yet another PR?

bin/templates/project/app/build.gradle
```diff
@@ -22,33 +22,32 @@ apply plugin: 'com.android.application'
...
         jcenter()
     }
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '4.4.0'
+    gradleVersion = '4.6.0'
```
davidschreiber: Any reason for not using the latest 4.10.2? It should deliver better build speeds and other benefits.

dpogue: Historically we've targeted the minimum required version of gradle

davidschreiber: Is there some specific reason for targeting the minimum version? I don't think that targeting older versions has benefits over targeting newer versions. Also, Gradle development between 4.6 and 4.10 greatly focused on speed

theunreal: Also, as I mentioned in https://github.com/apache/cordova-android/issues/596 - Since Gradle 4.9 it is possible to build App Bundle - a better way to publish apps to the Play Store.
Currently, Cordova doesn't support this and I have to manually change the app manifests after each build.

Issue #534, opened 2018-11-05
User agent check in Cordova.js raises exception in Android 9


pradiv-kumar:
```
androidExec.init = function() {
var check = navigator.userAgent.toLowerCase().match(android\s[0-9].[0-9]);
```

navigator.userAgent in devices running Android Pie returns 9 and not 9.0 and this raises an exception since we try to match the user agent with "android\s[0-9].[0-9]" which expects the user agent to be 9.x

PR #535, opened 2018-11-05
Fix user agent match issue android pie

pradiv-kumar:
### Platforms affected
Android

### What does this PR do?
Fix the user agent match issue in Android Pie devices https://github.com/apache/cordova-android/issues/534

### Checklist
- [x] [Reported an issue] https://github.com/apache/cordova-android/issues/534

### DIFF
--- a/bin/templates/project/assets/www/cordova.js
+++ b/bin/templates/project/assets/www/cordova.js
@@ -977,7 +977,7 @@ androidExec.init = function() {
     //using the Online Event bridge for communicating from Native to JS
     //
     //It's ugly, but it's necessary.
-    var check = navigator.userAgent.toLowerCase().match(/android\s[0-9].[0-9]/);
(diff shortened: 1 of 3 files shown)


brodybits:
Thanks for the contribution. Considering that `check` is only used to check for version range of 4.0...4.3 and minimum Android version is now 4.4 ([[1]](https://cordova.apache.org/docs/en/latest/guide/platforms/android/index.html#requirements-and-support)) I would suggest a much simpler solution: simply remove the lines that check `check` and `version_code`, along with some of the comments above.

Another note is that we generally use the `cordova-coho` tool to generate `bin/templates/project/assets/www/cordova.js`. I would prefer if that you remove the change to `bin/templates/project/assets/www/cordova.js` from this PR.

[1] <https://cordova.apache.org/docs/en/latest/guide/platforms/android/index.html#requirements-and-support>

PR #539, opened 2018-11-08
Fix dest overwrite

kkirby:
### Platforms affected

Android

### What does this PR do?

Makes sure that we don't overwrite the `target-dir` of a `source-file` if the project base is android studio and the `target-dir` is the proper install path and doesn't need to be re-mapped.

I've added a new unit test, confirmed the problem exists, fixed the issue and re-ran the tests.

### Checklist

### More details

Latest release (7.1.2) breaks our build. After investigation it appears that `source-file.install` handler was updated to extract the functionality of path remapping for android studio. However, if the path doesn't need to be remapped, the function returns undefined, thus updating dest to be undefined.

#### Related lines
https://github.com/apache/cordova-android/compare/7a97fd7...7.1.2#diff-4f73e5f70cec312ae82aa7d023e3c88bR320
https://github.com/apache/cordova-android/compare/7a97fd7...7.1.2#diff-4f73e5f70cec312ae82aa7d023e3c88bR50

#### Resolution

Just change line 50 to ` studioPathRemap(obj) || dest`

### DIFF
--- a/bin/templates/cordova/lib/pluginHandlers.js
+++ b/bin/templates/cordova/lib/pluginHandlers.js
@@ -34,7 +34,7 @@ var handlers = {
             // a later plugins release.  This is for legacy plugins to work with Cordova.
 
             if (options && options.android_studio === true) {
-                dest = studioPathRemap(obj);
+                dest = studioPathRemap(obj) || dest;
             }
 
             if (options && options.force) {
--- a/bin/templates/cordova/lib/pluginHandlers.js
+++ b/bin/templates/cordova/lib/pluginHandlers.js
@@ -34,7 +34,7 @@ var handlers = {
             // a later plugins release.  This is for legacy plugins to work with Cordova.
 
             if (options && options.android_studio === true) {
-                dest = studioPathRemap(obj) || dest;
+                dest = getInstallDestination(obj);
             }
 
             if (options && options.force) {
@@ -47,7 +47,7 @@ var handlers = {
             var dest = path.join(obj.targetDir, path.basename(obj.src));
 
             if (options && options.android_studio === true) {
-                dest = studioPathRemap(obj);
+                dest = getInstallDestination(obj);
             }
 
             // TODO: Add Koltin extension to uninstall, since they are handled like Java files
@@ -315,6 +315,11 @@ function removeFileAndParents (baseDir, destFile, stopper) {
 
 function generateAttributeError (attribute, element, id) {
     return 'Required attribute "' + attribute + '" not specified in <' + element + '> element from plugin: ' + id;
+}
+
+function getInstallDestination (obj) {
+    return studioPathRemap(obj) ||
+        path.join(obj.targetDir, path.basename(obj.src));
 }
 
 function studioPathRemap (obj) {
--- a/spec/fixtures/org.test.plugins.dummyplugin/plugin.xml
+++ b/spec/fixtures/org.test.plugins.dummyplugin/plugin.xml
@@ -70,6 +70,8 @@
 
         <source-file src="src/android/DummyPlugin.java"
                 target-dir="src/com/phonegap/plugins/dummyplugin" />
+        <source-file src="src/android/DummyPlugin2.java"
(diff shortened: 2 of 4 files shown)


brodybits:
Can you confirm the following:
* Does the test case show the functionality working on 7.1.1 but not 7.1.2?
* Is this an issue on the master branch?

I think this should solve the issue in #540.

brodybits:
Unfortunately this proposal does not resolve the issue reported in #540.

dpogue:
Thanks! I'd like to get this merged in.

There's one other spot in `uninstall` where the same pattern appears, does that also need to be updated?

brodybits:
> There's one other spot in `uninstall` where the same pattern appears, does that also need to be updated?

I think so, would like to see unit test coverage to ensure it does not break again. Not sure if we need to do this in the hotfix.

This fix is definitely needed in master as well.

jcesarmobile:
Supposedly, studioPathRemap should not remap if the path starts with `app/src/main` as that's how they should start. I think this PR breaks that remapping for old plugins that didn't update the target-dir paths yet

brodybits:
> I think this PR breaks that remapping for old plugins that didn't update the target-dir paths yet

The update is only for the case when `studioPathRemap` returns `null`. Without this change there is a case, reproduced in the proposed test update, where a working "target-dir" path would be overwritten with a null value.

I think it should not break a supported case that is already working. If I am mistaken a specific example would be really helpful.

jcesarmobile:
Yeah, sorry, see my edit, the remap function should not return null, it should return the path unmodified, that’s why I thought it will break it as I only checked the changes, once I checked the code I noticed that was missing

brodybits:
I just pushed changes to add unit test and fix the issue in case of uninstall.

Unfortunately I am not able to rebase this fix onto master since the code in master seems to assume that it always has Android Studio mapping. I will investigate a possible solution.

bin/templates/cordova/lib/pluginHandlers.js
```diff
@@ -34,7 +34,7 @@ var handlers = {
             // a later plugins release.  This is for legacy plugins to work with Cordova.
 
             if (options && options.android_studio === true) {
-                dest = studioPathRemap(obj);
+                dest = studioPathRemap(obj) || dest;
```
brodybits: It is not possible to cherry-pick or rebase this change to work on master. This change will be clobbered by another solution that works more easily on master.

brodybits:
I just pushed an update that would be much easier to cherry-pick between 7.1.x and master. Note that this update basically clobbers the change proposed by @kkirby in ed0991c.

At this point I can think of 2 ways forward:
- we rebase this PR onto master and take it out of PR #542
- we keep this PR on 7.1.x, merge PR #542 to fix on master, and cherry-pick other changes back to 7.1.x

@jcesarmobile please tell me which way you want to proceed with this.

I would like to express appreciation to @kkirby for pointing out the issue and contributing the unit test, even if we cannot use the originally proposed solution.

jcesarmobile:
I prefer the other one as it’s going to be easier to maintain in case we need more hot fixes before we have a 8.0.0 ready

brodybits:
What do you mean by "the other one"?

brodybits:
I just rebased on master (actually pushed the first 3 changes from PR #542 to resolve the original issue on master). Fix should be straightforward to cherry-pick on 7.1.x.

@jcesarmobile please give final confirmation so that I can merge this one master.

brodybits:
P.S. If someone else does a squash merge, please keep the "Co-authored-by" comments to give due credit to both @kkirby and myself.

jcesarmobile:
The “other” was #542, but as you keep changing both I don’t know anymore

brodybits:
Closing now in favor of PR #542. I just put this change back onto the 7.1.x branch (including the original commits by @kkirby).

I am closing this because the proposed change is only a partial solution for the issues we need to fix in #542, to resolve #540 and maybe some other issues with plugins using the old project structure.

@kkirby will absolutely get co-author credit for the added test and solution for the case when `studioPathRemap` returns null.

brodybits:
Now fixed in the latest patch release

Issue #540, opened 2018-11-09
Issue to copy library from plugin after updating to v7.1.2


afdev82:
Everything worked fine in v7.1.1

Now after updating to v7.1.2 I get this error:
```
BUILD FAILED in 7s
3 actionable tasks: 3 executed
/home/antonio/Projects/smartMirror/smartMirrorApk/platforms/android/gradlew: Command failed with exit code 1 Error output:
FAILURE: Build failed with an exception.

* What went wrong:
Could not resolve all files for configuration ':app:releaseCompileClasspath'.
> Could not find :deezer-sdk_library-release:.
  Searched in the following locations:
      file:/home/antonio/Projects/smartMirror/smartMirrorApk/platforms/android/app/libs/deezer-sdk_library-release.aar
  Required by:
      project :app
```

This is the line in my `plugin.xml` for the plugin that needs the library:
```
<source-file src="src/android/libs/deezer-sdk_library-release.aar" target-dir="app/libs/" />
```

I think this could be related to this change: https://github.com/apache/cordova-android/compare/7.1.1...7.1.2#diff-4f73e5f70cec312ae82aa7d023e3c88b

brodybits:
>
>  This is the line in my plugin.xml for the plugin that needs the library:
>
> <source-file src="src/android/libs/deezer-sdk_library-release.aar" target-dir="app/libs/" />
>
> I think this could be related to this change: 7.1.1...7.1.2
> diff-4f73e5f70cec312ae82aa7d023e3c88b
> <https://github.com/apache/cordova-android/compare/7.1.1...7.1.2diff-4f73e5f70cec312ae82aa7d023e3c88b>
>

Link does not work for me.

Can you confirm if you have a test case that shows the functionality
working on 7.1.1 but not on 7.1.2? What about the master branch?

I see you did raise a PR on 7.1.x (<
https://github.com/apache/cordova-android/pull/539>). We would likely need
the same change on master if it is accepted on 7.1.x.

>

brodybits:
>
> I see you did raise a PR on 7.1.x (<
> https://github.com/apache/cordova-android/pull/539>). We would likely need
> the same change on master if it is accepted on 7.1.x.
>


Sorry it was someone else who raise a proposed fix in <
https://github.com/apache/cordova-android/pull/539>.

>

afdev82:
[...]

PR #542, opened 2018-11-09
[WIP] Resolve GH-540 (hotfix)

brodybits:
- include commits with hotfix & unit test proposed in GH-539 (PR #539)
- revert 8a69e32 - CB-14125:(android) Increase old plugin compatibility
- include unit test from PR #541 to verify that GH-540 is resolved

## Explanation

Issues reported in #539 & #540, as reproduced in #541, were introduced in 7.1.2 due to the following changes:

* [CB-13830](https://issues.apache.org/jira/browse/CB-13830): Add handlers for plugins that use non-Java source files, such as Camera (commit ea6477f4d3e63ebb1f1717cc02e3117aa356ce84)
* [CB-14125](https://issues.apache.org/jira/browse/CB-14125): Increase old plugin compatibility (commit 8a69e328004a20810eec4466cbda7fa9013e21ec)

The major change in [CB-13830](https://issues.apache.org/jira/browse/CB-13830) was to factor some Android Studio path remapping code into a `studioPathRemap` function, but it had a small issue that is resolved in PR #539.

The change in [CB-14125](https://issues.apache.org/jira/browse/CB-14125) caused the issue reported in GH-540 and should not have been present in a patch release.

## TODO

One of the following options for the master branch, for the next major release:
1. Include all changes except for revert of [CB-14125](https://issues.apache.org/jira/browse/CB-14125)
2. Include all changes in the master branch

I would favor option 2, which reverts [CB-14125](https://issues.apache.org/jira/browse/CB-14125) in the next major release as well. Old plugins should be updated to reflect the new directory structure.


dpogue:
The change in CB-14125 was one of the main reasons for doing a patch release. There are hundreds of plugins that will never be updated, and we have lots of people stuck on the unsupported Cordova Android 6.x because of that.

brodybits:
> The change in CB-14125 was one of the main reasons for doing a patch release.

Understood, but it did break another plugin as reported in #540 which I think is unacceptable in a patch release and should not be done in a minor release. Any ideas how we can resolve this?

brodybits:
Closing for now, will raise the question on the email forum.

dpogue:
It kinda seems like this new issue is specific to .jar files, so we could maybe special-case them the same way we're handling the .java files?

jcesarmobile:
This is not a breaking change as .jar/.aar files should be in lib-file tags, not in source-file tag. That's what we broke in cordova-android 7.0.0, all plugins that broke was because of usage of source-file tags instead of other proper tags (resource-file for resources, lib-file for libraries, etc)

They "fixed" it for 7.x.x by updating their source-file path from `lib` to `app/lib` in the plugin, so it's not handled properly by the released changes at it doesn't expect them to have `app` on the path.

But not 100% sure our fix is ok for old plugins when using source-file tags for libs, we should check if the new lib path is `app/src/main/lib` or `app/lib`, if it's the second then we have to add an else to handle libs as @dpogue suggested as current code will put them in `app/src/main/`+the `target-dir`

brodybits:
Reopening as a WIP PR for now. I will take another look over the weekend.

I think PR #539 should be merged by itself, keeping the commits here for now since they are needed to help resolve GH-540.

A side point is that I did not see any test updates for CB-14125 (unit tests pass whether I keep or revert CB-14125 in my local workarea). Please correct me if I am mistaken. I think tests are really needed to help ensure it does not break in the future.

brodybits:
I just updated with a simple solution: do not do the path remapping if target-dir="app/lib" in source-file element. CB-14125 is no longer reverted in this proposal.

Since use of target-dir="app/lib" indicates that the plugin is using the new file structure, I think there should be no reason to mess with it.

I think this should not break anything. If I am mistaken a specific case would be really helpful.

An even simpler solution may be to not do the path remapping if the target-dir starts with "app".

Test case is still needed for uninstall case. I can work on it over the weekend.

brodybits:
Updated as follows:
- rebased on master
- added uninstall tests
- Solution for GH-539 needed rewrite to work on master branch
- Solution for GH-540 reworked to have effect on any target-dir with "app" target-dir

jcesarmobile:
Looks good, but we still need to handle it ends in .jar or .aar to remap to app/target-dir instead of app/src/main/target-dir. Or edit gradle to look for libraries there too.
But it can be done in a separate PR

brodybits:
> we still need to handle it ends in .jar or .aar to remap to app/target-dir instead of app/src/main/target-dir

Can you think of any examples of plugins where we need to do this?

> Or edit gradle to look for libraries there too.

I would vote against this option. I suspect it would be easier for this to go wrong and possibly harder to fix in case it does. I also suspect it would be harder to get this option right in a patch release. But I may be mistaken, still not a Gradle expert.

brodybits:
> But it can be done in a separate PR

I would favor that.

I am now waiting for final approval to merge #539 before merging this one. Quick review of #544 would also be appreciated.

Again I would really appreciate examples of plugins where we need to fix mapping of target-dir for JAR & AAR.

My time will be pretty limited for a couple days due to other commitments, should have some more time afterwards.

jcesarmobile:
All the plugin examples linked by @dpogue have .jar source-files except for the last one.

They also happen to have .so files, not sure how we should handle them, I think proper folder is `app/src/main/jniLibs`, but most of those plugins were just putting it in libs folder, so should be `app/libs`.

I think we should fix .jar and .aar for now in a separate PR and investigate the .so

afdev82:
> This is not a breaking change as .jar/.aar files should be in lib-file tags, not in source-file tag. That's what we broke in cordova-android 7.0.0, all plugins that broke was because of usage of source-file tags instead of other proper tags (resource-file for resources, lib-file for libraries, etc)

Didn't know that, thanks.
I can update the plugin to use the right tag in the next release.

Issue #547, opened 2018-11-12
Compatibility of old plugins with non-Java source-file entries (individual files)


brodybits:
From <https://github.com/apache/cordova-android/pull/542#issuecomment-437709608> & <https://github.com/apache/cordova-android/pull/542#issuecomment-437922322>:

Some old plugins with non-Java source-file entries still do not seem to work since version 7.0.0. Some examples, including examples from <https://github.com/apache/cordova-android/pull/542#issuecomment-437922322>:
* https://github.com/dylearning/cordova-plugin-hikvisionVedio
* https://github.com/wyne1986/cordova-plugin-sip-fjh
* https://github.com/mul1sh/corewise-plugin-fingerprintauth
* https://github.com/ispikit/ispikit-cordova
* `cordova-sqlite-storage@2.1.5` (not an issue since `cordova-sqlite-storage@2.2.0`)
* ~~`https://github.com/j3k0/cordova-plugin-purchase#v7.1.0` (version without any of the adaptations made to work with cordova-android@7, was part of #546)~~ (old version will not be supported)
* `https://github.com/adjust/cordova_sdk#master` (see #554)

Example file types to deal with here:
* `.jar` - generally has target-dir="lib" in old plugins
* `.aar` - also generally has target-dir="lib" in old plugins
* `.so` - generally has a "lib" subdirectory in old plugins

Purpose is that a massive number of plugins should be able to just work on both old and new Android project structure.

brodybits:
P.S. This *may* affect <https://github.com/apache/cordova-docs/issues/902>: Document how to use Android NDK libraries in plugins

brodybits:
Reproduction steps in <https://github.com/brodybits/cordova-sqlite-test-app>:
* `cordova plugin add cordova-sqlite-storage@2.1.5`
* `cordova platform add android@7.1.2`
* `cordova build android`

Then I get errors such as:
- `/Users/brodybits/Documents/cordova/cordova-sqlite-test-app/platforms/android/app/src/main/java/io/sqlc/SQLiteConnectorDatabase.java:29: error: package io.liteglue does not exist`
- `/Users/brodybits/Documents/cordova/cordova-sqlite-test-app/platforms/android/app/src/main/java/io/sqlc/SQLiteConnectorDatabase.java:41: error: cannot find symbol`

It worked when I did `cordova plugin add cordova-sqlite-storage@2.1.5` and `cordova platform add android@6`.

[...]

PR #550, opened 2018-11-13
Fix for old plugins with non-Java sources (GH-547)

brodybits:
## Proposed changes

- Unit tests to cover GH-547 for AIDL, AAR, JAR, and `.so` file with old target-dir scheme (cordova-android pre-7.0)
- Cleanup `getInstallDestination` function in `pluginHandlers.js`
- Fix mapping of AIDL files in `src` subdirectory
- Fix mapping of AAR, JAR, and `.so` files in `libs` subdirectory, `.so` file subdirectory mapping according to <https://github.com/dpa99c/cordova-plugin-hello-c/blob/master/plugin.xml#L39-L42>

The proposed changes will be rebased along with PR #542 (fix for GH-540) on 7.1.x if accepted.

### npm test

- [x] `npm run unit-tests` passes
- [x] `npm test` passes on  Travis CI
- [x] `npm test` passes on AppVeyor CI

### cordova-sqlite-storage 2.1.5

`cordova-sqlite-storage@2.1.5` has old JAR and `.so` target-dir scheme, which does not work on cordova-android@7.0.0 (I already make a workaround in `cordova-sqlite-storage@2.2.0`). But `cordova-sqlite-storage@2.1.5` works on `https://github.com/brodybits/cordova-android#gh-547-bugfix` with the proposed bug fix.

How tested in clone of <https://github.com/brodybits/cordova-sqlite-test-app>:
* `cordova plugin add cordova-sqlite-storage@2.1.5`
* `cordova platform add https://github.com/brodybits/cordova-android#gh-547-bugfix`
* Able to build and run on a test device
* Click on self test button results in an alert that the self test is OK (self test of SQLite CRUD operations in sqlite plugin)
* `cordova plugin remove cordova-sqlite-storage` is successful

### cordova-plugin-purchase v7.1.0

This `cordova-plugin-purchase` version has AIDL & XML files with old target-dir scheme, does not work on cordova-android@7.0.0. They had to do some ugly workarounds to work on both cordova-android@6 and cordova-android@7.

How tested in a new Cordova project:

* `cordova plugin add https://github.com/j3k0/cordova-plugin-purchase#v7.1.0 --variable BILLING_KEY="MIIB...AQAB"`
* `cordova platform add https://github.com/brodybits/cordova-android#gh-547-bugfix`
* `cordova build android` is successful
* `cordova plugin remove cc.fovea.cordova.purchase` is successful

### cordova-plugin-hikvisionVedio

<https://github.com/dylearning/cordova-plugin-hikvisionVedio> fails to build on cordova-android@7.1.2 since its JAR files as specified in `source-file` elements not installed in the right place.

How tested in a new Cordova project:

* `cordova plugin add https://github.com/dylearning/cordova-plugin-hikvisionVedio`
* `cordova platform add https://github.com/brodybits/cordova-android#gh-547-bugfix`
* `cordova build android` is successful
* `cordova plugin rm cordova-plugin-hikvisionVedio` is successful

## Quirk

JAR library files are mapped into `app/lib` subdirectory, while `.so` (NDK) library files are mapped into `app/src` subdirectory. I wonder if there is an easy way to resolve this inconsistency.

I think we should encourage plugins to use `lib-file` instead of `source-file` for AAR, JAR, and `.so` files, and add a deprecation message in case mapping of these library files is needed. I hope to propose this change in the near future.

## Next steps

- [ ] Rebase this fix along with PR #542 (GH-540) on 7.1.x for upcoming 7.1.3 patch release
- [ ] Tests should be renumbered in handlers.spec.js, maybe with suffix letters like they did in cordova-windows for greater future flexibility
- [ ] Add a deprecation message in case of mapping of library files as described above


jcesarmobile:
Looking at the plugin you linked in the comment, it also has .h and .c files that are treated as .java files.
They were `src/c/common/mylib/parts` and he updated them to `app/src/main/java/c/common/mylib`

We could add those two extensions in the .java check.

brodybits:
> `app/src/main` is used multiple times, might be worth to put it in a variable

I can do this, would favor doing the same thing with some other prefixes such as `app` for the sake of consistency. Ultimate solution would be to do this for all destination prefixes.

> [...] it also has .h and .c files that are treated as .java files.
> They were `src/c/common/mylib/parts` and he updated them to `app/src/main/java/c/common/mylib`
> 
> We could add those two extensions in the .java check.

Interesting idea, how should we test it? Anything beyond a unit test?

I will try to do the rework later tonight, would like to get it into the 7.1.3 patch if possible.

jcesarmobile:
I don’t know any plugin with such files (other than the one you linked and it’s fixed already), so unit tests should be enough

brodybits:
> [...] it also has `.h` and `.c` files that are treated as `.java` files.
> They were `src/c/common/mylib/parts` and he updated them to `app/src/main/java/c/common/mylib`
> 
> We could add those two extensions in the .java check.

I would not favor making a special case for these files. Right now cordova-android knows nothing about these files by itself and I think we should keep it that way. If a plugin wants to put these files in a special location I think it should specify the full prefix starting with `app`. I will add them to the unit test, though.

jcesarmobile:
It knows nothing about any file extension at all, this PR is the one placing files based on the extension

brodybits:
I think the primary rule should be to place files based on target-dir prefix, i.e. `app` vs `src` vs `lib`, and extensions should be treated as special cases. This is basically how it is implemented in this PR.

I would rather not make a special case for `.h` and `.c` files since cordova-android does not need to do anything with them. Use of `.h` and `.c` files should remain plugin-specific, plugin provides its own Gradle file to handle them.

For example: <https://github.com/dpa99c/cordova-plugin-hello-c> specifies the native build with `.h` and `.c` files in the following places:
* <https://github.com/dpa99c/cordova-plugin-hello-c/blob/master/plugin.xml#L47> to use `build-extras.gradle` (probably should have been a `framework` element)
* <https://github.com/dpa99c/cordova-plugin-hello-c/blob/master/src/android/build-extras.gradle#L8-L20> to use `Android.mk`
* <https://github.com/dpa99c/cordova-plugin-hello-c/blob/master/src/android/jni/Android.mk#L21-L36 specifies the actual source files for NDK build, including files shared with iOS platform

I would be happy to add these files to the unit test, for the sake of verification and understanding.

jcesarmobile:
So, what's the point of adding them to the tests if we are not handling them? There is already a test for the files that don't fall into any of the elses, so there is no point of adding other one

brodybits:
> `app/src/main` is used multiple times, might be worth to put it in a variable

Done. CanI get a final approval to merge?

jcesarmobile:
I still see the thanks comment&link, can you remove it?

brodybits:
Just removed the comment, my bad.

brodybits:
I just added these changes to PR #555 on 7.x for a patch release, hope we can make it soon!

janpio:
How can you add a non-merged PR to a new PR? 
Did you just cherry pick the individual commits? 
If so, I would advise against that (which would act as a "do not merge" blocker on the other PR) as you do not really know if this PR will be merged that way - especially as this one here should probably be squashed.

brodybits:
@janpio I was thinking to do a merge commit to keep the cleanup changes separate from the behavior change. I think it makes it easier to understand, and less to potentially revert just in case. If anyone has objections please do not hesitate.

In general I have been testing the changes on both master and 7.1.x to help ensure the patch goes smoothly. Proposal for patch updates in #555 is marked WIP to avoid a premature merge. In case this PR is not merged for some reason it would be really easy to remove the unwanted commits from #555.

Considering that this change has gone through review and is waiting for final approval, I would like to get it into the upcoming patch if possible. I think it would be a major benefit for a number of existing plugins. I hope we can merge and release soon. I put some client work on hold, at personal cost, in order to make the fixes and get some plugins working again.

brodybits:
Thanks @jcesarmobile for the detailed review!

janpio:
Was @jcesarmobile's feedback on how to merge or not merge this taken into account?

janpio:
So, No.

> I think the biggest downside to combining commits is in case there is a need to revert some but not all of the changes for some reason.

I prefer a clean and comprehensible commit history versus already planning having to revert stuff.

Issue #552, opened 2018-11-14
Cordova CLI project ignores build-extras after toolchain upgrade


DavidWiesner:

brodybits:
Please show exactly exactly what you did in plugin.xml for both 6.4 and 7.1.2. I think we will have to fix `bin/templates/cordova/lib/pluginHandlers.js` to resolve this issue without introducing a breaking change.

A major bonus is if you can show us a plugin we can use to test the solution.

DavidWiesner:
In my case I did nothing with plugins. I just add a `platforms/android/build-extras.gradle` file manually adding some dependencies.
## Steps for reproduce:
1. add a `platforms/android/build-extras.gradle` file with this content 
```
ext.postBuildExtras = { println('### build-extras.gradle called ###') }
```
2. run `cordova build android` 
3. no `### build-extras.gradle called ###` was printed in the console

adding the same file to `platforms/android/app/` will print the line

brodybits:
That is because the Android project structure was changed in 7.0. I just raised <https://github.com/apache/cordova-docs/issues/905> to update the documentation.

DavidWiesner:
I know the structure has changed. But why also change the location of the build-extras.gradle? All other configuration files e.g.: release-signing.properties are still referenced from the old location: platforms/android/

janpio:
> But why also change the location of the build-extras.gradle?

That was probably overlooked when changing the project structure back in 7.x. You would have to go back to those commits where the other paths were handled and see if there maybe is a comment in the PR or commit message.

DavidWiesner:
Here is one [issue (CB-13615)](https://issues.apache.org/jira/browse/CB-13615) reflecting the new structure and changing the paths accordingly. But this is only for the signing configurations.

brodybits:
From the history it looks like the work was done in PR #389. It seems like they did try to preserve compatibility, but evidently it was not preserved 100%.

> All other configuration files e.g.: release-signing.properties are still referenced from the old location: platforms/android/

I find that really interesting, just discovered that `release-signing.properties` is referenced by both `bin/templates/project/app/build.gradle` and `bin/templates/project/legacy/build.gradle`. I have no idea how or even if they would work together.

PR #553, opened 2018-11-14
#552: (android) check for build-extras.gradle in the app-parent directory

DavidWiesner:
### Platforms affected
- android

### What does this PR do?
- solves issue #552. Therefor include the build-extras.gradle from `platforms/android/`
- If requested I can change this pull request to check on both places `platforms/android/` and `platforms/android/app/` 

- in a local build environment

### Checklist

### DIFF
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -93,9 +93,9 @@ ext {
 // PLUGIN GRADLE EXTENSIONS START
 // PLUGIN GRADLE EXTENSIONS END
 
-def hasBuildExtras = file('build-extras.gradle').exists()
+def hasBuildExtras = file('../build-extras.gradle').exists()
 if (hasBuildExtras) {
-    apply from: 'build-extras.gradle'
+    apply from: '../build-extras.gradle'
 }
 
 // Set property defaults after extension .gradle files.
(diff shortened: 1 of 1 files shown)


janpio:
This would be a breaking change right now, so I guess it would be nicer to just check both locations?

brodybits:
I think we will need to adapt pluginHandlers.js instead. Followup question is coming on #552.

brodybits:
I just pushed an update to support both locations, didn't test it yet.

DavidWiesner:
Ok I can update the pull request. Wich file should win in the case both files existing? Should the apply exclusively or in sequence if both files existing?

brodybits:
Good question. I thought we should run both files, didn't think about the ordering.

I would personally like to see the old location deprecated at some point.

If my commit doesn't help then feel free to remove it with a force push. I am still not a real Gradle expert.

And we should update the documentation as well. If I follow your link and click the edit button it takes me to the following file: <https://github.com/apache/cordova-docs/blob/master/www/docs/en/dev/guide/platforms/android/index.md>

@janpio can you think of anything else?

DavidWiesner:
I think your commit is just fine. This was my prefered order and method. I hope the old location will stay. It was luck that I found this bug before shipping our alpha build to public. Otherwise ~2000 people local settings and database get lost.

PR #559, opened 2018-11-15
Ignore version file in eslint to resolve build error

brodybits:
This change should to give us a green build again.

Rationale is that the eslint update introduced in 393dad6349d8df55d8fca7d3e5014f36dfa4610c checks most of the files under `bin` and will not accept JavaScript that uses double-quotes for strings. The workaround in 393dad6349d8df55d8fca7d3e5014f36dfa4610c was to change the generated `bin/templates/cordova/version` file to use single-quotes in the version string.

But since cordova-coho will regenerate `bin/templates/cordova/version` with double-quotes when updating the version number, we need a different solution.

This is a quick workaround solution. I won't have a problem if anyone wants to propose an improved solution. Alternative may be to update cordova-coho to generate version file with single-quotes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "java-unit-tests": "node test/run_java_unit_tests.js",
     "eslint": "run-s -c eslint:*",
     "eslint:scripts": "eslint bin spec test",
-    "eslint:bins": "eslint 'bin/**/*' --ignore-pattern '**/*.*' --ignore-pattern '**/gitignore'"
+    "eslint:bins": "eslint 'bin/**/*' --ignore-pattern '**/*.*' --ignore-pattern '**/gitignore' --ignore-pattern 'bin/templates/cordova/version'"
   },
   "author": "Apache Software Foundation",
   "license": "Apache-2.0",


dpogue:
There might be an eslint ignore comment we can wrap around the version line to suppress the error?

brodybits:
> There might be an eslint ignore comment we can wrap around the version line to suppress the error?

That would involve updating the cordova-coho tool. I would rather not spend time on updating cordova-coho right now.

brodybits:
Build is green, merging now

brodybits:
> That would involve updating the cordova-coho tool.

I may be wrong here. Not sure if I would favor an extra eslint comment though. Build of `master` branch is green again, moving on.

dpogue:
> I may be wrong here. Not sure if I would favor an extra eslint comment though. Build of `master` branch is green again, moving on.

If the file is entirely generated by coho, then the best option is probably looking at how to generate it with the right quotes. The eslint comment suggestion was assuming that coho was just doing a find-and-replace on it in an existing file.

janpio:
Here is the actual code that should have been updated to fix this problem instead of polluting `cordova-android`: https://github.com/apache/cordova-coho/blob/2ae50f2200c352f2a69b731a79f10562c5c41bc6/src/versionutil.js#L101-L122

PR #563, opened 2018-11-16
fix(build): support tilde expansion on windows

hiepxanh:
### Platforms affected
Android, ios, Browser

### What does this PR do?
I struggeling with home path every time changing build computer or transfer code to my teammate computer. the `build.json` config does not find home path on window with "command line" or "power shell". there is no tutorial about this. I have to go to build file to know that there is `~` equivalent with `$HOME` and `%HOMEPATH%` on window.
I think it is feature?
### Checklist

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -20,6 +20,7 @@
 var path = require('path');
 var fs = require('fs');
 var nopt = require('nopt');
+const untildify = require('untildify');
 
 var Adb = require('./Adb');
 
@@ -82,9 +83,7 @@ function parseOpts (options, resolvedTarget, projectRoot) {
         if (config.android && config.android[ret.buildType]) {
             var androidInfo = config.android[ret.buildType];
             if (androidInfo.keystore && !packageArgs.keystore) {
-                if (androidInfo.keystore.substr(0, 1) === '~') {
-                    androidInfo.keystore = process.env.HOME + androidInfo.keystore.substr(1);
-                }
+                androidInfo.keystore = untildify(androidInfo.keystore);
                 packageArgs.keystore = path.resolve(path.dirname(buildConfig), androidInfo.keystore);
                 events.emit('log', 'Reading the keystore from: ' + packageArgs.keystore);
             }
--- a/package-lock.json
+++ b/package-lock.json
@@ -3202,6 +3202,11 @@
       "resolved": "https://registry.npmjs.org/universalify/-/universalify-2.0.0.tgz",
       "integrity": "sha512-hAZsKq7Yy11Zu1DE0OzWjw7nnLZmJZYTDZZyEFHZdUhV8FkH5MCfoU1XMaxXovpyW5nq5scPqq0ZDP9Zyl04oQ=="
     },
+    "untildify": {
+      "version": "4.0.0",
+      "resolved": "https://registry.npmjs.org/untildify/-/untildify-4.0.0.tgz",
+      "integrity": "sha512-KK8xQ1mkzZeg9inewmFVDNkg3l5LUhoq9kN6iWYB/CC9YMG8HA+c1Q8HwDe6dEX7kErrEVNVBO3fWsVq5iDgtw=="
+    },
     "uri-js": {
       "version": "4.4.1",
       "resolved": "https://registry.npmjs.org/uri-js/-/uri-js-4.4.1.tgz",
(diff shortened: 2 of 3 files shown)


bin/templates/cordova/lib/build.js
```diff
@@ -82,7 +82,9 @@ function parseOpts (options, resolvedTarget, projectRoot) {
         if (config.android && config.android[ret.buildType]) {
             var androidInfo = config.android[ret.buildType];
             if (androidInfo.keystore && !packageArgs.keystore) {
-                if (androidInfo.keystore.substr(0, 1) === '~') {
+                if (androidInfo.keystore.substr(0, 1) === '~' 
+                    || androidInfo.keystore.substr(0, 5) === '$HOME' 
+                    || androidInfo.keystore.substr(0, 10) === '%HOMEPATH%') {
                     androidInfo.keystore = process.env.HOME + androidInfo.keystore.substr(1);
```
janpio: This line is now broken as `androidInfo.keystore.substr(1)` removes the first character from the keystore string - which only works for `~`

janpio:
Hm, first time I see this. Is the `~` behavior documented anywhere? Where did you learn that you can do it this way?

bin/templates/cordova/lib/build.js
hiepxanh: I dont know how it can be broken?

hiepxanh:
> Hm, first time I see this. Is the ~ behavior documented anywhere? Where did you learn that you can do it this way?
=> It take me to 5 hours to learn that there is no documented or any suggestion from google. I read your source code. The document is lack of this. 
The situation when you have a team with alot of computer, you have different home path.

> Also: Isn't the code here currently already exactly for making it possible to use ~ on Windows, where this isn't supported out of the box?
=> yes it support windows. that very nice, accept: 
not every body install `$HOME` enviroment variable as home path. Only after I read source code. I found out that I have to use `~` and it match to variable `$HOME`. Then I have to update my window enviroment variable with this path. 

that take a lot of time.

bin/templates/cordova/lib/build.js
hiepxanh: oh I see that, I should make an other check

hiepxanh: I fix that

janpio:
After writing this, I decided to look into why this tilde code is there in the first place:
https://github.com/apache/cordova-android/commit/d78ae309f1234cec4f1651a7140508ff1b1dda32
https://issues.apache.org/jira/browse/CB-10105
Maybe this is not needed any more in the first place as the problem in node was fixed?

Anyway, the usage of environment variables in the path itself is a totally new and different feature and I am not sure this is the way to go.

brodybits:
Thanks @hiepxanh for the proposal, my response is in <https://github.com/apache/cordova-cli/pull/359#issuecomment-439707679>.

janpio:
Thanks @raphinesse for this necessary context. 

So if we decide to keep this functionality, it would probably make sense to switch over to using this package (any maybe make this available to all platforms along the way?). The package uses `os.homedir()` internally, which is probably a _lot_ more stable than any hack we can place in our code ourselves. Agree?

brodybits:
> So if we decide to keep this functionality, it would probably make sense to switch over to using this package (any maybe make this available to all platforms along the way?).

Agreed on my part and I would definitely favor making this available to all platforms. (I was about to suggest the same idea.)

janpio (dismissed):
suggested changes don't actually solve the underlying problem and create more unrelated problems

brodybits:
> suggested changes don't actually solve the underlying problem [...]

I think that means changes suggested 2 days ago and not the recent suggestion to use untildify.

janpio:
Pull Request reviews relate to the code being suggested in the Pull Request, so yes - I just made sure it is clear that this should not be merged by anyone.

hiepxanh:
@janpio @brodybits thank you for reviewing my small idea. I think this will be helpful and help alot of people like me. Have to build code with team have different develop enviroment. I agree that my solution not work with all platforms and not a global solution. So I think that we should use @raphinesse suggested library. It will be nice. 
thank you for helping me. Should I try to use that library in my code then you can continue review or you will help me to do that and I will close this issue?

raphinesse (approved):
I redid the fix using untildify

erisu (approved):

PR #568, opened 2018-11-21
Move the PLUGIN GRADLE EXTENSIONS comments

### LINKED ISSUE #569: SDK versions are resolved after plugins gradle extensions
I also noticed that gradle extensions get injected before these versions are determined, meaning any plugin relying on these sees them as `null`.

```
// ...

// PLUGIN GRADLE EXTENSIONS START
apply from: "../cordova-plugin-browsertab/mobile-BrowserTab.gradle"
// PLUGIN GRADLE EXTENSIONS END

def hasBuildExtras = file('build-extras.gradle').exists()
if (hasBuildExtras) {
    apply from: 'build-extras.gradle'
}

// Set property defaults after extension .gradle files.
if (ext.cdvCompileSdkVersion == null) {
    ext.cdvCompileSdkVersion = privateHelpers.getProjectTarget()
    //ext.cdvCompileSdkVersion = project.ext.defaultCompileSdkVersion
}
if (ext.cdvBuildToolsVersion == null) {
    ext.cdvBuildToolsVersion = privateHelpers.findLatestInstalledBuildTools()
    //ext.cdvBuildToolsVersion = project.ext.defaultBuildToolsVersion
}
if (ext.cdvDebugSigningPropertiesFile == null && file('../debug-signing.properties').exists()) {
    ext.cdvDebugSigningPropertiesFile = '../debug-signing.properties'
}
if (ext.cdvReleaseSigningPropertiesFile == null && file('../release-signing.properties').exists()) {
    ext.cdvReleaseSigningPropertiesFile = '../release-signing.properties'
}

// Cast to appropriate types.
ext.cdvBuildMultipleApks = cdvBuildMultipleApks == null ? false : cdvBuildMultipleApks.toBoolean();
ext.cdvVersionCodeForceAbiDigit = cdvVersionCodeForceAbiDigit == null ? false : cdvVersionCodeForceAbiDigit.toBoolean();
ext.cdvMinSdkVersion = cdvMinSdkVersion == null ? defaultMinSdkVersion : cdvMinSdkVersion
ext.cdvVersionCode = cdvVersionCode == null ? null : Integer.parseInt('' + cdvVersionCode)

//...
```

tombell:
### Platforms affected

Android

### What does this PR do?

Move the comment block for where gradle plugins from extensions are injected. This means they have proper access to correct version variables which are currently resolved after they are injected. I noticed this due to using https://github.com/google/cordova-plugin-browsertab which has this https://github.com/google/cordova-plugin-browsertab/blob/master/plugin/src/android/BrowserTab.gradle which always sets to 16, because the versions are resolved after plugins. This broke our code because we set our minimum SDK, however this plugin would force it lower to 16, due to the version code resolution being after.

Fix #569

Testing in our application, and run the tests in the repository. Existing test coverage _should_ cover this change, because it's just an ordering change.

### Checklist

### DIFF
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -90,9 +90,6 @@ ext {
     cdvPluginPostBuildExtras = []
 }
 
-// PLUGIN GRADLE EXTENSIONS START
-// PLUGIN GRADLE EXTENSIONS END
-
 def hasBuildExtras1 = file('build-extras.gradle').exists()
 if (hasBuildExtras1) {
     apply from: 'build-extras.gradle'
@@ -124,6 +121,9 @@ ext.cdvBuildMultipleApks = cdvBuildMultipleApks == null ? false : cdvBuildMultip
 ext.cdvVersionCodeForceAbiDigit = cdvVersionCodeForceAbiDigit == null ? false : cdvVersionCodeForceAbiDigit.toBoolean();
 ext.cdvMinSdkVersion = cdvMinSdkVersion == null ? defaultMinSdkVersion : Integer.parseInt('' + cdvMinSdkVersion)
 ext.cdvVersionCode = cdvVersionCode == null ? null : Integer.parseInt('' + cdvVersionCode)
+
+// PLUGIN GRADLE EXTENSIONS START
+// PLUGIN GRADLE EXTENSIONS END
 
 def computeBuildTargetName(debugBuild) {
     def ret = 'assemble'


brodybits:
Thanks @tombell for the proposal. I don't really understand how moving the comments would help solve a problem with extensions that are injected.

From what I see I think I would actually favor keeping the "EXTENSIONS START" comment where it is and moving the "EXTENSIONS END" as you proposed.

As a general comment I still do not have a super clear understanding of the structure and flow of the cordova-android Gradle files. Comments with a better explanation or maybe a better cleanup someday would be nice.

dpogue:
@brodybits When installing plugins, any gradle files that the plugins reference are added to the build.gradle file in between the `EXTENSIONS START` and `EXTENSIONS END` comments. There must not be any actual code in between those comments otherwise it will get overwritten during the prepare step.

I'm experimenting with/reviewing how cordova-android's gradle stuff is set up, and will potentially be opening a PR to reorganize it, taking open issues and PRs like this one into account.

tombell:
@brodybits basically how @dpogue explained it, below is a snippet of my `build.gradle` after adding the platform with my plugins:

```gradle

// PLUGIN GRADLE EXTENSIONS START
apply from: "../cordova-plugin-browsertab/mobile-BrowserTab.gradle"
// PLUGIN GRADLE EXTENSIONS END

```

So you can see the plugin stuff is added between the two comments, I had to manually move that block below, as per this PR, so that the plugin `.gradle` file had access to the correct version values.

brodybits:
I have a couple major criticisms now:
* Moving these comments *could* prove to be a breaking change, probably in a very limited number of cases. It should be OK in a major release if we document it properly. I would rather we take some time to consider the possible consequences before we apply the change.
* I think the root cause is that the plugin sets the minimum SDK version in a non-standard manner. It should have just set `android:minSdkVersion` in `uses-sdk` element as documented in <https://cordova.apache.org/docs/en/latest/plugin_ref/spec.html#edit-config>.

I think it would be ideal if we could make a separate Gradle file for the plugin Gradle extensions rather than having the scripts edit `build.gradle` in-place.

tombell:
@brodybits I'm happy to close this off, in favour of using post-platform add scripts to modify my own `build.gradle` for now, until a better solution comes up.

brodybits:
Thanks @tombell, I think that sounds like a good plan. I just labeled this as a bug in #569, along with a brief explanation.

PR #572, opened 2018-11-22
Resolve issue with plugin target-dir="app*" subdirs

### LINKED ISSUE #571: Plugin is not correctly installed if the package contains "app"
Example from `cordova-plugin-local-notifications` https://github.com/katzer/cordova-plugin-local-notifications
`src/de/appplant/cordova/plugin/localnotification`

We should probably check for `"app/"`, but might still affect others. Maybe use `startsWith` instead of `includes`?

brodybits:
(subdirectories) such as "appco", with unit tests to verify correct operation

Needed for @katzer plugins that use `de/appplant` subdirectory, for example:
* <https://github.com/katzer/cordova-plugin-local-notifications>, with `target-dir="src/de/appplant/cordova/plugin/notification"`
* <https://github.com/katzer/cordova-plugin-badge>, with `target-dir="src/de/appplant/cordova/plugin/badge"`
* <https://github.com/katzer/cordova-plugin-background-mode>, with `target-dir="src/de/appplant/cordova/plugin/background"`

Also needed for `cordova-plugin-inappbrowser`

Resolves #571

### DIFF
--- a/spec/unit/pluginHandlers/handlers.spec.js
+++ b/spec/unit/pluginHandlers/handlers.spec.js
@@ -164,7 +164,7 @@ describe('android project handler', function () {
                     path.join('app/src/main/jniLibs/x86/libnative.so'), false);
             });
 
-            it('Test#006j : should allow installing sources with target-dir that includes "appco"', function () {
+            it('Test#006j : should allow installing sources with target-dir that includes "app"', function () {
                 android['source-file'].install(valid_source[10], dummyPluginInfo, dummyProject, {android_studio: true});
                 expect(copyFileSpy)
                     .toHaveBeenCalledWith(dummyplugin, 'src/android/DummyPlugin2.java', temp, path.join('app/src/main/java/com/appco/DummyPlugin2.java'), false);
--- a/spec/unit/pluginHandlers/handlers.spec.js
+++ b/spec/unit/pluginHandlers/handlers.spec.js
@@ -163,6 +163,12 @@ describe('android project handler', function () {
                     'src/android/jniLibs/x86/libnative.so', temp,
                     path.join('app/src/main/jniLibs/x86/libnative.so'), false);
             });
+
+            it('Test#006j : should allow installing sources with target-dir that includes "appco"', function () {
+                android['source-file'].install(valid_source[10], dummyPluginInfo, dummyProject, {android_studio: true});
+                expect(copyFileSpy)
+                    .toHaveBeenCalledWith(dummyplugin, 'src/android/DummyPlugin2.java', temp, path.join('app/src/main/java/com/appco/DummyPlugin2.java'), false);
+            });
         });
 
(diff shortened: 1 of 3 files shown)


spec/unit/pluginHandlers/handlers.spec.js
```diff
@@ -163,6 +163,12 @@ describe('android project handler', function () {
                     'src/android/jniLibs/x86/libnative.so', temp,
                     path.join('app/src/main/jniLibs/x86/libnative.so'), false);
             });
+
+            it('Test#006j : should allow installing sources with target-dir that includes "appco"', function () {
```
janpio: The bug was that it included `app`, right?
```suggestion
            it('Test#006j : should allow installing sources with target-dir that includes "app"', function () {
```

spec/unit/pluginHandlers/handlers.spec.js
```diff
@@ -372,6 +378,12 @@ describe('android project handler', function () {
                 android['source-file'].uninstall(valid_source[9], dummyPluginInfo, dummyProject, {android_studio: true});
                 expect(removeFileSpy).toHaveBeenCalledWith(temp, path.join('app/src/main/jniLibs/x86/libnative.so'));
             });
+
+            it('Test#019j : should remove stuff by calling common.deleteJava for Android Studio projects, with target-dir that includes "appco"', function () {
```
janpio:
```suggestion
            it('Test#019j : should remove stuff by calling common.deleteJava for Android Studio projects, with target-dir that includes "app"', function () {
```

kamilbrk:
FYI the in**app**browser plugin is affected, does not load at all on 7.1.3

brodybits:
Thanks @janpio for the review. I got a call in 5 minutes, will test with more plugins and merge afterwards.

brodybits:
And thanks to @kamilbrk for pointing out the issue with `cordova-plugin-inappbrowser`.

brodybits:
## Tested in a new Cordova project

### reproduced using cordova-android@latest (7.1.3)

* `cordova platform add android@latest`
* `cordova plugin add cordova-plugin-inappbrowser`

then the plugin Java files were incorrectly installed in the following relative directory path: `platforms/android/src/org/apache/cordova/inappbrowser`

### Verified on proposed bug fix

* `cordova platform add https://github.com/brodybits/cordova-android#app-subdir-bugfix` (based on the `master` branch, which includes bleeding-edge updates for Cordova 9)
* `cordova plugin add cordova-plugin-inappbrowser`

then the plugin Java files are correctly installed in the following relative directory path: `platforms/android/app/src/main/java/org/apache/cordova/inappbrowser`

I also tested that `cordova-plugin-inappbrowser` Java files install correctly on cordova-android@7.1.2.

Merging now.

brodybits:
7.1.4 patch release is now published to resolve the issue.

PR #573, opened 2018-11-22
7.1.4 patch updates

brodybits:
* Update android-versions to `1.4.0` (seems to add support for Android Pie)
* Some other minor dependencies updates in 7.1.x
* Output current package name if package name can't be validated (#567)
* Resolve issue with plugin target-dir="app*" subdirs (#572)

If approved, I would like to merge later today and make a 7.1.4 patch release.


janpio:
Sure that you want to include the dep updates?

Besides that: let's get this out as fast as possible.

guyromb:
Can you please release it? Thanks!

brodybits:
On Thu, Nov 22, 2018, 10:38 AM Jan Piotrowski <notifications@github.com>
wrote:

> Sure that you want to include the dep updates?
>
I had wanted to get the patch update at least for android-versions, which
was done its own commit. I think this should be valid for a patch release,
any other feedback would be welcome. I can take the patch updated out if
you want.

Besides that: let's get this out as fast as possible.
>
I will make the release for review vote in 2-3 hours, going out to lunch
with family now. I'm fine if someone else wants to make it sooner.

>

Issue #575, opened 2018-11-23
Target directory generation in source-file for legacy plugin is not enough restrictive


Jule-:
This line :
https://github.com/apache/cordova-android/blob/a68f9fd752a25bb680da96414a3b956357c2dfcb/bin/templates/cordova/lib/pluginHandlers.js#L323

Should be written like this :
```js
if (obj.targetDir.startsWith('app/')) {
```

Or maybe you should treat paths as directory arrays and check the first one.

Current released version is :
https://github.com/apache/cordova-android/blob/e1befaca5a0c52ebc62cbb94870558d428d92fc4/bin/templates/cordova/lib/pluginHandlers.js#L323

Which is leading plugin with package names containing somewhere `app` to be wrongly installed in `platforms/android/src/`.
This is the case for : [cordova-plugin-local-notifications](https://github.com/katzer/cordova-plugin-local-notifications) (`de.appplant.cordova.plugin.localnotification.LocalNotification`).

And with the actual form packages like `applause.fun.example` should get trapped in that issue.

Jule-:
Oh maybe I am wrong, since directory should start with `src/applause/fun/example`... Need confirmation!

Jule-:
But I think it doesn't hurt to put the `/` no? ;)

brodybits:
My comments coming in PR #576

brodybits:
Resolved by #589

PR #576, opened 2018-11-23
Improve target-dir restriction for detecting new android project…

Jule-:
…structure used in plugin.xml. (#575)

### What does this PR do?
It just checks that `source-file` `target-dir` starts with `app/` more than `app` in order to detect the new android project structure usage in `plugin.xml`.
All explanations in #575.

### DIFF
--- a/bin/templates/cordova/lib/pluginHandlers.js
+++ b/bin/templates/cordova/lib/pluginHandlers.js
@@ -341,7 +341,7 @@ function getInstallDestination (obj) {
             } else {
                 return path.join('app', obj.targetDir, path.basename(obj.src));
             }
-        } else if (obj.targetDir.startsWith('src/main/')) {
+        } else if (targetDirArray[0] === 'src' && targetDirArray[1] === 'main') {
             return path.join('app', obj.targetDir, path.basename(obj.src));
         }
 
--- a/bin/templates/cordova/lib/pluginHandlers.js
+++ b/bin/templates/cordova/lib/pluginHandlers.js
@@ -320,7 +320,7 @@ function generateAttributeError (attribute, element, id) {
 function getInstallDestination (obj) {
     var APP_MAIN_PREFIX = 'app/src/main';
 
-    if (obj.targetDir.startsWith('app')) {
+    if (obj.targetDir.startsWith('app/')) {
         // If any source file is using the new app directory structure,
         // don't penalize it
         return path.join(obj.targetDir, path.basename(obj.src));
--- a/bin/templates/cordova/lib/pluginHandlers.js
+++ b/bin/templates/cordova/lib/pluginHandlers.js
@@ -320,26 +320,33 @@ function generateAttributeError (attribute, element, id) {
 function getInstallDestination (obj) {
     var APP_MAIN_PREFIX = 'app/src/main';
 
-    if (obj.targetDir.startsWith('app/')) {
+    var targetDirArray = obj.targetDir.split('/');
+
+    if (targetDirArray[0] === 'app') {
         // If any source file is using the new app directory structure,
         // don't penalize it
         return path.join(obj.targetDir, path.basename(obj.src));
-    } else if (obj.src.endsWith('.java')) {
-        return path.join(APP_MAIN_PREFIX, 'java', obj.targetDir.substring(4), path.basename(obj.src));
-    } else if (obj.src.endsWith('.aidl')) {
-        return path.join(APP_MAIN_PREFIX, 'aidl', obj.targetDir.substring(4), path.basename(obj.src));
-    } else if (obj.targetDir.includes('libs')) {
-        if (obj.src.endsWith('.so')) {
-            return path.join(APP_MAIN_PREFIX, 'jniLibs', obj.targetDir.substring(5), path.basename(obj.src));
-        } else {
+    } else {
+        // plugin ignores the new app directory structure (DEPRECATED)
+        if (obj.src.endsWith('.java')) {
+            return path.join(APP_MAIN_PREFIX, 'java', obj.targetDir.replace(/^src\/?/, ''),
+                path.basename(obj.src));
+        } else if (obj.src.endsWith('.aidl')) {
+            return path.join(APP_MAIN_PREFIX, 'aidl', obj.targetDir.replace(/^src\/?/, ''),
+                path.basename(obj.src));
+        } else if (targetDirArray[0] === 'libs') {
+            if (obj.src.endsWith('.so')) {
+                return path.join(APP_MAIN_PREFIX, 'jniLibs', obj.targetDir.replace(/^libs\/?/, ''),
+                    path.basename(obj.src));
+            } else {
+                return path.join('app', obj.targetDir, path.basename(obj.src));
+            }
+        } else if (obj.targetDir.startsWith('src/main/')) {
             return path.join('app', obj.targetDir, path.basename(obj.src));
         }
-    } else if (obj.targetDir.includes('src/main')) {
-        return path.join('app', obj.targetDir, path.basename(obj.src));
-    } else {
+
         // For all other source files not using the new app directory structure,
         // add 'app/src/main' to the targetDir
         return path.join(APP_MAIN_PREFIX, obj.targetDir, path.basename(obj.src));
     }
-
 }
--- a/bin/templates/cordova/lib/pluginHandlers.js
+++ b/bin/templates/cordova/lib/pluginHandlers.js
@@ -318,30 +318,34 @@ function generateAttributeError (attribute, element, id) {
 }
 
 function getInstallDestination (obj) {
-    var APP_MAIN_PREFIX = 'app/src/main';
+    const APP_MAIN_PREFIX = 'app/src/main';
+    const PATH_SEPARATOR = '/';
 
-    var targetDirArray = obj.targetDir.split('/');
+    const appReg = new RegExp(`^app\\${PATH_SEPARATOR}?`);
+    const libsReg = new RegExp(`^libs\\${PATH_SEPARATOR}?`);
+    const srcReg = new RegExp(`^src\\${PATH_SEPARATOR}?`);
+    const srcMainReg = new RegExp(`^src\\${PATH_SEPARATOR}main\\${PATH_SEPARATOR}?`);
 
-    if (targetDirArray[0] === 'app') {
+    if (appReg.test(obj.targetDir)) {
         // If any source file is using the new app directory structure,
         // don't penalize it
         return path.join(obj.targetDir, path.basename(obj.src));
     } else {
         // plugin ignores the new app directory structure (DEPRECATED)
         if (obj.src.endsWith('.java')) {
-            return path.join(APP_MAIN_PREFIX, 'java', obj.targetDir.replace(/^src\/?/, ''),
+            return path.join(APP_MAIN_PREFIX, 'java', obj.targetDir.replace(srcReg, ''),
                 path.basename(obj.src));
         } else if (obj.src.endsWith('.aidl')) {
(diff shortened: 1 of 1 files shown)


bin/templates/cordova/lib/pluginHandlers.js
```diff
@@ -320,7 +320,7 @@ function generateAttributeError (attribute, element, id) {
 function getInstallDestination (obj) {
     var APP_MAIN_PREFIX = 'app/src/main';
 
-    if (obj.targetDir.startsWith('app')) {
+    if (obj.targetDir.startsWith('app/')) {
```
brodybits: What if a plugin author uses `target-dir` that starts with `app` but not `app/`?

What if a plugin author decides to specify `target-dir="app"`?

Jule-: Indeed... Maybe I'm wrong but the point is to detect old fashioned cordova plugins in order to append `app/src/main` in front of them in order to the source files to be recognized and to be compiled by the project.

So if `target-dir` is starting with `app/` then it is a new fashioned plugin that meets new android project requirement and so new cordova-android requirements to put source files in `app/src/main/` (and I think this is what we should put in the `startWith` method!). 

If not then it is an old fashioned plugin that doesn't meets new android project requirements. So if it starts by `app` like `appsomething` this should go to `app/src/main/appsomething`.

I don't know if there is some plugin patterns that take advantage of custom `build.gradle` file in order to compile external android app structure project files but I think I have never see one yet.

And even if plugin author decides to specify `target-dir="app"` I think it is again some old fashion way and we don't want to treat it like a new fashion way. I mean `<source-file />` is for **source files** and source files in android app structure are in at least `app/src/main/`. I think it makes sense. So whenever we have source file we want to put it there. Simple and reliable.

```js
if (obj.targetDir.startsWith('app/src/main/')) {
  // plugin targets explicitly source file root => use targetDir as it is.
  return path.join(obj.targetDir, path.basename(obj.src));
} else {
  // plugin ignores underlying android app structure
  // @deprecated: BACKWARD COMPATIBILITY

  // I don't have enough history on this part so I don't know if there is specific
  // folder to put files into in order to compile correctly but I guess:
  if (obj.src.endsWith('.java')) {
    return path.join(APP_MAIN_PREFIX, 'java', obj.targetDir.replace(/^src\//, ''),
      path.basename(obj.src));
  } else if (obj.src.endsWith('.aidl')) {
    return path.join(APP_MAIN_PREFIX, 'aidl', obj.targetDir.replace(/^src\//, ''),
      path.basename(obj.src));
  } else if (obj.targetDir.startsWith('libs/')) {
    if (obj.src.endsWith('.so')) {
      // Here we should have kept the old form with .substring(5) since it is safe 
      // after .startsWith('libs/').
      // But I don't think we need optimization here more than readability.
      return path.join(APP_MAIN_PREFIX, 'jniLibs', obj.targetDir.replace(/^libs\//, ''),
        path.basename(obj.src));
    } else {
      // Don't know what sort of lib this should be handling
      return path.join('app', obj.targetDir, path.basename(obj.src));
    }
  } else if (obj.targetDir.startsWith('src/main/')) {
    // Seems to be an awkward form handling
    return path.join('app', obj.targetDir, path.basename(obj.src));
  }

  // For all other source files not using the new app directory structure,
  // add 'app/src/main' to the targetDir
  return path.join(APP_MAIN_PREFIX, obj.targetDir, path.basename(obj.src));
}
```

Jule-:
Indeed you've found same issues as me... Sorry for the delay and parallel answer...

brodybits:
@Jule- I am not sure what you mean here.

As I said in <https://github.com/apache/cordova-android/issues/578#issuecomment-441303211>: if you want to propose and discuss more changes to the `getInstallDestination` function I suggest you raise a new PR for discussion and review.

In general I would like to keep the number of "old fashioned" directory path transformations to a minimum, no more than required to avoid breaking old plugins for a while. "Old fashioned" plugins would generally have target-dir starting with `src` or `libs`. Plugin authors should have the freedom to put artifacts wherever they want into `app`. I also raised #580 to deprecate the "old fashioned" directory path transformations, to be removed in some future major release.

bin/templates/cordova/lib/pluginHandlers.js
```diff
@@ -320,26 +320,33 @@ function generateAttributeError (attribute, element, id) {
 function getInstallDestination (obj) {
     var APP_MAIN_PREFIX = 'app/src/main';
 
-    if (obj.targetDir.startsWith('app/')) {
+    var targetDirArray = obj.targetDir.split('/');
+
+    if (targetDirArray[0] === 'app') {
```
Jule-: I think this way is the safest way to handle all forms:
- `app`
- `app/`
- `app/whatever`

bin/templates/cordova/lib/pluginHandlers.js
```diff
@@ -320,26 +320,33 @@ function generateAttributeError (attribute, element, id) {
...
-    } else if (obj.src.endsWith('.aidl')) {
-        return path.join(APP_MAIN_PREFIX, 'aidl', obj.targetDir.substring(4), path.basename(obj.src));
-    } else if (obj.targetDir.includes('libs')) {
-        if (obj.src.endsWith('.so')) {
-            return path.join(APP_MAIN_PREFIX, 'jniLibs', obj.targetDir.substring(5), path.basename(obj.src));
-        } else {
+    } else {
+        // plugin ignores the new app directory structure (DEPRECATED)
+        if (obj.src.endsWith('.java')) {
+            return path.join(APP_MAIN_PREFIX, 'java', obj.targetDir.replace(/^src\/?/, ''),
```
Jule-: I have used RegExp but I should have used:
```js
targetDirArray.slice(1).join('/')
```

bin/templates/cordova/lib/pluginHandlers.js
```diff
@@ -320,26 +320,33 @@ function generateAttributeError (attribute, element, id) {
...
+            return path.join(APP_MAIN_PREFIX, 'aidl', obj.targetDir.replace(/^src\/?/, ''),
+                path.basename(obj.src));
+        } else if (targetDirArray[0] === 'libs') {
+            if (obj.src.endsWith('.so')) {
+                return path.join(APP_MAIN_PREFIX, 'jniLibs', obj.targetDir.replace(/^libs\/?/, ''),
+                    path.basename(obj.src));
+            } else {
+                return path.join('app', obj.targetDir, path.basename(obj.src));
+            }
+        } else if (obj.targetDir.startsWith('src/main/')) {
```
Jule-: Should be changed in:
```js
targetDirArray[0] === 'src' && targetDirArray[1] === 'main'
```

bin/templates/cordova/lib/pluginHandlers.js
Jule-: Actually I don't know if we want to only remove `src` directory or if we want to remove other form based on 3 chars...

bin/templates/cordova/lib/pluginHandlers.js
```diff
@@ -320,26 +320,33 @@ function generateAttributeError (attribute, element, id) {
 function getInstallDestination (obj) {
     var APP_MAIN_PREFIX = 'app/src/main';
 
-    if (obj.targetDir.startsWith('app')) {
+    var targetDirArray = obj.targetDir.split('/');
```
brodybits: I would favor it if you would define '/' as a var or constant above, like I defined `APP_MAIN_PREFIX`.

bin/templates/cordova/lib/pluginHandlers.js
brodybits: Agreed, would favor a test case that shows the actual problem you are solving here.

If a plugin target-dir would start with "app" but not one of the correct forms then it will not be a valid location anyway.

bin/templates/cordova/lib/pluginHandlers.js
```diff
@@ -320,26 +320,33 @@ function generateAttributeError (attribute, element, id) {
...
         return path.join(obj.targetDir, path.basename(obj.src));
-    } else if (obj.src.endsWith('.java')) {
-        return path.join(APP_MAIN_PREFIX, 'java', obj.targetDir.substring(4), path.basename(obj.src));
-    } else if (obj.src.endsWith('.aidl')) {
-        return path.join(APP_MAIN_PREFIX, 'aidl', obj.targetDir.substring(4), path.basename(obj.src));
-    } else if (obj.targetDir.includes('libs')) {
-        if (obj.src.endsWith('.so')) {
-            return path.join(APP_MAIN_PREFIX, 'jniLibs', obj.targetDir.substring(5), path.basename(obj.src));
-        } else {
+    } else {
```
brodybits: Why did you move the other conditions all into an else clause?

(I actually did some work to change it the other way.)

I think it does make the code a little more readable and more clear that the remaining conditions are based on the old directory structure. Makes it a little harder to review, though.

bin/templates/cordova/lib/pluginHandlers.js
```diff
@@ -320,26 +320,33 @@ function generateAttributeError (attribute, element, id) {
...
+        // plugin ignores the new app directory structure (DEPRECATED)
```
brodybits: I find the statement "ignores the new app directory structure" to be a little confusing.

I would reword it to something like: "Plugin using deprecated target directory structure"

bin/templates/cordova/lib/pluginHandlers.js
brodybits: P.S. It would be nice if we could have used the constant in the regexs below, but I think it is not possible. Maybe no need to define the constant then.

bin/templates/cordova/lib/pluginHandlers.js
brodybits: I was confused at first, coming around to favor using the RegExp you proposed.

Slice and join is additional string manipulation, should be safe though.

bin/templates/cordova/lib/pluginHandlers.js
brodybits: and I would like to see a pointer to GH-580 (#580) which states that we would like to deprecate and eventually remove support for the old target directory structure.

bin/templates/cordova/lib/pluginHandlers.js
Jule-: I get it. I think the more logical and compartmentalized the code is, the more error prone and maintainable it is. Plus, when you will want to remove deprecated code you could safely remove the entire bloc. It is your call, I don't care, just need to make it work.

bin/templates/cordova/lib/pluginHandlers.js
brodybits: The "new RegExp" should probably be its own variable as well, to make it more readable.

bin/templates/cordova/lib/pluginHandlers.js
brodybits: I think you meant "less error prone", agreed on my part.

> Plus, when you will want to remove deprecated code you could safely remove the entire bloc

Yes and this function would probably be reduced to a single `path.join` operation, we could even get rid of it when we remove the deprecated code. This function was basically introduced with another name to deal with the deprecated "old fashioned" target directory scheme.

I would personally favor the way you proposed it, with a more clear comment as I requested before.

bin/templates/cordova/lib/pluginHandlers.js
brodybits: And I would not favor backticks since they would not work on deprecated Node.js 4. (We cannot drop Node.js 4 from 7.1.x release. Even if you rebase it on master, I would rather keep it without backticks so we can cherry-pick it to 7.1.x if needed.)

janpio:
That should have been fixed with 7.1.4 which was release quickly after this bug was discovered.

PR #577, opened 2018-11-23
Other target-dir remapping fixes

brodybits:
Use startsWith instead of `includes` to check for remapping of `target-dir` that starts with `libs` or `src/main`

(Followup to PR #572)

Discovered while reviewing code related to change proposed in __PR #576.__

I think this is a very rare case that is not covered by the unit tests. I don't think we should delay 7.1.4 hotfix to get this fix in.

~~I am not sure whether or not it is worth the time to add unit tests to cover this edge case.~~

__Tests added, expected to be a rare edge case__


bin/templates/cordova/lib/pluginHandlers.js
```diff
@@ -303,13 +303,13 @@ function getInstallDestination (obj) {
         return path.join(APP_MAIN_PREFIX, 'java', obj.targetDir.substring(4), path.basename(obj.src));
     } else if (obj.src.endsWith('.aidl')) {
         return path.join(APP_MAIN_PREFIX, 'aidl', obj.targetDir.substring(4), path.basename(obj.src));
-    } else if (obj.targetDir.includes('libs')) {
+    } else if (obj.targetDir.startsWith('libs')) {
```
brodybits: From PR #576: What if there is a character other than `/` after `libs`?

bin/templates/cordova/lib/pluginHandlers.js
```diff
@@ -303,13 +303,13 @@ function getInstallDestination (obj) {
...
         if (obj.src.endsWith('.so')) {
             return path.join(APP_MAIN_PREFIX, 'jniLibs', obj.targetDir.substring(5), path.basename(obj.src));
         } else {
             return path.join('app', obj.targetDir, path.basename(obj.src));
         }
-    } else if (obj.targetDir.includes('src/main')) {
+    } else if (obj.targetDir.startsWith('src/main')) {
```
brodybits: From PR #576: What if there is a character other than `/` after `src/main `?

Jule-:
Test#006b is erroneous! It only succeed because the jar lib is called `TestLib.jar` with `Lib` inside. It seems that a specific treatment for `.jar` is needed. For test purpose we should rename it in `Test.jar`.

https://github.com/apache/cordova-android/blob/f79eda477bc5b68c758f78309312dec52ee957da/spec/unit/pluginHandlers/handlers.spec.js#L113-L117

brodybits:
> It only succeed because the jar lib is called `TestLib.jar` with `Lib` inside.

Not true since `startsWith()` is case sensitive. I just pushed an update to show that `Test#006b` succeeds with `test1.jar`.

Jule-:
@brodybits sorry I have read `src` as `targetDir`... 😩

brodybits:
Closing in favor of the updates proposed in PR #576. The updates proposed in #576 seem to be a better solution for this and other corner cases.

PR #612, opened 2019-01-09
Android Platform Release Preparation (Cordova 9)

erisu:
### Platforms affected
android

### What does this PR do?
This PR contains final preparations for the Cordova 9 release goals. 

See [Cordova 9 Release Plan](https://github.com/apache/cordova/issues/10).
 
- Remove unused promise-matchers dependency
- Bumped `cordova-common@^3.1.0`
- Bumped Dev Dependencies with ESLint Correction
  - `eslint@^5.12.0`
  - `eslint-config-semistandard@^13.0.0`
  - `eslint-config-standard@^12.0.0`
  - `eslint-plugin-import@^2.14.0`
  - `eslint-plugin-node@^8.0.1`
  - `eslint-plugin-promise@^4.0.1`
  - `eslint-plugin-standard@^4.0.0`
  - `jasmine@^3.3.1`
  - `rewire@^4.0.1`
- Updated Dependencies
  - `elementtree@^0.1.7`
  - `nopt@^4.0.1`
  - `properties-parser@^0.3.1`
- Replace istanbul dependency with `nyc@^13.1.0`
- `package.json` cleanup
  - Dropped `npm-run-all`

- `npm t`
  - eslint
  - coverage (unit testing)
  - java unit test
- [Travis CI](https://travis-ci.org/erisu/cordova-android/builds/477286097)
- [AppVeyor](https://ci.appveyor.com/project/erisu/cordova-android/builds/21489376)
- Platform Add, Build and Run
  ```
  $ npx cordova@nightly create androidTest com.foobar.androidTest
  $ npx cordova@nightly platform add github:erisu/cordova-android\#cordova9-prep
  $ npx cordova@nightly run android
  $ npx cordova@nightly build android
  ```

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,9 +33,9 @@
   "dependencies": {
     "android-versions": "^1.3.0",
     "cordova-common": "^3.1.0",
-    "elementtree": "0.1.6",
-    "nopt": "^3.0.1",
-    "properties-parser": "^0.2.3",
+    "elementtree": "^0.1.7",
+    "nopt": "^4.0.1",
+    "properties-parser": "^0.3.1",
     "q": "^1.4.1",
     "shelljs": "^0.5.3"
   },
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "android-versions": "^1.3.0",
-    "cordova-common": "^3.0.0",
+    "cordova-common": "^3.1.0",
     "elementtree": "0.1.6",
     "nopt": "^3.0.1",
     "properties-parser": "^0.2.3",
--- a/package.json
+++ b/package.json
@@ -40,18 +40,18 @@
     "shelljs": "^0.5.3"
   },
   "devDependencies": {
-    "eslint": "^3.19.0",
-    "eslint-config-semistandard": "^11.0.0",
-    "eslint-config-standard": "^10.2.1",
-    "eslint-plugin-import": "^2.3.0",
-    "eslint-plugin-node": "^5.0.0",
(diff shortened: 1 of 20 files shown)


package.json
```diff
@@ -48,7 +46,6 @@
     "eslint-plugin-promise": "^4.0.1",
     "eslint-plugin-standard": "^4.0.0",
     "jasmine": "^3.3.1",
-    "npm-run-all": "^4.1.5",
```
brodybits: Why get rid of npm-run-all?

erisu: I didn't think it was necessary to keep for only one use case.

There were two cases, but that second case was not actually needed. 

IF we still had a pattern matching requirement and didn't require any order, maybe it would have been worth keeping. 

In the end, I felt it became more of bloatware since there was only one use case left and all it was doing was shorting the line by 13 chars.

```
run-s eslint unit-tests java-unit-tests e2e-tests
```
and
```
npm run eslint && npm run cover && npm run java-unit-tests
```

The package itself is 156kb in size but with all its dependencies it was about 2.4mb. Even though some of the sub-dependencies we may already have from other packages.

I felt it might be better to just cut/downsize.

brodybits: For `dependencies` I would be strongly in favor of this kind of a reduction. But for `devDependencies` which is build tooling I do not really see the benefit of getting rid of `npm-run-all`. Even if we would only use it once I can imagine it would be nice if we could use it in more tasks someday.

The main thing is that I personally find it nicer if we don't have to do a lot of shell things in `package.json` itself.

But I don't feel super strongly about one way or the other, leaving it up to your best judgement.

Issue #623, opened 2019-01-17
Java version check does not print detected version to console


FabianTe:
Hi!

Disclaimer: I am unsure if this is the correct repository, feel free to guide me to the correct spot if I am lost 😉 

So, When using Cordova and having Java 11 and Java 8 installed on my Windows machine, running `cordova emulate android` will output this error many people know: 

```
ANDROID_HOME=C:\Users\fabit\AppData\Local\Android\Sdk
JAVA_HOME=C:\Program Files\Java\jdk1.8.0_202
Requirements check failed for JDK 1.8
```

I then thought "hmm, but my `JAVA_HOME` is pointing to the correct JDK, version 8.
Running the command with `--verbose` showed my that in fact Cordova checks not the ENV variable but some other. This happens in this function:

This can be found here: https://github.com/apache/cordova-android/blob/d9c08f12a7bceb47ee97ca8ca0e75548e78f999d/bin/templates/cordova/lib/check_reqs.js#L355-L369

As you can see, the check is happening on `values[0]`, not the `JAVA_HOME` ENV variable.

This is misleading and caused me (and propably others) to uninstall the correct Java version and still getting the error since I didn't suspect Cordova picking up Java 11.

FabianTe:
I added a `console.log` command in the linked PR.

janpio:
Ugh, why does it output that ENV var there then at all? Do you have a suggestion for better output here that would actually help users to understand what is going on? (besides just outputting the compared value like in your PR?)

FabianTe:
Well, I could rewrite it to display "Checking your Java and Android SDK version". Printing the ENV variable isn't correct.

Looks like in my case, there was more to it, i fixed my error:

As mentioned above I have Java 11 and 8 installed. What I just now found was, that I hat a `PATH` ENV entry pointing to my Java 11 installation. In some way this ended up in the called function, not my `JAVA_HOME` ENV variable. Figuring this out will take a while, but isn't related to this issue.

brodybits:
I think it is good to log both the JAVA_HOME setting and the result of `javac -version` to help the user figure out what is going wrong.

FabianTe:
@brodybits Theoretically it is. But what I have discovered is that these may not correspond to what Cordova checks. Both my ENV variables and the javac --version showed the 1.8 installation. Yet somehow Cordova used my OpenJDK 11 installation.

brodybits:
[...]

PR #624, opened 2019-01-17
Add missing log to Java version check

FabianTe:
### Platforms affected
Every platform

### What does this PR do?
Add a log that clarifies the Java validation

No testing needed, this just adds a `console.log` call.

### Checklist
- [X] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database -  #623
- [X] Added automated test coverage as appropriate for this change. - None needed

### DIFF
--- a/bin/templates/cordova/lib/check_reqs.js
+++ b/bin/templates/cordova/lib/check_reqs.js
@@ -356,10 +356,12 @@ module.exports.check_android_target = function (originalError) {
 module.exports.run = function () {
     return Q.all([this.check_java(), this.check_android()]).then(function (values) {
         console.log('Checking Java JDK and Android SDK versions');
+        console.log('ANDROID_HOME=' + process.env['ANDROID_HOME']);
 
         if (!String(values[0]).startsWith('1.8.')) {
             throw new CordovaError(
-              `Requirements check failed for JDK 8 ('1.8.*')! Detected version: ${values[0]}\nCheck your JAVA_HOME / PATH environment variables.`
+                'Requirements check failed for JDK 8 (\'1.8.*\')! Detected version: ' + values[0] + '\n' +
+                'Check your JAVA_HOME / ANDROID_HOME / PATH environment variables.'
             );
         }
 
--- a/bin/templates/cordova/lib/check_reqs.js
+++ b/bin/templates/cordova/lib/check_reqs.js
@@ -357,6 +357,7 @@ module.exports.run = function () {
     return Q.all([this.check_java(), this.check_android()]).then(function (values) {
         console.log('ANDROID_HOME=' + process.env['ANDROID_HOME']);
         console.log('JAVA_HOME=' + process.env['JAVA_HOME']);
+        console.log('Detected Java version: ' + values[0]);
 
         if (!String(values[0]).startsWith('1.8.')) {
             throw new CordovaError(`Requirements check failed for JDK 8 ('1.8.*')`);
--- a/bin/templates/cordova/lib/check_reqs.js
+++ b/bin/templates/cordova/lib/check_reqs.js
@@ -355,16 +355,16 @@ module.exports.check_android_target = function (originalError) {
 // Returns a promise.
 module.exports.run = function () {
     return Q.all([this.check_java(), this.check_android()]).then(function (values) {
-        console.log('ANDROID_HOME=' + process.env['ANDROID_HOME']);
-        console.log('JAVA_HOME=' + process.env['JAVA_HOME']);
-        console.log('Detected Java version: ' + values[0]);
+        console.log('Checking Java JDK and Android SDK versions');
 
         if (!String(values[0]).startsWith('1.8.')) {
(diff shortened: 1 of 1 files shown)


bin/templates/cordova/lib/check_reqs.js
```diff
@@ -355,16 +355,16 @@ module.exports.check_android_target = function (originalError) {
 // Returns a promise.
 module.exports.run = function () {
     return Q.all([this.check_java(), this.check_android()]).then(function (values) {
-        console.log('ANDROID_HOME=' + process.env['ANDROID_HOME']);
```
brodybits: I don't think we should get rid of this log statement. This is information that can help a developer in case something goes wrong.

bin/templates/cordova/lib/check_reqs.js
```diff
@@ -355,16 +355,16 @@ module.exports.check_android_target = function (originalError) {
...
-        console.log('JAVA_HOME=' + process.env['JAVA_HOME']);
```
brodybits: Ditto: I think this is information that can help a user in case something goes wrong.

bin/templates/cordova/lib/check_reqs.js
janpio: Correct, the `check_android` check definitely checks this env var.

janpio:
Maybe the `check_*` function should output this information when it checks the actual values?

bin/templates/cordova/lib/check_reqs.js
janpio: Here I don't really understand the code of `check_java` - is `JAVA_HOME` really used? Or is this misleading?

brodybits: JAVA_HOME is used:

<https://github.com/apache/cordova-android/blob/d9c08f12a7bceb47ee97ca8ca0e75548e78f999d/bin/templates/cordova/lib/check_reqs.js#L147>

<https://github.com/apache/cordova-android/blob/d9c08f12a7bceb47ee97ca8ca0e75548e78f999d/bin/templates/cordova/lib/check_reqs.js#L152>

and some other places.

brodybits: Another thing is that `check_reqs.js` will actually set `process.env['JAVA_HOME']` if necessary to keep track of the actual Java JDK location. I think we should fix `check_reqs.js` to track this information in an internal variable instead. I would favor working on this in a separate PR, with some more leisure.

janpio: I think this is misleading: Isn't it actually the output of `javac -version` that defines the version that is used for comparison? What exactly does the code _do_ with `JAVA_HOME` there? Is this just for the final comparison at https://github.com/apache/cordova-android/blob/d9c08f12a7bceb47ee97ca8ca0e75548e78f999d/bin/templates/cordova/lib/check_reqs.js#L209-L212? Why is `JAVA_HOME` relevant?

brodybits: JAVA_HOME can help determine where javac is run from. I would consider that to be relevant.

bin/templates/cordova/lib/check_reqs.js
```diff
@@ -355,15 +355,16 @@ module.exports.check_android_target = function (originalError) {
...
+        console.log('Checking Java JDK and Android SDK versions');
 
         if (!String(values[0]).startsWith('1.8.')) {
-            throw new CordovaError(`Requirements check failed for JDK 8 ('1.8.*')`);
+            throw new CordovaError(
+              `Requirements check failed for JDK 8 ('1.8.*')! Detected version: ${values[0]}\nCheck your JAVA_HOME / PATH environment variables.`
```
brodybits: Build failed due to eslint error. I think you need to indent this line 2 more spaces to make eslint happy, for some reason.

I would also favor splitting the message into multiple lines, like I we did here (among some other places): <https://github.com/apache/cordova-android/blob/d9c08f12a7bceb47ee97ca8ca0e75548e78f999d/bin/templates/cordova/lib/check_reqs.js#L205-L208>

bin/templates/cordova/lib/check_reqs.js
brodybits: From <https://github.com/apache/cordova-android/issues/623#issuecomment-455327547> my understanding is that the logging of JAVA_HOME may be misleading. So I would agree to take it out, at least for now. I would like to add some more detailed logging, in a non-misleading manner, someday in the future. I will mark this conversation as resolved.

bin/templates/cordova/lib/check_reqs.js
FabianTe: The eslint error is now fixed.

bin/templates/cordova/lib/check_reqs.js
FabianTe: I re-added the logging of ANDROID_HOME.

PR #625, opened 2019-01-17
Update & cleanup framework/project.properties

brodybits:
- target SDK version 28
- remove the other properties, which Android Studio reports to be unused properties
- update comments to reflect the actual situation
  (in a similar fashion to `bin/templates/project/project.properties`)

triggered by discussion in apache/cordova#69

Rationale:
- value of `target` in `framework/project.properties` is read by the Gradle in <https://github.com/apache/cordova-android/blob/f1f1ac3cbd7e75aa1dfeb6faf856a90335d2197c/framework/cordova.gradle#L30-L40> then subsequently referenced & used directly and indirectly in the following places:
  - <https://github.com/apache/cordova-android/blob/f1f1ac3cbd7e75aa1dfeb6faf856a90335d2197c/framework/cordova.gradle#L191>
  - <https://github.com/apache/cordova-android/blob/f1f1ac3cbd7e75aa1dfeb6faf856a90335d2197c/framework/build.gradle#L21>
  - <https://github.com/apache/cordova-android/blob/f1f1ac3cbd7e75aa1dfeb6faf856a90335d2197c/framework/build.gradle#L48>
- target SDK of the Cordova lib should match the target SDK of the rest of the app, if possible
- When I tried opening the generated project in Android Studio (3.3) and inspecting `platforms/android/CordovaLib/project.properties`, which seems to come from `framework/project.properties`, the editor told me that the other properties were unused. I think the unused properties should be removed for the sake of clarity.


brodybits:
Closing for now, will explain later.

PR #632, opened 2019-01-18
Ignore more Gradle build artifacts in Android project

brodybits:
I hope we can get this in before the next major release for Cordova 9 (apache/cordova#10).

P..S. These artifacts are generated by Gradle when the user does `cordova build android` or `cordova run android`.


janpio:
Why?
Why wasn't this in there before? Why should this be in now?

brodybits:
I just added the info to the description and resolved the conflict. I hope it is clear now, would be happy to answer any other questions.

brodybits:
Thanks @janpio! I didn't see that link before, just added a couple more artifacts.

janpio:
Umm, do we actually have those? Otherwise those are not really necessary to include. But well, it's merged already.

PR #633, opened 2019-01-18
[WIP] Quick fix for inconsistent SDK versions

### LINKED ISSUE #508: Inconsistent handling of min/target SDK values
cordova-android currently has a few different ways to specify a minimum SDK version and the target SDK version, and they are not equally exposed at various levels of tooling, and they can override each other in non-obvious ways (to non-native Android devs).

Traditionally, we've relied on the [`<uses-sdk>`](https://developer.android.com/guide/topics/manifest/uses-sdk-element) element in AndroidManifest.xml to specify the minimum, target, and (optionally) the maximum SDK versions. These can be customized via [`<preference>`](https://cordova.apache.org/docs/en/latest/config_ref/index.html#preference) elements in an app's config.xml (`android-minSdkVersion`, `android-targetSdkVersion`, and `android-maxSdkVersion` respectively).

We also allow specifying these values via [Gradle variables](https://cordova.apache.org/docs/en/latest/guide/platforms/android/index.html#configuring-gradle) (`cdvMinSdkVersion`, and `cdvCompileSdkVersion`). If these variables are not specified, we [default them](https://github.com/apache/cordova-android/blob/master/bin/templates/project/app/build.gradle#L101-L121).

According to [Google's documentation](https://developer.android.com/studio/publish/versioning#minsdkversion):
> If you specify API level requirements directly in your app's manifest file, the corresponding settings in the build files will override the settings in the manifest file. [...] [T]o avoid potential overwriting when the manifest is merged, you should remove these attributes from the `<uses-sdk>` element and define your API level settings in the Gradle build files instead.

In particular, I'm concerned about the possibility of someone specifying a minimum SDK version in config.xml, and that getting overridden by the default minimum SDK version of 19 via gradle.

Google seems to recommend using Gradle properties, which was also [brought up](https://github.com/apache/cordova-android/pull/507#issuecomment-424865560) in my PR about updating the Gradle build files.  
It seems to me like the correct thing to do is to take the SDK versions specified in config.xml and transform them into both `<uses-sdk>` attributes (as is done currently), but also into gradle properties. We don't currently have a good way to do that, but @erisu's PR https://github.com/apache/cordova-android/pull/465 provides a gradle.properties file that we could inject `cdvMinSdkVersion` and `cdvCompileSdkVersion` into.

Thoughts/Concerns/Feedback?

### LINKED ISSUE #629: I have a question about development with android stuido 3.2 and higher
When opening the `platform / android` project created after executing the following command with android studio 3.2 or later, the following error occurs and it can not be built.

Does this have problems with my environment? Or is it that it does not correspond to android stuido 3.2 or higher with correct behavior?

create android project
```
$ cordova create test com.example.test Test && cd $_
$ cordova platform add android@7.1.1
```

error
```
ERROR: The minSdk version should not be declared in the android manifest file. You can move the version from the manifest to the defaultConfig in the build.gradle file.
Move minSdkVersion to build files and sync project
Affected Modules: CordovaLib, app

WARNING: The targetSdk version should not be declared in the android manifest file. You can move the version from the manifest to the defaultConfig in the build.gradle file.
Move targetSdkVersion to build file and sync project
Affected Modules: app
```

environment
```
npm 6.4.1
node v10.15.0
cordova 8.1.2
```

### LINKED ISSUE #631: Compile SDK / Target SDK version not consistent with defaults in root build.gradle
If I would do `cordova platform add android@nightly` in a new Cordova project and then do `cordova build android`, the APK has target SDK version 27 while SDK 28 is specified as default in: <https://github.com/apache/cordova-android/blob/f1f1ac3cbd7e75aa1dfeb6faf856a90335d2197c/bin/templates/project/build.gradle#L45-L46>

I used the following command to check the SDK versions in the APK:

```sh
aapt list -a platforms/android/app/build/outputs/apk/debug/app-debug.apk | grep SdkVersion
```

This issue *may* be related to #508 - Inconsistent handling of min/target SDK values

I hope we can resolve this one before the next major release for Cordova 9 (apache/cordova#10).

P.S. I expanded the scope of this issue to cover the compile SDK version. This issue needs to be resolved before we can continue with AndroidX support (#565).

brodybits:
Resolves #508 - Inconsistent handling of min/target SDK values

This is a quick fix to use the minimum & target SDK version values from AndroidManifest.xml.

I have tested it in a newly generated Cordova project from the command line.

This proposal also resolves #631 - Target SDK version not consistent with defaults in root build.gradle bug

~~TODO items:~~

- [x] ~~update unit tests to succeed with the new code change~~
- [x] ~~update unit tests to cover extracting minimum & target SDK version values from AndroidManifest.xml~~

Unfortunately this does NOT resolve the issue reported in #629 (errors encountered on Android Studio 3.2 & 3.3).

I really hope this kind of a solution will be part of the upcoming major release for Cordova 9 (apache/cordova#10).

### DIFF
--- a/bin/templates/cordova/lib/builders/ProjectBuilder.js
+++ b/bin/templates/cordova/lib/builders/ProjectBuilder.js
@@ -89,7 +89,7 @@ class ProjectBuilder {
         };
     }
 
-    extractRealProjectNameFromManifest () {
+    getAndroidManifestInfo () {
         var manifestPath = path.join(this.root, 'app', 'src', 'main', 'AndroidManifest.xml');
         var manifestData = fs.readFileSync(manifestPath, 'utf8');
         var m = /<manifest[\s\S]*?package\s*=\s*"(.*?)"/i.exec(manifestData);
@@ -97,9 +97,21 @@ class ProjectBuilder {
             throw new CordovaError('Could not find package name in ' + manifestPath);
         }
 
-        var packageName = m[1];
-        var lastDotIndex = packageName.lastIndexOf('.');
-        return packageName.substring(lastDotIndex + 1);
+        const m2 = /<uses-sdk[\s\S]*?android:minSdkVersion\s*=\s*"(.*?)"/i.exec(manifestData);
+        if (!m2) {
+            throw new CordovaError('Could not find android:minSdkVersion in ' + manifestPath);
+        }
+
+        const m3 = /<uses-sdk[\s\S]*?android:targetSdkVersion\s*=\s*"(.*?)"/i.exec(manifestData);
+        if (!m3) {
+            throw new CordovaError('Could not find android:targetSdkVersion in ' + manifestPath);
+        }
+
+        return {
+            packageName: m[1],
+            minSdkVersion: parseInt(m2[1], 10),
+            targetSdkVersion: parseInt(m3[1], 10)
+        };
     }
 
     // Makes the project buildable, minus the gradle wrapper.
@@ -128,7 +140,13 @@ class ProjectBuilder {
                 checkAndCopy(subProjects[i], this.root);
             }
         }
-        var name = this.extractRealProjectNameFromManifest();
+
+        const androidManifestInfo = this.getAndroidManifestInfo();
+
+        const packageName = androidManifestInfo.packageName;
+        const lastDotIndex = packageName.lastIndexOf('.');
+        const name = packageName.substring(lastDotIndex + 1);
+
         // Remove the proj.id/name- prefix from projects: https://issues.apache.org/jira/browse/CB-9149
         var settingsGradlePaths = subProjects.map(function (p) {
             var realDir = p.replace(/[/\\]/g, ':');
@@ -144,7 +162,24 @@ class ProjectBuilder {
             '// GENERATED FILE - DO NOT EDIT\n' +
             'include ":"\n' + settingsGradlePaths.join(''));
 
-        // Update dependencies within build.gradle.
+        // Update SDK versions within root build.gradle - quick fix:
+        let rootGradle = fs.readFileSync(path.join(this.root, 'build.gradle'), 'utf8');
+        rootGradle = rootGradle.replace(
+            /(defaultMinSdkVersion)\s*=\s*\d+\s*\/\//,
+            '$1=' + androidManifestInfo.minSdkVersion + ' //');
+        // Note that we are using the configured targetSdkVersion
+        // to update both defaultCompileSdkVersion and
+        // defaultTargetSdkVersion in the root build.gradle
+        rootGradle = rootGradle.replace(
+            /(defaultTargetSdkVersion)\s*=\s*\d+\s*\/\//,
+            '$1=' + androidManifestInfo.targetSdkVersion + ' //');
+        rootGradle = rootGradle.replace(
+            /(defaultCompileSdkVersion)\s*=\s*\d+\s*\/\//,
+            '$1=' + androidManifestInfo.targetSdkVersion + ' //');
+        // and store it (in the root grade):
+        fs.writeFileSync(path.join(this.root, 'build.gradle'), rootGradle);
+
+        // Update dependencies within app/build.gradle.
         var buildGradle = fs.readFileSync(path.join(this.root, 'app', 'build.gradle'), 'utf8');
         var depsList = '';
         var root = this.root;
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -102,6 +102,9 @@ if (hasBuildExtras2) {
     apply from: '../build-extras.gradle'
 }
 
+// Needed for quick fix below:
+def customCompileSdkVersion = cdvCompileSdkVersion;
+
 // Set property defaults after extension .gradle files.
 if (ext.cdvCompileSdkVersion == null) {
     ext.cdvCompileSdkVersion = privateHelpers.getProjectTarget()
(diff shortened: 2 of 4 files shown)


bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -89,17 +89,29 @@ class ProjectBuilder {
...
         if (!m) {
             throw new CordovaError('Could not find package name in ' + manifestPath);
         }
 
-        var packageName = m[1];
-        var lastDotIndex = packageName.lastIndexOf('.');
-        return packageName.substring(lastDotIndex + 1);
+        const m2 = /<uses-sdk[\s\S]*?android:minSdkVersion\s*=\s*"(.*?)"/i.exec(manifestData);
+        if (!m2) {
+            throw new CordovaError('Could not android:minSdkVersion in ' + manifestPath);
```
fredgalvao: Was this supposed to be `Could not <find> android:min...`? The same applies to the `target` one bellow.

bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -89,17 +89,29 @@ class ProjectBuilder {
...
+        }
+
+        const m3 = /<uses-sdk[\s\S]*?android:targetSdkVersion\s*=\s*"(.*?)"/i.exec(manifestData);
+        if (!m2) {
```
fredgalvao: Was this `if` supposed to look for `!m3` instead?

bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -144,7 +162,21 @@ class ProjectBuilder {
...
+        let rootGradle = fs.readFileSync(path.join(this.root, 'build.gradle'), 'utf8');
+        rootGradle = rootGradle.replace(
+            /(defaultMinSdkVersion)\s*=\s*\d+\s*\/\//,
+            '$1=' + androidManifestInfo.minSdkVersion + ' //');
+        rootGradle = rootGradle.replace(
+            /(defaultTargetSdkVersion)\s*=\s*\d+\s*\/\//,
+            '$1=' + androidManifestInfo.targetSdkVersion + ' //');
+        rootGradle = rootGradle.replace(
+            /(defaultCompileSdkVersion)\s*=\s*\d+\s*\/\//,
+            '$1=' + androidManifestInfo.targetSdkVersion + ' //');
```
fredgalvao: Just to confirm, we're using the `targetSdkVersion` as the `defaultCompileSdkVersion`, right?

bin/templates/cordova/lib/builders/ProjectBuilder.js
brodybits: Now fixed.

bin/templates/cordova/lib/builders/ProjectBuilder.js
brodybits: Comment added

bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -89,17 +89,29 @@ class ProjectBuilder {
...
-        var lastDotIndex = packageName.lastIndexOf('.');
-        return packageName.substring(lastDotIndex + 1);
+        const m2 = /<uses-sdk[\s\S]*?android:minSdkVersion\s*=\s*"(.*?)"/i.exec(manifestData);
+        if (!m2) {
+            throw new CordovaError('Could not android:minSdkVersion in ' + manifestPath);
+        }
+
+        const m3 = /<uses-sdk[\s\S]*?android:targetSdkVersion\s*=\s*"(.*?)"/i.exec(manifestData);
+        if (!m2) {
+            throw new CordovaError('Could not android:targetSdkVersion in ' + manifestPath);
```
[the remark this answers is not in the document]
brodybits: > Missing "find" here as well (adding this comment so it does not get forgotten)

Now fixed

bin/templates/cordova/lib/builders/ProjectBuilder.js
brodybits: > Was this `if` supposed to look for `!m3` instead?

Now fixed.

brodybits:
Updated as follows:

- Updated ProjectBuilder.spec.js according to changes, with additional tests to cover reading of `android:minSdkVersion` and `android:targetSdkVersion` values from `AndroidManifest.xml`
- use `parseInt` on `android:minSdkVersion` and `android:targetSdkVersion` values
- fix use of wrong variable in an if statement, as discussed in review comments
- add missing "find" in a couple error message, as discussed in review comments

erisu:
@brodybits Can you take a look into the AppVeyor build failure?

brodybits:
Now green after an empty commit

brodybits:
Thanks @erisu for the review. I would consider this to be a stopgap or makeshift solution since it does not solve #508 and #631 the best way possible.

I would much rather find a way to resolve #508 and #631 *without* injecting the minimum & default SDK values into `AndroidManifest.xml` as a side effect. I think injecting the values into `AndroidManifest.xml` implies that the settings in `AndroidManifest.xml` are still relevant, which is not true if these settings are otherwise configured in Gradle.

An additional problem was already reported in #629: if we inject these SDK values in `AndroidManifest.xml`, recent versions of Android Studio would complain that these values should not be defined in a manifest file and offer to update manifest and Gradle files. This is behavior we would rather that our users do not see.

There was discussion in #508 about using `GradlePropertiesParser.js` to set the values in Gradle based on the settings in `config.xml`. I wonder if there would be an easy way to do this?

In case we do continue with this PR, I would definitely favor adding the `@todo` comment and changing "quick fix" to "stopgap" in all places including title, commit message, and comments.

The motivation for this stopgap solution was to solve the problem without blocking the new release for Cordova 9.

I just marked this proposal as [WIP] again, pending discussion whether to continue or find a better way to solve the problem.

brodybits:
Closing in favor of alternatives discussed in <https://github.com/apache/cordova-android/issues/508#issuecomment-461672598> and <https://github.com/apache/cordova-android/issues/508#issuecomment-461679701>

PR #635, opened 2019-01-18
Ignore Android Studio .idea files in project

brodybits:
It would be nice for this to be part of the next major release for Cordova 9 (apache/cordova#10).


Menardi:
I would say squash the commits before merging, but otherwise this is a good addition.

Issue #621, opened 2019-01-16
<input type="file" multiple> not working


StarStep:
Hi, 
I just noticed that on android is impossibile to get multiple files trought file choser.
Code used: `<input type="file" name="attachment[]" multiple="multiple"/>`

Tested on iOS and works as should. Work on Google Chrome too.
Searched on the web for any solution... but found only this:
https://stackoverflow.com/questions/5907369/file-upload-in-webview

Maybe you know how to deal with it? I tried other filechoisers but none of them supports misc files.
It would be great to be able to select files just like on the other platforms.
Thanks you very much!

artuska:
+1

StarStep:
Work for me on Android 8.1.0. 
@MatusFiala, thanks!

gabriele-sacchi:
@MatusFiala please note that the lack fo the intent type and category prevents this to work using ng-file-upload.

These 2 lines fix the problem as reported on [this new issue](https://github.com/apache/cordova-android/issues/695).

Thanks!!

shubhamtakode:
I'm facing same issue tried name="files[]" but still not working on android

marcusdiy:
Currently on android browser, using multiple attribute, allows the direct selection of multiple files. On cordova is required to hold an element to enter multiple selection mode. Would it be possibile to do the same? As its more user friendly actually...

jcesarmobile:
multiple is working fine in cordova-android 9, as marcusdiy said, you have to hold a file and then you can pick more.

And no, that behavior can't be changed, is provided by the webview picker.

PR #651, opened 2019-02-07
GH-621 android: Added multiple selection for filepicker

MatusFiala:
### Platforms affected
Android

### What does this PR do?
Allows filepicker to select multiple files.

Checked locally

### Notes on testing on devices 
Tested on following devices:
- Moto G5 (Android 7.0)
- Nexus 5X (Android 8.1)
- Samsung Galaxy Tab S4 (Android 8.1)
- Sony Xperia XZ2 Compact (Android 9.0)
- Nexus 7 (Android 6.0.1)
- Sony Xperia Z2 (Android 5.1.1)

The multiple file selection works basically, but behaves different on different devices/software.
#### Device difference:
On the Moto G5 and Samsung Tab it behaves like expected. Simply use long press (on file explorer, media-gallery, etc ) to start multiple selection.
The Nexus starts the filepicker with multiple selection enabled without using long-press.
#### Software-difference (on Samsung Tab):
- normal file-picker: Works like expected
- Google pictures-app: Starts with multiple selection enabled
- Samsung Gallery: Multiple selection does not work when using as filepicker
#### Software-difference (on Sony Xperia XZ2 Compact + Z2):
- normal file-picker: Works like expected
- Google pictures-app: Starts with multiple selection enabled
- Sony Picture Gallery: Multiple selection does not work when using as filepicker

The reason for this different behaviour is due to the different retun value of the filepicker-intent.
Normally when selecting one file, the intent writes the URL to the 'normal' data field, and when multiple files are selected to the clipData array. This works fine with the Samsung tab and Moto G5. But the intent of the Nexus always writes the first selected file to the data field, even though multiple files are selected. That's why I first check the clipData before the normal data.

Used this stackoverflow anser as inspiration: https://stackoverflow.com/questions/39390781/crosswalk-cordova-android-multiple-file-select

### Update 25.03
Tested all nadorid versions from 5 - 9. Always the same behavior:
- normal filepicker works perfect
- brand specific file picker can only choose 1 object
- Google Fotos has multiple selection always activated

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
+++ b/framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
@@ -250,13 +250,34 @@ public class SystemWebChromeClient extends WebChromeClient {
     @TargetApi(Build.VERSION_CODES.LOLLIPOP)
     @Override
     public boolean onShowFileChooser(WebView webView, final ValueCallback<Uri[]> filePathsCallback, final WebChromeClient.FileChooserParams fileChooserParams) {
+        // Check if multiple-select is specified
+        Boolean selectMultiple = false;
+        if (fileChooserParams.getMode() == WebChromeClient.FileChooserParams.MODE_OPEN_MULTIPLE) {
+            selectMultiple = true;
+        }
         Intent intent = fileChooserParams.createIntent();
+        intent.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, selectMultiple);
         try {
             parentEngine.cordova.startActivityForResult(new CordovaPlugin() {
                 @Override
                 public void onActivityResult(int requestCode, int resultCode, Intent intent) {
-                    Uri[] result = WebChromeClient.FileChooserParams.parseResult(resultCode, intent);
-                    LOG.d(LOG_TAG, "Receive file chooser URL: " + result);
+                    Uri[] result = null;
+                    if (resultCode ==  Activity.RESULT_OK && intent != null) {
+                        if (intent.getClipData() != null) {
+                            // handle multiple-selected files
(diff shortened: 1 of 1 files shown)


brodybits:
I would favor a solution where `onShowFileChooser` would be handled by one or more plugins.

dpogue:
You should not need plugins for `<input type="file" multiple>` to work. That's basic web functionality

brodybits:
Right.

So the Android WebChromeClient interface provides us a way to customize how we should handle this functionality, and we blindly start a built-in file chooser. I wonder if we should consider a way to do this in a more flexible manner?

artuska:
When this PR will be merged?

janpio:
When it is fully tested and reviewed. Can you confirm this feature works (on all Android versions= and doesn't break anything else?

StarStep:
Work great on android 8.1.0

MatusFiala:
What do you need for a merge? Do you need another tests for Androdi 5, 6 and 9 (7 and 8.1 already tested)? How can I help?

MatusFiala:
Tested on Sony Xperia XZ2 Compact (Android 9). 
Same behavior as the Samsung Tab.
Normal filepicker works, brand specific filepicker does not support multiple selection. 

I am gonna test Android 5 and 6 next week. Just need some devices.

MatusFiala:
Finally found some devices with Android 5 and 6.
Tested devices:
- Nexus 7 (Android 6.0.1)
- Sony Xperia Z2 (Android 5.1.1)

Same behaviour as every other device.
- normal filepicker works perfect
- brand specific file picker can only choose 1 object
- Google Fotos has multiple selection always activated

PR #654, opened 2019-02-08
Quick updates to top-level project.properties

brodybits:
This proposal updates the default SDK version to 28, in both AndroidManifest.xml and the resulting APK build.

There is an additional change to update the header comments according to reality, in a similar fashion to `bin/templates/project/project.properties`.


brodybits:
Thanks @jcesarmobile for the quick review!

PR #655, opened 2019-02-08
Use custom Gradle properties to read minSdkVersion value from config.xml

### LINKED ISSUE #508: Inconsistent handling of min/target SDK values

brodybits:
Based on proposal by @erisu in <https://github.com/apache/cordova-android/issues/508#issuecomment-461672598>, along with a few more changes:
- some minor fixes to GradlePropertiesParser configuration
- quick fix to handle `minSdkVersion` correctly
- updates from review
- now limited to `minSdkVersion` setting

I think some more improvements are needed to make the behavior easier to understand and maintain, hope I will get a chance to revisit after the new release is available.

FUTURE TODO:

- [ ] some cleanup & improvements by @dpogue in a branch referenced below are not included here

~~Resolves~~ ref: #508


bin/templates/cordova/lib/prepare.js
```diff
@@ -43,8 +43,16 @@ module.exports.prepare = function (cordovaProject, options) {
 
     this._config = updateConfigFilesFrom(cordovaProject.projectConfig, munger, this.locations);
 
+    // Prepare the min and target SDK version
+    const minSdkVersion = this._config.getPreference('android-minSdkVersion', 'android');
+    const targetSdkVersion = this._config.getPreference('android-targetSdkVersion', 'android');
+
+    let gradlePropertiesUserConfig = {};
+    if (minSdkVersion) gradlePropertiesUserConfig.minSdkVersion = minSdkVersion;
```
erisu: Please confirm if `gradlePropertiesUserConfig.minSdkVersion` is the correct property name to be setting or if it should be prefixed with `cdv` such as `gradlePropertiesUserConfig.cdvMinSdkVersion`.

In my example code, I used `minSdkVersion` because it was reference in [Google Android Developer Docs](https://developer.android.com/studio/publish/versioning#minsdkversion).

If you looked at @dpogue commit, he used the prefix version. https://github.com/dpogue/cordova-android/commit/d18f8d18a10a50084130300d5edbb84901b59707

bin/templates/cordova/lib/prepare.js
```diff
@@ -43,8 +43,16 @@ module.exports.prepare = function (cordovaProject, options) {
...
+    if (targetSdkVersion) gradlePropertiesUserConfig.targetSdkVersion = targetSdkVersion;
```
erisu: Again, please confirm if `gradlePropertiesUserConfig.targetSdkVersion` is the correct property name to be setting or if it should be prefixed with `cdv`. Reference from [Google Android Developer Docs](https://developer.android.com/studio/publish/versioning#minsdkversion).

In @dpogue commit, he does not use `targetSdkVersion`. He uses `cdvCompileSdkVersion`. https://github.com/dpogue/cordova-android/commit/d18f8d18a10a50084130300d5edbb84901b59707

bin/templates/cordova/lib/prepare.js
dpogue: I believe we want the `cdv`-prefixed versions, because that's what [build.gradle is looking for](https://github.com/apache/cordova-android/blob/master/bin/templates/project/app/build.gradle#L46-L90)

brodybits: Agreed on my part to use cdv prefix

bin/templates/cordova/lib/prepare.js
brodybits: > He uses `cdvCompileSdkVersion`. [dpogue@d18f8d1](https://github.com/dpogue/cordova-android/commit/d18f8d18a10a50084130300d5edbb84901b59707)

agreed on my part

brodybits:
Marked as WIP for now. I would like to resolve #657 before continuing with rework on this one.

brodybits:
I had some trouble using `cdvCompileSdkVersion` to configure the target SDK version, which is related to issue #657. My idea is to go ahead with the proposed changes for cordova-android 8 and then revisit and refine at some point in the future. If anyone has a better idea please come forward.

I would like to get PRs #656, #663, and #664 merged first, and test these changes with master before we merge this one.

bin/templates/cordova/lib/prepare.js
brodybits: now done

brodybits:
I resolved the review comments above, now limited the changes to handle `minSdkVersion`. I encountered more problems dealing with compile or target SDK value, would like to keep outside the scope of this PR.

erisu:
I am dismissing the review block as the changes were applied. I will review it again when it is no longer a WIP PR.

As for the `targetSdkVersion`, I not sure what issues you saw. With my original changes, it worked but I  didn't  use the `cdv` prefix and is maybe why I didn't see what you saw.

brodybits:
> I will review it again when it is no longer a WIP PR.

Thanks. I cannot promise when I will be able to continue with testing and fixing this one right now. I wouldn't mind if you or anyone else wants to take it over.

> As for the `targetSdkVersion`, I not sure what issues you saw.

#657 was the main one. I tried quite a few things, nothing worked right for me. The compileSdkVersion and targetSdkVersion do not always seem to be the same thing, more details in #657.

I also had some trouble using `privateHelpers.getProjectTarget` to get the targetSdkVersion value.

I gave some reasons in <https://github.com/apache/cordova-android/issues/508#issuecomment-462964865> why I would favor limiting these changes to `minSdkVersion`.

You are welcome to try it yourself, I don't have much time to test or discuss any further right now.

brodybits:
Now tested, requesting final review from @erisu.

brodybits:
Thanks @erisu for the extra testing.

So the `android-minSdkVersion` preference is now properly handled by the Gradle scripts, while the `android-targetSdkVersion` preference works because the value is copied to `AndroidManifest.xml` and **not** overridden by the Gradle scripts. (I think the second one should eventually go away, once we can reapply #664 (remove `uses-sdk` from AndroidManifest.xml), for another major release in the future.)

I did also try with the following command, which seemed to work properly:

```
cordova build android -- --gradleArg=-PcdvMinSdkVersion=21
```

Issue #658, opened 2019-02-10
Gradle default values ignored


brodybits:
From the default values in <https://github.com/apache/cordova-android/blob/a6f30b6977156df4bb2e3a488178ac05a96f0671/bin/templates/project/build.gradle#L43-L46>

the following default values seem to be ignored:

- `defaultBuildToolsVersion`
- `defaultTargetSdkVersion`
- `defaultCompileSdkVersion`

The default values were introduced in f790aeb8f677eb047ce6e82fceaeb814cfc9d4fc (part of PR #389), where `defaultTargetSdkVersion` was introduced but never used.

The code that uses `defaultBuildToolsVersion` and `defaultCompileSdkVersion` was commented out in c74192d578e647a8386ac8663d6c99521f28b110 (also part of PR #389).

I discovered this issue through investigation of code related to issue #657 & PR #655.

erisu:
Closing ticket as the code in question had been removed in Cordova-Android 10.

PR #663, opened 2019-02-10
[WIP] Cleanup build.gradle defaults not used

brodybits:
- remove `defaultTargetSdkVersion` which was never used
- comment out `defaultBuildToolsVersion` & `defaultCompileSdkVersion` for now

This is a quick-fix solution for #658.

### DIFF
--- a/bin/templates/project/build.gradle
+++ b/bin/templates/project/build.gradle
@@ -40,11 +40,13 @@ allprojects {
 
     //This replaces project.properties w.r.t. build settings
     project.ext {
-      // FUTURE TBD
+      // Not used
       // (see <https://github.com/apache/cordova-android/issues/659>):
       // defaultBuildToolsVersion="28.0.3" //String
+
       defaultMinSdkVersion=19 //Integer - Minimum requirement is Android 4.4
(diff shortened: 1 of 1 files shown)


oliversalzburg:
LGTM

brodybits:
> `FUTURE TBD` is not something that should be in the code but an issue on GitHub.

Reworded, with blank lines inserted as well. I *did* already raise issues concerning the defaults which should be used someday in the future.

> Are we sure those are not needed anywhere? Why were they here before? Did you investigate why/when they were added?

I would say yes. I *did* grep through the Gradle files to hopefully ensure that I got it right. Details about where and how the defaults were introduced are in issue #658.

janpio:
(I didn't actually want to approve this, as I have no idea what it could break. But I can't find a way to remove the approval without requesting changes - which I also don't want to do. Please someone else approve this before merging.)

brodybits:
I am closing this one for now.

As I said in <https://github.com/apache/cordova-android/pull/655#issuecomment-462584932> we may still need to deal with target SDK version.

Issue #658 (default SDK values never used) remains unsolved at this point.

PR #664, opened 2019-02-10
Remove uses-sdk from AndroidManifest.xml

brodybits:
since uses-sdk values are now superseded by Gradle files

see #629 

P.S. Changes should be merged as a squash merge.

P.P.S. Here is the context:

- the SDK values in AndroidManifest.xml are now ignored by the Gradle scripts, as discussed in #508;
- keeping the SDK values in AndroidManifest.xml leads to some error messages in Android Studio, as discussed in #629;
- a fix to use the correct SDK values was already proposed in PR #655. I would like to get this proposal reviewed and merged before finishing the rework in PR #655.


oliversalzburg:
LGTM

janpio:
This PR removes the handling of the values/attributes that should not be used any more. Do we also have the part where the replacement is set in `build.gradle` somewhere? I am missing this context here.

brodybits:
I just added the context to the description, hope it is clear enough now.

janpio:
So merging this will break the code without #655 being merged as well?

brodybits:
Merging this should not break anything that is not already broken. There exists issue #508 since Gradle does not use the information from AndroidManifest.xml. So my idea is to fix it in multiple steps:

- first step: remove the thing that is no longer wanted, as proposed here
- second step: fix the problem, as proposed in #508

brodybits:
Thanks @janpio for the quick review!

brodybits:
Something went wrong. I discovered that a result of this change is that the APK is generated with a nonsense target SDK value. This is because while the existing Gradle scripts *do* supersede the minSdkValue from AndroidManifest.xml, it *do not* supersede the targetSdkValue from AndroidManifest.xml. I will raise a new issue to describe the problem and discuss alternative solutions really soon.

My apologies.

PR #685, opened 2019-02-26
Use wwwExclude preference to filter www files.

### LINKED ISSUE #684: Would like to be able to filter files copied from www directory.
# Feature Request

## Motivation Behind Feature

Currently, Cordova will copy all files from the project www folder to the assets folder during the prepare process. However, the project www folder may contain files that should not be part of the built application and as such, they should not be copied.

## Feature Description

It would be nice if Cordova could use a list of excludes to avoid copying unneeded files. I would suggest using a preference such as wwwExcludes. This preference would contain a list of glob strings that would be used to filter the list of files copied to the assets folder during the prepare process.

The only current alternative is to use an after_prepare hook to remove the unneeded files. 

In my case, Cordova is copying VCS files (ie from .git or .hg) which can be quite large.

wifisher:
### Platforms affected
Android

### Motivation and Context

Resolves #684 

### Description

Modified the prepare process to use the wwwExcludes preference to provide a list of glob strings to use as excludes. The list is provided to FileUpdater.mergeAndUpdateDir() in its options.excludes parameter.

With this change, files and directories can be excluded from the list of files copied to the assets www directory.

### Testing

Used "cordova prepare android -d" before and after the change to prepare.js and reviewed the list of files copied to the assets/www directory.

### Checklist

- [] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))

### DIFF
--- a/bin/templates/cordova/lib/prepare.js
+++ b/bin/templates/cordova/lib/prepare.js
@@ -151,11 +151,11 @@ function updateWww (cordovaProject, destinations) {
         sourceDirs.push(path.join('merges', 'android'));
     }
 
-    var excludes = "";
-    var excludePref = cordovaProject.projectConfig.getPreference("wwwExcludes");
+    var excludes = '';
+    var excludePref = cordovaProject.projectConfig.getPreference('wwwExcludes');
     if (excludePref) {
-        excludes = excludePref.split(",");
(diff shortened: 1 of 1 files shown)


jcesarmobile:
Response on the linked issue

PR #724, opened 2019-05-01
Add Node.js 12 to CI Services

erisu:
### Platforms affected
none

### Motivation and Context
Add Node 12 too CI

### Description
- Updated CI Service Config Files

### Testing
none

### Checklist

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,10 +13,6 @@ environment:
         - nodejs_version: 10
         - nodejs_version: 12
 
-matrix:
-    allow_failures:
-        - nodejs_version: 12
-
 install:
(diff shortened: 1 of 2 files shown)


appveyor.yml
```diff
@@ -11,6 +11,11 @@ environment:
         - nodejs_version: 6
         - nodejs_version: 8
         - nodejs_version: 10
+        - nodejs_version: 12
+
+matrix:
+    allow_failures:
```
brodybits: I really do not like this setting. I would rather we fix it instead.

brodybits: Removed as it is no longer needed

PR #733, opened 2019-05-11
fix(splashscreen): nav & title bar showing in fullscreen mode

distinctdan:
### Platforms affected
Android
cordova-plugin-splashscreen

### Motivation and Context
This is a fix for an issue with the splashscreen plugin where if the app is in fullscreen mode, the nav and title bars briefly show after the splashscreen hides. https://github.com/apache/cordova-plugin-splashscreen/issues/180

### Description

The splashscreen plugin hides the main CordovaActivity initially and while the splashscreen is visible. However, the main activity doesn't call `setSystemUiVisibility` until it's focused. This causes the nav and title bars to briefly show after the splashscreen hides. This is happening because `setSystemUiVisibility` only applies while a view is visible. When the splashscreen hides, the system ui reverts back to visible before the main activity gets focused. So, to fix it, I'm calling `setSystemUiVisibility` in `onCreate` in addition to on focus. I'm new to the project, so I'm open to suggestions or adjustments on how to fix it.

### Testing

Tested on Nexus 5x and simulator.
Tested with splashscreen plugin

### Checklist

        - Current tests are failing for me without any changes.
- [X] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))


breautek:
I realise this kinda old. I'm trying to reproduce the issue this resolves, so that I can then apply this PR to see if it actually resolves the issue, but I'm failing to reproduce the issue in the first place.

Is it possible that this isn't an issue anymore?

I have an app set to `fullscreen`, and used the splashscreen plugin to provide a splashscreen. The auto hide doesn't appear to trigger any flash on the statusbar. Calling the splashscreen's `hide()/show()` methods manually in the console also doesn't trigger any flashing.

I'm testing on an Android 10/API 29 emulator, against cordova-android dev (currently 8.2.0-dev).

distinctdan:
Hmm, perhaps my language wasn't clear enough. I didn't mean to specify a literal flash, just that the status bar and nav controls would pop in while the splashscreen is visible. The correct behavior is that they should be completely hidden if the splashscreen specifies fullscreen mode. For example, here's a screenshot of my app's splashscreen running on the API 29 emulator. The app is set to fullscreen, but both the statusbar and nav controls are still visible. To get them to go away, I've had to apply both this change, and my fork of the splashscreen plugin, which still has an open PR for the corresponding splashscreen fix: https://github.com/apache/cordova-plugin-splashscreen/pull/213

Screen Shot 2019-11-03 at 6 34 35 PM

breautek:
Ahh gotcha, definitely was a misunderstanding on my part. I'll try to take another look soon

distinctdan:
I've edited the issue to try to make it clearer.

brodybits:
I just updated the milestone, hope someone can review it before we make the next major release.

peitschie:
Hi!

Just noting, I've tested this on Android 10, and it definitely improves the situation immensely for me.

Is this still being blocked by something? The code presented here makes good sense to me.

breautek (approved):

erisu (approved):
Code LGTM

Issue #722, opened 2019-04-29
Keyboard focus not restored to last element on app resume


goffioul:
# Bug Report

## Problem
When resuming a cordova app, the keyboard focus is moved instead of being restored to the last focused element.
### What is expected to happen?
When the cordova app is paused and there was a focused element in the webview content, one expect the focus to be restored to that element when the app is resumed.

### What does actually happen?
The focus is moved to the first element in the page that has a non-negative tabIndex, otherwise to the document body (ignoring the elements with negative tabIndex).

## Information
This behavior has been introduced with this commit: https://github.com/apache/cordova-android/commit/c2cafb4b45fcbb307a113828dfc9f723f8a6433b

### Command or Code
- Create a new cordova app and add android platform.
- Add the following to `div.class` element in www/index.html:
```
<div id="focusable1" tabIndex="0">I am focusable</div>
<div id="focusable2" tabIndex="0">I am also focusable</div>
<input type="text">
```
- Add the following to www/css/index.css:
```
#focusable1:focus { background: yellow; }
#focusable2:focus { background: red; }
```
- Add the following to `app.receivedEvents` method in www/js/index.js:
```
document.getElementById('focusable1').focus();
```
- Compile and install the app
- On app launch, the first focusable element should have the focus, with yellow background
- Using mouse or keyboard, focus the second element, with red background
- Using adb, launch another app, e.g. `adb shell am start com.android.settings/.Settings`
- Resume cordova app, e.g. quit settings using the BACK button

Focus will be restored to the first element, with yellow background. If using `tabIndex="-1"` instead for the focusable elements, focus will be moved to the input element.

### Version information
cordova-android@7.0.0

## Checklist

- [X] I searched for existing GitHub issues
- [X] I updated all Cordova tooling to most recent version
- [X] I included all the necessary information above

janpio:
(Great issue description! 💕 )

Looking at [the commit](https://github.com/apache/cordova-android/commit/c2cafb4b45fcbb307a113828dfc9f723f8a6433b) (from 2014!) that added this behavior, it seems to have been added to fix a specific problem. Do you have a suggestion what should be done to fix the bug you reported? Should the code that resets the focus just be removed? Or are there other proper ways to handle this? "Manually" maybe?

goffioul:
[...]

PR #750, opened 2019-06-20
Don't request focus explicitly if not needed

goffioul:
Related to issue #722.

Requesting the focus explicitly actually makes the child WebView to move the focus to the first visible focusable element on the page. This makes it impossible to simply let the WebView restore the focus to the last focused element, before the activity was paused. To prevent this problem on devices other that the Samsung Galaxy Note 3, only request the focus if necessary (it might as well be possible that the original fix is not needed anymore on newer versions of Android and/or WebView).


janpio:
(Restart CI)

Issue #513, opened 2018-10-09
BuildHelper provokes ClassNotFoundException when applicationid is not same as manifest package name


cdmahoney:
Error occurs  when trying to take picture  with camera plugin. Abbreviated stack trace:

~~~~
10-09 13:04:53.164: W/System.err(17950): java.lang.ClassNotFoundException: <application id>.BuildConfig
10-09 13:04:53.164: W/System.err(17950): 	at java.lang.Class.classForName(Native Method)
10-09 13:04:53.164: W/System.err(17950): 	at java.lang.Class.forName(Class.java:453)
10-09 13:04:53.164: W/System.err(17950): 	at java.lang.Class.forName(Class.java:378)
10-09 13:04:53.164: W/System.err(17950): 	at org.apache.cordova.BuildHelper.getBuildConfigValue(BuildHelper.java:54)
10-09 13:04:53.164: W/System.err(17950): 	at org.apache.cordova.camera.CameraLauncher.execute(CameraLauncher.java:140)
~~~~

Error can be reproduced by  defining applicationIdSuffix in build-extras.gradle, for example with
~~~~
		debug {
			applicationIdSuffix ".debug"
			versionNameSuffix ".debug"
		}
~~~~

The problem is caused by BuildHelper calling 'context.getPackageName', which returns the application id, not the package name.

Ironically, the application id is what CameraLauncher is trying to get in the first place, so I've been able to fix the problem modifying code in CameraLauncher:

~~~~
        Context context = cordova.getContext();
        this.applicationId = context.getPackageName();
        // this.applicationId = (String) BuildHelper.getBuildConfigValue(cordova.getActivity(), "APPLICATION_ID");
~~~~

But I think this should really work out of the box... (maybe add a method to CordovaInterface to return BuildConfig, which could be informed optionally from the application's CordovaActivity implementation?)

cdmahoney:
@FreddyCrugger, if you're getting the java.lang.ClassNotFoundException error try manually editing the execute method in CameraLauncher.java (in folder platform\android\app\src\main\java\org\apache\cordova\camera) to use the value returned by `context.getPackageName()` directly, as in the example I gave above. 

If that fixes the problem you might want to set up a hook to copy the patched version of CameraLauncher.java to platform\android\app\src\main\java\org\apache\cordova\camera to prevent losing the changes if the plugin is reinstalled (be careful too if you update the plugin - you may need to update the patched file too.)

FreddyCrugger:
[...]

PR #751, opened 2019-06-22
Use java package name for loading BuildConfig

kristianheljas:
### Platforms affected

Android

### Motivation and Context

Resolving `BuildConfig` using android package name (`Context.getPackageName`) will only work when java and android package names are the same.

Android package name is effectively its `applicationId` which does not have to match java package name.
A common practice is to use different android package names for different built variants/flavours, ie:
```
debug {
	applicationIdSuffix ".debug"
}
```
Now the applicationId (and android package name) might be something like `com.example.debug`, but `BuildConfig `will still resolve by `com.example.BuildConfig` instead of `com.example.debug.ClassName`

This pull request resolves issue #513.

### Description

Now using `java.lang.Class.getPackage()` instead of `android.content.Context.getPackageName()`

### Testing

Tested with and without `applicationIdSuffix` with single Android 9 device.

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/BuildHelper.java
+++ b/framework/src/org/apache/cordova/BuildHelper.java
@@ -51,7 +51,7 @@ public class BuildHelper {
     {
         try
         {
-            Class<?> clazz = Class.forName(ctx.getPackageName() + ".BuildConfig");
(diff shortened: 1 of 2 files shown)


janpio:
(Restart CI)

brodybits:
Looks OK to me. It would be great if another member can review this. And I wonder if we can cover this in the test suite?

Issue #729, opened 2019-05-08
Support new Android App Bundle package format


fortunella:
# Feature Request

## Motivation Behind Feature
Google recommends using the new Android App Bundle format for apps uploaded to Google Play. When you upload an APK a warning is displayed which encourages to use the newer format.

## Feature Description
Google provides information on the new package format on:
https://developer.android.com/platform/technology/app-bundle
It would be great if Cordova would support this new package format out of the box or with a parameter for the build command.

niren777:
I'm able to generate both 32-bit and 64-bit versions using `cordova build android --release` and `cordova build android --release --xwalk64bit` respectively

As Android App Bundle package is not supported by cordova-android yet, should I upload both `android-armv7-debug.apk` and `android-arm64-release.apk` to play store to support both 32-bit and 64-bit mobiles?

By uploading both the apk files, I'll end up creating one more project for the 64-bit version and also with different version code as mentioned by android here https://developer.android.com/google/play/publishing/multiple-apks

breautek:
> I'm able to generate both 32-bit and 64-bit versions using cordova build android --release and cordova build android --release --xwalk64bit respectively

As Android App Bundle package is not supported by cordova-android yet, should I upload both android-armv7-debug.apk and android-arm64-release.apk to play store to support both 32-bit and 64-bit mobiles?

Because crosswalk uses native libraries, yes you will have to build a 32bit and a 64bit version of the APK and upload them both to the google play store. This is what I do for my apps currently since I haven't moved away from crosswalk yet.

niren777:
@breautek Thanks for the comment. 
Should I need to create another project in Google play for 64-bit version of apk? or one project is enough to support both the versions?
Also should I need to generate both the apk files with different android version code?

breautek:
Both APKs needs to have the same package ID. So they get uploaded to the same project. When you enable multiple APKs, the build will add a digit to your versionCode to represent 32bit/64bit. Google play store will automatically select the proper APK for the device that is installing the app.

To prevent hijacking this thread, if you have anymore questions about multiple APKs you should open up  a new issue.

malvarez-troop:
Hi, 
We generated the bundle with Android Studio.

The steps are:

1) Launch Android Studio
2) Go To Import Project (Eclipse ADT, Gradle, etc).
3) Select Android platform directory in your project (/platforms/android).
4) Wait to finish the Sync
5) Go to Build > Generate Sign Bundle
6) Complete Signing info
[...]

PR #764, opened 2019-07-11
feat: Build app bundles (.aab files)

breautek:
### Platforms affected
-  Android

### Motivation and Context
This PR adds support to build android bundle (aab) files.
Fixes:  #729 

### Description

~Adds an `--bundle` flag to the `cordova build` command. If present, cordova will do the necessary steps to run the `gradlew bundleDebug` command (or `gradlew bundleRelease` if `--release` flag is present)~

Accepts `packageType` property from `build.json`, or optionally the `--packageType=apk|bundle` command line argument. The `packageType` property was chosen to keep consistency with the iOS platform. 

This PR does not change the current build commands. Building APKs is still necessary for deploying apps to test devices, or of course to upload to the Google Play store using the traditional way. Bundle files are *exclusively* for deploying to Google Play Store. 

You can however use the google provided `bundletool` program to build APKs from a bundle file so this PR still provides the ability to build bundles in both debug and release variants.

If `build.json` is present in the cordova project, just like building APKs, bundles will automatically be signed using the key and password as defined in `build.json`.

This PR does alter a previous constraint where if a signing property is present, but is missing the required properties for signing a warning would be displayed. This was altered to only show the warning if a property for signing is present is missing the required properties for signing. In other words, if `packageType` is the only property defined in the `build.json` file, then a warning about missing required properties for signing won't be emitted.

The `cordova run android` command has been modified to produce an error if `packageType` is set to `bundle`. This is because you cannot execute/deploy a bundle directly to the phone. I believe it is possible to replace all usages of the `adb` tool and instead use Google's `bundletool`, which provides tools to deploy bundle files to devices and thus support the `cordova run` command using bundles, but that would be another PR for another day.

### Testing

I have done the following:
- Added unit tests for using the `--packageType` flag
- Ran `npm test` and ensure all tests passed (Running Node 10.x & cordova@9)
-  I installed `cordova-android` platform to a test app via `cordova platform add https://github.com/breautek/cordova-android.git#app-bundle` and manually ran build commands and all combinations of `--debug`, `--release`, and `--bundle`
- I have created a test app on Google Play store to ensure that google will accept a signed aab file created by `cordova build android --bundle --release`.
- I have tested app-bundle branch against one of my work projects which involves a more complicated project structure and includes the use of native libraries (crosswalk) and observed successful builds.

### Checklist

### TODO

- [x] Additionally fallback to build.json `packageType` to decide whether to build an APK or bundle. Overridable by --packageType (Removing the --bundle flag)
- [x] Update build.json documentation
- [x] Fix the added unit test so that they are consistent (one uses `withContext`, the other doesn't)
- [x] Revert devDependencies `nyc` and `jasmine`

### DIFF
--- a/bin/templates/cordova/lib/build.js
+++ b/bin/templates/cordova/lib/build.js
@@ -114,13 +114,13 @@ function parseOpts (options, resolvedTarget, projectRoot) {
         let shouldWarn = false;
         const signingKeys = ['keystore', 'alias', 'storePassword', 'password', 'keystoreType'];
 
-        Object.keys(packageArgs).forEach((key) => {
+        for (let key in packageArgs) {
             if (!shouldWarn && signingKeys.indexOf(key) > -1) {
                 // If we enter this condition, we have a key used for signing a build,
                 // but we are missing some required signing properties
                 shouldWarn = true;
             }
-        });
+        }
 
         if (shouldWarn) {
             events.emit('warn', '\'keystore\' and \'alias\' need to be specified to generate a signed archive.');
--- a/bin/templates/cordova/lib/build.js
+++ b/bin/templates/cordova/lib/build.js
@@ -45,12 +45,14 @@ function parseOpts (options, resolvedTarget, projectRoot) {
         alias: String,
         storePassword: String,
         password: String,
-        keystoreType: String
+        keystoreType: String,
+        bundle: Boolean
     }, {}, options.argv, 0);
 
     // Android Studio Build method is the default
     var ret = {
         buildType: options.release ? 'release' : 'debug',
+        isBundle: options.bundle,
         prepEnv: options.argv.prepenv,
         arch: resolvedTarget && resolvedTarget.arch,
         extraArgs: []
@@ -148,10 +150,17 @@ module.exports.run = function (options, optResolvedTarget) {
             return;
         }
         return builder.build(opts).then(function () {
-            var apkPaths = builder.findOutputApks(opts.buildType, opts.arch);
-            events.emit('log', 'Built the following apk(s): \n\t' + apkPaths.join('\n\t'));
+            var paths;
+            if (opts.isBundle) {
+                paths = builder.findOutputBundles(opts.buildType);
+                events.emit('log', 'Built the following bundle(s): \n\t' + paths.join('\n\t'));
+            } else {
+                paths = builder.findOutputApks(opts.buildType, opts.arch);
+                events.emit('log', 'Built the following apk(s): \n\t' + paths.join('\n\t'));
+            }
+
             return {
-                apkPaths: apkPaths,
+                paths: paths,
                 buildType: opts.buildType
             };
         });
@@ -264,6 +273,7 @@ PackageInfo.prototype = {
 module.exports.help = function () {
     console.log('Usage: ' + path.relative(process.cwd(), path.join('../build')) + ' [flags] [Signed APK flags]');
     console.log('Flags:');
+    console.log('    \'--bundle\': will build a bundle (.aab) file instead of APK. Used for uploading to Google Play Store');
     console.log('    \'--debug\': will build project in debug mode (default)');
     console.log('    \'--release\': will build project for release');
     console.log('    \'--nobuild\': will skip build process (useful when using run command)');
--- a/bin/templates/cordova/lib/build.js
+++ b/bin/templates/cordova/lib/build.js
@@ -30,6 +30,7 @@ var builders = require('./builders/builders');
 var events = require('cordova-common').events;
 var spawn = require('cordova-common').superspawn.spawn;
 var CordovaError = require('cordova-common').CordovaError;
+var PackageType = require('./PackageType');
 
 module.exports.parseBuildOptions = parseOpts;
 function parseOpts (options, resolvedTarget, projectRoot) {
@@ -46,13 +47,12 @@ function parseOpts (options, resolvedTarget, projectRoot) {
         storePassword: String,
         password: String,
         keystoreType: String,
-        bundle: Boolean
+        packageType: String
     }, {}, options.argv, 0);
 
     // Android Studio Build method is the default
     var ret = {
         buildType: options.release ? 'release' : 'debug',
-        isBundle: options.bundle,
         prepEnv: options.argv.prepenv,
         arch: resolvedTarget && resolvedTarget.arch,
         extraArgs: []
@@ -70,14 +70,14 @@ function parseOpts (options, resolvedTarget, projectRoot) {
 
     if (options.argv.keystore) { packageArgs.keystore = path.relative(projectRoot, path.resolve(options.argv.keystore)); }
 
-    ['alias', 'storePassword', 'password', 'keystoreType'].forEach(function (flagName) {
+    ['alias', 'storePassword', 'password', 'keystoreType', 'packageType'].forEach(function (flagName) {
         if (options.argv[flagName]) { packageArgs[flagName] = options.argv[flagName]; }
     });
 
     var buildConfig = options.buildConfig;
 
     // If some values are not specified as command line arguments - use build config to supplement them.
-    // Command line arguemnts have precedence over build config.
+    // Command line arguments have precedence over build config.
     if (buildConfig) {
         if (!fs.existsSync(buildConfig)) {
             throw new Error('Specified build config file does not exist: ' + buildConfig);
@@ -95,9 +95,11 @@ function parseOpts (options, resolvedTarget, projectRoot) {
                 events.emit('log', 'Reading the keystore from: ' + packageArgs.keystore);
             }
 
-            ['alias', 'storePassword', 'password', 'keystoreType'].forEach(function (key) {
+            ['alias', 'storePassword', 'password', 'keystoreType', 'packageType'].forEach(function (key) {
                 packageArgs[key] = packageArgs[key] || androidInfo[key];
             });
+
+            ret.packageType = packageArgs.packageType;
         }
     }
 
@@ -110,6 +112,16 @@ function parseOpts (options, resolvedTarget, projectRoot) {
         if (Object.keys(packageArgs).length > 0) {
             events.emit('warn', '\'keystore\' and \'alias\' need to be specified to generate a signed archive.');
         }
+    }
+
+    if (ret.packageType) {
+        const VALID_PACKAGE_TYPES = [PackageType.APK, PackageType.BUNDLE];
+        if (VALID_PACKAGE_TYPES.indexOf(ret.packageType) === -1) {
+            events.emit('warn', '"' + ret.packageType + '" is an invalid packageType. Valid values are: ' + VALID_PACKAGE_TYPES.join(', ') + '\nDefaulting packageType to ' + PackageType.APK);
+            ret.packageType = PackageType.APK;
+        }
+    } else {
+        ret.packageType = PackageType.APK;
     }
 
     return ret;
@@ -151,7 +163,7 @@ module.exports.run = function (options, optResolvedTarget) {
         }
         return builder.build(opts).then(function () {
             var paths;
-            if (opts.isBundle) {
+            if (opts.packageType === PackageType.BUNDLE) {
                 paths = builder.findOutputBundles(opts.buildType);
                 events.emit('log', 'Built the following bundle(s): \n\t' + paths.join('\n\t'));
             } else {
@@ -273,7 +285,6 @@ PackageInfo.prototype = {
 module.exports.help = function () {
     console.log('Usage: ' + path.relative(process.cwd(), path.join('../build')) + ' [flags] [Signed APK flags]');
     console.log('Flags:');
-    console.log('    \'--bundle\': will build a bundle (.aab) file instead of APK. Used for uploading to Google Play Store');
     console.log('    \'--debug\': will build project in debug mode (default)');
     console.log('    \'--release\': will build project for release');
     console.log('    \'--nobuild\': will skip build process (useful when using run command)');
@@ -283,6 +294,7 @@ module.exports.help = function () {
     console.log('    \'--maxSdkVersion=#\': Override maxSdkVersion for this build. (Not Recommended)');
     console.log('    \'--targetSdkVersion=#\': Override targetSdkVersion for this build.');
     console.log('    \'--gradleArg=<gradle command line arg>\': Extra args to pass to the gradle command. Use one flag per arg. Ex. --gradleArg=-PcdvBuildMultipleApks=true');
+    console.log('    \'--packageType=<apk|bundle>\': Builds an APK or a bundle');
     console.log('');
     console.log('Signed APK flags (overwrites debug/release-signing.proprties) :');
     console.log('    \'--keystore=<path to keystore>\': Key store used to build a signed archive. (Required)');
--- a/bin/templates/cordova/lib/build.js
+++ b/bin/templates/cordova/lib/build.js
@@ -98,8 +98,6 @@ function parseOpts (options, resolvedTarget, projectRoot) {
             ['alias', 'storePassword', 'password', 'keystoreType', 'packageType'].forEach(function (key) {
                 packageArgs[key] = packageArgs[key] || androidInfo[key];
             });
-
-            ret.packageType = packageArgs.packageType;
         }
     }
 
@@ -114,11 +112,13 @@ function parseOpts (options, resolvedTarget, projectRoot) {
         }
     }
 
-    if (ret.packageType) {
+    if (packageArgs.packageType) {
         const VALID_PACKAGE_TYPES = [PackageType.APK, PackageType.BUNDLE];
-        if (VALID_PACKAGE_TYPES.indexOf(ret.packageType) === -1) {
-            events.emit('warn', '"' + ret.packageType + '" is an invalid packageType. Valid values are: ' + VALID_PACKAGE_TYPES.join(', ') + '\nDefaulting packageType to ' + PackageType.APK);
+        if (VALID_PACKAGE_TYPES.indexOf(packageArgs.packageType) === -1) {
+            events.emit('warn', '"' + packageArgs.packageType + '" is an invalid packageType. Valid values are: ' + VALID_PACKAGE_TYPES.join(', ') + '\nDefaulting packageType to ' + PackageType.APK);
             ret.packageType = PackageType.APK;
+        } else {
+            ret.packageType = packageArgs.packageType;
         }
     } else {
         ret.packageType = PackageType.APK;
--- a/bin/templates/cordova/lib/build.js
+++ b/bin/templates/cordova/lib/build.js
@@ -107,7 +107,20 @@ function parseOpts (options, resolvedTarget, projectRoot) {
     }
 
     if (!ret.packageInfo) {
-        if (Object.keys(packageArgs).length > 0) {
+        // The following loop is to decide whether to print a warning about generating a signed archive
+        // We only want to produce a warning if they are using a config property that is related to signing, but
+        // missing the required properties for signing. We don't want to produce a warning if they are simply
+        // using a build property that isn't related to singing, such as --packageType
+        let shouldWarn = false;
+        const signingKeys = ['keystore', 'alias', 'storePassword', 'password', 'keystoreType'];
+        for (let key in packageArgs) {
+            if (signingKeys.indexOf(key) > -1) {
+                shouldWarn = true;
+                break;
+            }
+        }
+
+        if (shouldWarn) {
             events.emit('warn', '\'keystore\' and \'alias\' need to be specified to generate a signed archive.');
         }
     }
--- a/bin/templates/cordova/lib/build.js
+++ b/bin/templates/cordova/lib/build.js
@@ -110,15 +110,17 @@ function parseOpts (options, resolvedTarget, projectRoot) {
         // The following loop is to decide whether to print a warning about generating a signed archive
         // We only want to produce a warning if they are using a config property that is related to signing, but
         // missing the required properties for signing. We don't want to produce a warning if they are simply
-        // using a build property that isn't related to singing, such as --packageType
+        // using a build property that isn't related to signing, such as --packageType
         let shouldWarn = false;
         const signingKeys = ['keystore', 'alias', 'storePassword', 'password', 'keystoreType'];
-        for (let key in packageArgs) {
-            if (signingKeys.indexOf(key) > -1) {
+
+        Object.keys(packageArgs).forEach((key) => {
+            if (!shouldWarn && signingKeys.indexOf(key) > -1) {
+                // If we enter this condition, we have a key used for signing a build,
+                // but we are missing some required signing properties
                 shouldWarn = true;
-                break;
             }
-        }
+        });
 
         if (shouldWarn) {
             events.emit('warn', '\'keystore\' and \'alias\' need to be specified to generate a signed archive.');
--- a/bin/templates/cordova/lib/builders/ProjectBuilder.js
+++ b/bin/templates/cordova/lib/builders/ProjectBuilder.js
@@ -384,7 +384,14 @@ function findOutputApksHelper (dir, build_type, arch) {
     return ret;
 }
 
+// This method was a copy of findOutputApksHelper and modified to look for bundles
+// While replacing shell with fs-extra, it might be a good idea to see if we can
+// generalise these findOutput methods.
 function findOutputBundlesHelper (dir, build_type) {
+    // This is an unused variable that was copied from findOutputApksHelper
+    // we are pretty sure it was meant to reset shell.config.silent back to
+    // the original value. However shell is planned to be replaced,
+    // it was left as is to avoid unintended consequences.
     const shellSilent = shell.config.silent;
     shell.config.silent = true;
 
--- a/bin/templates/cordova/lib/builders/ProjectBuilder.js
+++ b/bin/templates/cordova/lib/builders/ProjectBuilder.js
@@ -43,23 +43,25 @@ class ProjectBuilder {
     }
 
     getArgs (cmd, opts) {
-        var args;
+        let args;
         if (opts.packageType === PackageType.BUNDLE) {
+            let buildCmd;
             if (cmd === 'release') {
-                cmd = ':app:bundleRelease';
+                buildCmd = ':app:bundleRelease';
             } else if (cmd === 'debug') {
-                cmd = ':app:bundleDebug';
+                buildCmd = ':app:bundleDebug';
             }
 
-            args = [cmd, '-b', path.join(this.root, 'build.gradle')];
+            args = [buildCmd, '-b', path.join(this.root, 'build.gradle')];
         } else {
+            let buildCmd;
             if (cmd === 'release') {
-                cmd = 'cdvBuildRelease';
+                buildCmd = 'cdvBuildRelease';
             } else if (cmd === 'debug') {
-                cmd = 'cdvBuildDebug';
+                buildCmd = 'cdvBuildDebug';
             }
 
-            args = [cmd, '-b', path.join(this.root, 'build.gradle')];
+            args = [buildCmd, '-b', path.join(this.root, 'build.gradle')];
 
             if (opts.arch) {
                 args.push('-PcdvBuildArch=' + opts.arch);
@@ -383,16 +385,14 @@ function findOutputApksHelper (dir, build_type, arch) {
 }
 
 function findOutputBundlesHelper (dir, build_type) {
-    var shellSilent = shell.config.silent;
+    const shellSilent = shell.config.silent;
     shell.config.silent = true;
 
     // list directory recursively
-    var ret = shell.ls('-R', dir).map(function (file) {
-        // ls does not include base directory
-        return path.join(dir, file);
+    const ret = shell.ls('-R', dir).map(function (file) {
+        return path.join(dir, file); // ls does not include base directory
     }).filter(function (file) {
-        // find all bundles
-        return file.match(/\.aab?$/i);
+        return file.match(/\.aab?$/i); // find all bundles
     }).filter(function (candidate) {
         // Need to choose between release and debug bundle.
         if (build_type === 'debug') {
--- a/bin/templates/cordova/lib/builders/ProjectBuilder.js
+++ b/bin/templates/cordova/lib/builders/ProjectBuilder.js
@@ -37,23 +37,35 @@ const TEMPLATE =
 class ProjectBuilder {
     constructor (rootDirectory) {
         this.root = rootDirectory || path.resolve(__dirname, '../../..');
-        this.binDir = path.join(this.root, 'app', 'build', 'outputs', 'apk');
+        this.apkDir = path.join(this.root, 'app', 'build', 'outputs', 'apk');
+        this.aabDir = path.join(this.root, 'app', 'build', 'outputs', 'bundle');
     }
 
     getArgs (cmd, opts) {
-        if (cmd === 'release') {
-            cmd = 'cdvBuildRelease';
-        } else if (cmd === 'debug') {
-            cmd = 'cdvBuildDebug';
+        var args;
+        if (opts.isBundle) {
+            if (cmd === 'release') {
+                cmd = ':app:bundleRelease';
+            } else if (cmd === 'debug') {
+                cmd = ':app:bundleDebug';
+            }
+
+            args = [cmd, '-b', path.join(this.root, 'build.gradle')];
+        } else {
+            if (cmd === 'release') {
+                cmd = 'cdvBuildRelease';
+            } else if (cmd === 'debug') {
+                cmd = 'cdvBuildDebug';
+            }
+
+            args = [cmd, '-b', path.join(this.root, 'build.gradle')];
+
+            if (opts.arch) {
+                args.push('-PcdvBuildArch=' + opts.arch);
+            }
+
+            args.push.apply(args, opts.extraArgs);
         }
-
-        let args = [cmd, '-b', path.join(this.root, 'build.gradle')];
-
-        if (opts.arch) {
-            args.push('-PcdvBuildArch=' + opts.arch);
-        }
-
-        args.push.apply(args, opts.extraArgs);
 
         return args;
     }
@@ -287,7 +299,11 @@ class ProjectBuilder {
     }
 
     findOutputApks (build_type, arch) {
-        return findOutputApksHelper(this.binDir, build_type, arch).sort(apkSorter);
+        return findOutputApksHelper(this.apkDir, build_type, arch).sort(apkSorter);
+    }
+
+    findOutputBundles (build_type) {
+        return findOutputBundlesHelper(this.aabDir, build_type).sort(bundleSorter);
     }
 
     fetchBuildResults (build_type, arch) {
@@ -309,6 +325,19 @@ function apkSorter (fileA, fileB) {
         return -1;
     }
 
+    // De-prioritize unsigned builds
+    var unsignedRE = /-unsigned/;
+    if (unsignedRE.exec(fileA)) {
+        return 1;
+    } else if (unsignedRE.exec(fileB)) {
+        return -1;
+    }
+
+    var timeDiff = fs.statSync(fileB).mtime - fs.statSync(fileA).mtime;
+    return timeDiff === 0 ? fileA.length - fileB.length : timeDiff;
+}
+
+function bundleSorter (fileA, fileB) {
     // De-prioritize unsigned builds
     var unsignedRE = /-unsigned/;
     if (unsignedRE.exec(fileA)) {
@@ -361,6 +390,33 @@ function findOutputApksHelper (dir, build_type, arch) {
             return path.basename(p).indexOf('-' + arch) !== -1;
         });
     }
+
+    return ret;
+}
+
+function findOutputBundlesHelper (dir, build_type) {
+    var shellSilent = shell.config.silent;
+    shell.config.silent = true;
+
+    // list directory recursively
+    // console.log(dir, shell.ls('-R', dir));
+    var ret = shell.ls('-R', dir).map(function (file) {
+        // ls does not include base directory
+        return path.join(dir, file);
+    }).filter(function (file) {
+        // find all APKs
+        return file.match(/\.aab?$/i);
+    }).filter(function (candidate) {
+        var na = path.basename(candidate);
+        // Need to choose between release and debug .apk.
+        if (build_type === 'debug') {
+            return /debug/.exec(candidate);
+        }
+        if (build_type === 'release') {
+            return /release/.exec(candidate);
+        }
+        return true;
+    }).sort(apkSorter);
 
     return ret;
 }
--- a/bin/templates/cordova/lib/builders/ProjectBuilder.js
+++ b/bin/templates/cordova/lib/builders/ProjectBuilder.js
@@ -27,6 +27,7 @@ var spawn = require('cordova-common').superspawn.spawn;
 var events = require('cordova-common').events;
 var CordovaError = require('cordova-common').CordovaError;
 var check_reqs = require('../check_reqs');
+var PackageType = require('../PackageType');
 
 const MARKER = 'YOUR CHANGES WILL BE ERASED!';
 const SIGNING_PROPERTIES = '-signing.properties';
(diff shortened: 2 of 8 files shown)


breautek:
Travis seems to be failing to build because it is missing sdkmanager?

dpogue:
Thanks for the pull request!!

I haven't looked at this in any detail yet, but in my recent musing about how we might add support for this feature I had imagined we might reuse the `packageType` field from build.json (with Android-specific values) to determine whether to make a bundle or an APK.

janpio:
> Travis seems to be failing to build because it is missing sdkmanager?

I opened https://github.com/apache/cordova-android/pull/765 to start investigating what is going wrong in the current CI configuration and hopefully fix it.

janpio:
https://github.com/apache/cordova-android/pull/765 is ready and should be merged soon, then you can rebase your PR onto `master` again and have proper tests running again.

janpio:
Restart CI with now passing tests in `master`.

breautek:
> Thanks for the pull request!!
> 
> I haven't looked at this in any detail yet, but in my recent musing about how we might add support for this feature I had imagined we might reuse the `packageType` field from build.json (with Android-specific values) to determine whether to make a bundle or an APK.

I'm open to suggestions and will gladly modify this PR. I opened this PR with exactly that in mind actually.

I do have one question about using a setting from `build.json`. For some context, I'll describe a bit of android bundles that I understand to make sure we are on the same page.

Bundles won't be replacing APKs. Developers actively developing their app will still build APKs and deploy to their devices/emulators for testing. But one might opt to build a bundle instead of an APK to release to the app store.

So I think the general use case is *most* of the time, developers will be building APKs, just like they have always done traditionally. Then they'll build a bundle when they are ready to make a release. If the setting to decide whether to build an APK is in a json file, I think most people will end up requiring to edit the file constantly based on what they want to build. This is why I opted to add a `--bundle` flag instead, something that is easy to tack on when required, and easy to configure CI as well.

Interested to hear your thoughts on this.

janpio:
This is probably not about either/or, but about "additionally". I generate the `build.json` for builds on CI for example, and being able to switch to a bundle there would match my use case pretty well I think.

breautek:
> This is probably not about either/or, but about "additionally". I generate the `build.json` for builds on CI for example, and being able to switch to a bundle there would match my use case pretty well I think.

Ok that makes sense to me. Later tonight I can make an edit where I can grab the setting from build.json `packageType`. Overriding it if `--bundle` is present on the command line.

dpogue:
I think what we do on iOS with `packageType` is that we try to read it from build.json, but it can also be specified on the command-line as `--packageType=value`. The values we want to support are probably just going to be `apk` and `bundle`, with `apk` being the default if it's not specified.

bin/templates/cordova/lib/build.js
```diff
@@ -148,10 +150,17 @@ module.exports.run = function (options, optResolvedTarget) {
...
+                paths = builder.findOutputBundles(opts.buildType);
+                events.emit('log', 'Built the following bundle(s): \n\t' + paths.join('\n\t'));
+            } else {
+                paths = builder.findOutputApks(opts.buildType, opts.arch);
+                events.emit('log', 'Built the following apk(s): \n\t' + paths.join('\n\t'));
+            }
+
             return {
-                apkPaths: apkPaths,
+                paths: paths,
```
dpogue: Technically speaking, changing the name of the key in the returned object is probably a breaking API change, and might require changes to stuff like the `run` command or cordova-lib and other tooling.

breautek: I could revert `paths` back to `apkPaths` but the naming will be kind of misleading as it might not have APKs paths, but instead AAB paths now.

But I do agree it is a dangerous change. Perhaps I should just add a comment instead?

janpio: How could this be used externally @dpogue? Isn't this just an internal API?

dpogue: Ah, you might be right. I see now that we wrap the result in `Api.prototype.build` before returning it to external callers, so hopefully it's fine.

We'll probably end up doing a major version bump to support Android Q later in the summer, so I don't necessarily think it's a blocker to make this change even it is potentially breaking.

breautek: #729 milestone is for android 9, so I assumed this will be only included for android 9 as well.

Does this mean no changes is required for this at this time? Can I resolve conversation?

dpogue: No changes required :)

erisu:
Is the bumping of the devDependencies `jasmine` and `nyc` required for this PR to pass?
Could they be in a separate PR or made in a separate commit (for merge vs squash merge)?

breautek:
> Is the bumping of the devDependencies `jasmine` and `nyc` required for this PR to pass?
> Could they be in a separate PR or made in a separate commit (for merge vs squash merge)?

`nyc` had a high vulnerability, but no I don't believe it was required to change these versions for the tests to pass. I could revert these changes if it is more proper for them to be made in a separate PR.

brodybits:
I would really favor keeping the `jasmine` & `nyc` updates to devDependencies in a separate PR, as I think they have nothing to do with this feature. I would also really favor getting these devDependencies merged before we merge this one, for the sake of cleaning up the warning asap.

breautek:
> I would really favor keeping the `jasmine` & `nyc` updates to devDependencies in a separate PR, as I think they have nothing to do with this feature. I would also really favor getting these devDependencies merged before we merge this one, for the sake of cleaning up the warning asap.

Thanks for your explanation, I'm a little new to this. :) Didn't think of potential conflicts that could occur later. The changeset was more or less and artefact of me messing around with jasmine when I having troubles with Node12, that I forgot to revert before I made the PR. I've added a TODO item for myself to revert these devDependency changes later tonight.

bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -365,6 +382,31 @@ function findOutputApksHelper (dir, build_type, arch) {
     return ret;
 }
 
+function findOutputBundlesHelper (dir, build_type) {
+    var shellSilent = shell.config.silent;
+    shell.config.silent = true;
+
+    // list directory recursively
+    var ret = shell.ls('-R', dir).map(function (file) {
```
erisu: I just wanted to make a note that we been trying to drop `shelljs` as a dependency across repos when possible.

https://issues.apache.org/jira/browse/CB-14140

breautek: Good to know.. I was mimicking what was already in the codebase for this. Is there a replacement for `shelljs`?

brodybits: > https://issues.apache.org/jira/browse/CB-14140

You may want to check out apache/cordova-common#21 & apache/cordova-lib#613 where we did the replacement on a couple major modules.

FYI We did encounter apache/cordova#121 on Windows, which was eventually solved by a workaround on fs-extra@8.

And for spawning external processes, we migrated from Node.js “child_process” to `cross-spawn` in apache/cordova-common#50, started discussing further work in apache/cordova#16.

I am thinking we should stick with `shelljs` for a minor release, to resolve the issue, and consider migrating to `fs-extra` in the next major release, to avoid an accidentally breaking change.

> Dropping `shelljs` I think would make more sense to be done as a separate PR?

That would be my vote.

erisu: > Dropping shelljs I think would make more sense to be done as a separate PR?

and

> I am thinking we should stick with shelljs for a minor release, to resolve the issue, and consider migrating to fs-extra in the next major release, to avoid an accidentally breaking change.

Dropping the usage of `shelljs` that already exists would be in a separate PR from this ticket and would be major change as @brodybits pointed out, except what I am pointing at is brand new code that is introduced in this PR that could be avoided from being added.

It would not be a breaking change.

Anyways, you can keep it in, if you prefer, and we remove it later in the next major. The next major will deprecates node 6 and 8 which then we could a cleaner recursive directory scan.

Example: https://stackoverflow.com/a/45130990/1644129

breautek: I think it would be best to wait so that the `shelljs` refactor can be done all at once. Otherwise I'd have to include new dependencies and some of the code base will be using the new replacement while other parts will be still using `shelljs`.

brodybits:
What is the status of this proposal? Is anything blocking it?

janpio:
From my viewpoint this looks good!

But as it is a significant addition, I think @breautek should write a short email to the dev list for lazy consensus on how this functionality is added. We also need a brother PR for the Android docs that can be merged when the next Android release is done (minor release should be fine as it just adds functionality).

breautek:
Sorry, currently the the only thing missing is documentation, which I've been too busy this weekend to write it. But i think that's done in `cordova-docs` repo anyway.

I'd also like to do some more testing since I've removed the `--bundle` flag and replaced it with using the `build.json` stuff. I haven't thoroughly tested it myself outside of using `npm test` and quick manual checks.

brodybits:
Thanks @breautek for your work on this. I tried to help by updating `nyc` and getting AppVeyor & Travis CI with `npm test` working on Node.js 12. Please let us know if there is anything else we can do to help (no pressure intended).

I really think that this is not the easiest codebase for new contributors to work with. Thanks again for your help so far.

breautek:
> Thanks @breautek for your work on this. I tried to help by updating `nyc` and getting AppVeyor & Travis CI with `npm test` working on Node.js 12. Please let us know if there is anything else we can do to help (no pressure intended).
> 
> I really think that this is not the easiest codebase for new contributors to work with. Thanks again for your help so far.

Thanks, I should have time to work on the last bits tonight. I think I'll probably have questions regarding how to update the documentation in the `cordova-docs` repo, but I'll ask more specific questions when/if I run into problems.

janpio:
Pointers re docs:
https://cordova.apache.org/docs/en/latest/guide/platforms/android/index.html#signing-an-app
https://github.com/apache/cordova-docs/blob/master/www/docs/en/dev/guide/platforms/android/index.md
Probably compare to how iOS documents its parameters.

breautek:
Currently, if you don't have a `build.json` file, or otherwise don't have the `keystore` and `alias` parameters filled but have `--packageType` argument present, the following warning is displayed.

`'keystore' and 'alias' need to be specified to generate a signed archive.`

It's just a warning, but it feels a little weird how the presence of `--packageType` triggers the warning as the `--packageType` flag doesn't really have to do with signing.

These lines are the trigger of the warning
https://github.com/apache/cordova-android/blob/a64d459c8e339ec1b480672c028eef01027a1c02/bin/templates/cordova/lib/build.js#L102-L111

To re-iterate this occurs when `keystore` and `alias` is both undefined, but `--packageType` is displayed. Because `packageType is a `build.json` key, it is inside `packageArgs, thus `Object.keys(packageArgs)` returns 1, which allows it to enter line 109.

I feel like this should be changed so perhaps the warning only gets emitted if one of the signing arguments is present (`keystore`/`alias`/`password`/`storePassword`), but lacks `keystore` or `alias` arguments. Thoughts?

janpio:
> I feel like this should be changed so perhaps the warning only gets emitted if one of the signing arguments is present (`keystore`/`alias`/`password`/`storePassword`), but lacks `keystore` or `alias` arguments. Thoughts?

+1

(Although might indicate that using `--packageType` or `build.json` for the bundle option might not follow the original thoughts behind that - but I assume you didn't encounter any other problems and it feels right? Then that should be fine.)

breautek:
Alright, assuming travis/appveyor comes back successful, this PR is ready for more criticism.

bin/templates/cordova/lib/build.js
```diff
@@ -105,11 +107,36 @@ function parseOpts (options, resolvedTarget, projectRoot) {
...
 
     if (!ret.packageInfo) {
-        if (Object.keys(packageArgs).length > 0) {
+        // The following loop is to decide whether to print a warning about generating a signed archive
+        // We only want to produce a warning if they are using a config property that is related to signing, but
+        // missing the required properties for signing. We don't want to produce a warning if they are simply
+        // using a build property that isn't related to singing, such as --packageType
+        let shouldWarn = false;
+        const signingKeys = ['keystore', 'alias', 'storePassword', 'password', 'keystoreType'];
+        for (let key in packageArgs) {
```
brodybits: Coding nit: I think this for loop could be replaced by an array `reduce()` call. (An array `forEach()` call would also be a little cleaner than using a for loop.) I would be happy to make a quick proposal if you like.

breautek: Yah probably. I'm a little bit old school with my javascript, I'm not used to using these new `Array` methods. I can update it to use `reduce` instead of a for loop.

breautek: Actually I think `forEach` makes a bit more sense here, because `reduce` is more of an aggregation method and the `accumulator` value begins as the first item in the array, which I think is error prone in this case.

brodybits: I am also not so used to these new Array methods, actually did a little research before writing the coding nit. Next idea: what about using `concat()` to start with a bogus element, such as this (not tested, coded with help from MDN & other sites):

```js
const shouldWarn = [‘bogus’].concat(packageArgs)
    .reduce((shouldWarn, key) => (
        shouldWarn || signingKeys.indexOf(key) !== -1)
    ), false);
```

(I would probably add a quick comment to explain the less obvious aspects of this code.)

Or am I getting too fancy here?

breautek: I personally think that's even more confusing. What I have currently is...

```
let shouldWarn = false;
const signingKeys = ['keystore', 'alias', 'storePassword', 'password', 'keystoreType'];

Object.keys(packageArgs).forEach((key) => {
    if (!shouldWarn && signingKeys.indexOf(key) > -1) {
        // If we enter this condition, we have a key used for signing a build,
        // but we are missing some required signing properties
        shouldWarn = true;
    }
});
```

However, I did miss that `reduce` have an initial parameter.

Keep in mind... that `packageArgs` is not an array, it is an object... which is why I opted for `for .. in` originally.

bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -365,6 +382,31 @@ function findOutputApksHelper (dir, build_type, arch) {
...
+        // ls does not include base directory
+        return path.join(dir, file);
+    }).filter(function (file) {
+        // find all bundles
+        return file.match(/\.aab?$/i);
+    }).filter(function (candidate) {
```
brodybits: I might rewrite this part as one big expression with a big ternary:
```js
    …).filter((candidate) => (
        // May need to choose between release and debug bundle:
        (build_type === 'debug')
            ? /debug/.exec(candidate)
            : (build_type === 'release')
                ? /release/.exec(candidate);
                : true;
    ));
```

or maybe use a switch statement for this?

I am writing ternaries like they do in Prettier, seems to be a new fad that I like. Or maybe it is not right here?

breautek:
> You can't "run" a bundle, but doing `cordova run android -- --packageType=bundle` will still build a bundle, but only install the latest apk built, or it will error if you've never built an apk.
> 
> I think `--packageType` should be forced to `apk` if using `cordova run`.

Not sure what's the best way to handle this. I believe that `cordova-lib` repo is what fires the build process, before `cordova-android` `run` script executes when using `cordova run android` and it doesn't look like `cordova-lib` offers a chance for the platform to manipulate the build options before the `build` method is invoked.

https://github.com/apache/cordova-lib/blob/dfbab745d918de1cb24b2b0fe6bd1652ac2b227b/src/cordova/run.js#L46-L47

I don't see any other way where I can force `--packageType` to be `apk` when using `cordova run` so I think what I originally want will quickly become out of scope in this PR... 

The other alternative is to make `cordova run` support bundles... which will involve incorporating the new  [bundletool](https://developer.android.com/studio/command-line/bundletool) which will needed to be downloaded and installed on the users machine. Note that using `bundletool` is basically a replacement of using `adb` as it has commands and flags to install APKs to connected devices, so I also see this quickly becoming out of scope of this PR.

Will it be okay if Cordova just exits with an useful message stating that `--packageType` must be `apk` when using `cordova run android`? If so, how is this typically done in cordova? I've tried using 

`events.emit('error', 'msg...');` 

which produced... 

`Unhandled error. ('Package type "bundle" is not supported during cordova run. use Package Type "apk" instead')`

... seemingly because the the error isn't caught/handled upstream.

bin/templates/cordova/lib/build.js
breautek: I guess I shall keep the original for loop in this case then since we are handling an object and not an array?

bin/templates/cordova/lib/build.js
```diff
@@ -93,7 +95,7 @@ function parseOpts (options, resolvedTarget, projectRoot) {
                 events.emit('log', 'Reading the keystore from: ' + packageArgs.keystore);
             }
 
-            ['alias', 'storePassword', 'password', 'keystoreType'].forEach(function (key) {
+            ['alias', 'storePassword', 'password', 'keystoreType', 'packageType'].forEach(function (key) {
                 packageArgs[key] = packageArgs[key] || androidInfo[key];
```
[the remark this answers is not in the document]
breautek: `||=` is invalid syntax. Both `||` and `=` operator needs expressions on both sides of the operator.

bin/templates/cordova/lib/build.js
brodybits: I would favor the idea you had to do `Object.keys(packageArgs).forEach(…)`, as I still think it looks nicer than the for loop. What I meant is that I should have accounted for using `Object.keys(…)` when I made my original comment on this code.

IMHO I would still favor using `reduce()`, together with `Object.keys()`, but not very strongly at this point. I think the way you did `Object.keys(packageArgs).forEach(…)` is the perfect compromise.

I also have no objection in case you decide to keep the original for loop, as I think this is primarily about coding style which we can always change later.

bin/templates/cordova/lib/build.js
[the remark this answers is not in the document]
brodybits: My bad

bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -37,23 +38,35 @@ const TEMPLATE =
...
 
     getArgs (cmd, opts) {
-        if (cmd === 'release') {
-            cmd = 'cdvBuildRelease';
-        } else if (cmd === 'debug') {
-            cmd = 'cdvBuildDebug';
-        }
+        var args;
+        if (opts.packageType === PackageType.BUNDLE) {
+            if (cmd === 'release') {
```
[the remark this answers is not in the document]
breautek: It would be quite a nested ternary statement if we merge both the `packageType` comparison and the `cmd` comparison together. There are 4 different outcomes for `buildCmd`.

Therefore the ternary statement would look more like:
```javascript
const buildCmd =
    (opts.packageType === PackageType.BUNDLE)
        ? (cmd === 'release')
            ? ':app:bundleRelease'
            : ':app:bundleDebug'
         : (cmd === 'release')
            ? 'cdvBuildRelease'
            : 'cdvBuildDebug';
```
I don't have a problem making that change but personally I find that more confusing to look at.

bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -365,6 +382,31 @@ function findOutputApksHelper (dir, build_type, arch) {
     return ret;
 }
 
+function findOutputBundlesHelper (dir, build_type) {
+    var shellSilent = shell.config.silent;
```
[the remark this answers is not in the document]
breautek: Hmm, majority of this method was copied from `findOutputApksHelper`, but they don't reset `shell.config.silent` either. Instead they do `shellSilent = shellSilent;` which I think is also a bug... It definitely looks like it is meant to be reset before returning.

bin/templates/cordova/lib/builders/ProjectBuilder.js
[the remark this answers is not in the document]
brodybits: > Therefore the ternary statement would look more like:
> […]

Agreed on my part.

Leaving it up to you. Maybe better to leave some of these ideas for a followup PR, as these changes seem to have good test coverage.

bin/templates/cordova/lib/builders/ProjectBuilder.js
breautek: I don't mind writing nested ternaries if that is the Cordova standard. If you're asking my personal opinion, I typically stay away from ternaries if you have more than 2 outcomes because I find explicit blocks easier to read.

brodybits: > I typically stay away from ternaries if you have more than 2 outcomes
> […]

Let’s hold off on the extended ternaries, then.

About the Cordova code base, a lot of it seems to have been originally written for less modern JavaScript over a half-decade ago. People have been modernizing the code base in several different ways, not sure how consistent it really is at this point.

I think it would be ideal for us to discuss some general coding style updates we should do in <https://github.com/apache/cordova/issues>, like we did for some other recent improvements, and work on applying these changes through the different components. All of these ideas take time, and I think none of us have any truly extra time these days.

bin/templates/cordova/lib/build.js
janpio: I think the current code is perfectly fine and simple to understand.

bin/templates/cordova/lib/builders/ProjectBuilder.js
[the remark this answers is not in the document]
janpio: if/else is perfectly readable here IMO. a bit verbose, but simple and effective.

janpio:
> > I think `--packageType` should be forced to `apk` if using `cordova run`.
> 
> Not sure what's the best way to handle this.

Can we somehow ignore `--packageType` for `cordova run` and output a warning message to the user that `run` will ignore that parameter and build an APK anyway maybe? I think that would be the most intuitive for a user - or the "error" thing you already tried.

brodybits:
> > You can't "run" a bundle, but doing `cordova run android -- --packageType=bundle` will still build a bundle, but only install the latest apk built, or it will error if you've never built an apk.
> > I think `--packageType` should be forced to `apk` if using `cordova run`.
> 
> Not sure what's the best way to handle this.
> […]

I would favor a dumb but foolproof agile approach where `cordova run android -- --packageType=bundle` would fail with an ugly error message. I think it should be easy to do this in `run.js`.

At this point I think it would be an OK rough edge for Cordova to first attempt to build the bundle then fail with an ugly error message in `run.js`. My idea is get this into the hands of the user community now, then look into improving the whole thing in a new major release.

breautek:
> I would favor a dumb but foolproof agile approach where `cordova run android -- --packageType=bundle` would fail with an ugly error message. I think it should be easy to do this in `run.js`.

Okay, I won't touch `run.js` for the time being then. The setback here is that `run.js` is only executed __after__ it decides to build. `cordova-lib` appears to control this flow.

This means that if you decide to use `cordova run android -- -- packageType=bundle` __or__ you have `
`"packageType": "bundle" inside `build.json`, then it will build a bundle, but try to deploy an APK. The APK that will be deployed will be the APK that was last build.

bin/templates/cordova/lib/builders/ProjectBuilder.js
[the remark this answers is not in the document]
breautek: Resolving this conversation. If #786 dictates style changes, then another PR can be submitted.

bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -37,23 +38,35 @@ const TEMPLATE =
...
+        this.aabDir = path.join(this.root, 'app', 'build', 'outputs', 'bundle');
     }
 
     getArgs (cmd, opts) {
-        if (cmd === 'release') {
-            cmd = 'cdvBuildRelease';
-        } else if (cmd === 'debug') {
-            cmd = 'cdvBuildDebug';
-        }
+        var args;
```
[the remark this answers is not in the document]
breautek: I have a pending commit that does this change.

bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -37,23 +38,35 @@ const TEMPLATE =
...
+        if (opts.packageType === PackageType.BUNDLE) {
+            if (cmd === 'release') {
+                cmd = ':app:bundleRelease';
+            } else if (cmd === 'debug') {
+                cmd = ':app:bundleDebug';
+            }
+
+            args = [cmd, '-b', path.join(this.root, 'build.gradle')];
+        } else {
+            if (cmd === 'release') {
```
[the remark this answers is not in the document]
breautek: I have a pending commit to avoid overwriting the function argument `cmd`, but as per discussion at #786 I did not use ternary style condition statements.

brodybits:
I find it interesting that your recent commit be85316 both reverts a code style change and removes some changes you had made to `run.js`. The changes you had made to `run.js` in ca2fa615060c1f95414435ac74da68dda135ed77 looks to me like it would have been a good thing to keep in this PR. While tried to say it was not absolutely necessary, I do think it would be nice to have if it works correctly.

bin/templates/cordova/lib/builders/ProjectBuilder.js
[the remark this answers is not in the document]
breautek: I don't want to make changes without fully understanding the consequences. `findOutputBundlesHelper` mimics the existing method `findOutputApksHelper`.

Because `shell` is going to be removed soon and replaced with `fs-extra` I won't be making any changes relating to shell. I will however remove the usages of `var` and replace them with `const` or `let` as appropriate.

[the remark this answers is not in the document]
brodybits: Thanks @breautek.

> […]
> `findOutputBundlesHelper` mimics the existing method `findOutputApksHelper`.

I wonder if we (you) should add this information to a comment?

> Because `shell` is going to be removed soon and replaced with `fs-extra`
> […]

Whoever replaces `shelljs` with `fs-extra` may want to factor out some duplicated code at the same time, if possible.

breautek:
> The changes you had made to run.js in ca2fa61 looks to me like it would have been a good thing to keep in this PR. While tried to say it was not absolutely necessary, I do think it would be nice to have if it works correctly.

Sorry must have misunderstood. The `events.emits('error')` did work give some meaningful feedback. I can re-add that commit.

bin/templates/cordova/lib/builders/ProjectBuilder.js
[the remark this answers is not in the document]
breautek: As said in another critic, because we will be replacing `shell` with `fs-extra` in the near future, I won't be making any changes to the usages of `shell`, but I will be replacing `var` with `const`.

bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -365,6 +382,31 @@ function findOutputApksHelper (dir, build_type, arch) {
...
 
+function findOutputBundlesHelper (dir, build_type) {
+    var shellSilent = shell.config.silent;
+    shell.config.silent = true;
+
+    // list directory recursively
+    var ret = shell.ls('-R', dir).map(function (file) {
+        // ls does not include base directory
+        return path.join(dir, file);
+    }).filter(function (file) {
```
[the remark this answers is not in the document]
breautek: The suggested change is in a pending commit.

bin/templates/cordova/lib/build.js
breautek: Keeping the original for the time being, pending discussion at #786

breautek:
I believe this PR is pretty sound now, for the first iteration of android bundle support anyway.

bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -383,16 +385,14 @@ function findOutputApksHelper (dir, build_type, arch) {
 }
 
 function findOutputBundlesHelper (dir, build_type) {
-    var shellSilent = shell.config.silent;
+    const shellSilent = shell.config.silent;
```
brodybits: Nitpicking again:

While I now completely understand from the discussion why we do have the unused `const shellSilent` declaration here, …

… I would now really favor having a comment in the code that explains *why* we have the unused `const` declaration here.

And a warning that yet another nitpick is coming up.

bin/templates/cordova/lib/builders/ProjectBuilder.js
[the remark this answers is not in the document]
breautek: I could write a comment stating that with the `fs-extra` refactor it may be a good idea to try to generalise the "findOutput" methods..

brodybits:
I think my final nit is that it would be good for a test to verify that the error is produced if `run` is done with packageType=bundle. I think this test should be in `spec/unit/run.spec.js`. I would be happy to make this test if you like.

Otherwise I think this change is good to go.

breautek:
>  I would be happy to make this test if you like.

Sure thing! Unit testing is actually still a learning experience for me, and I'm not sure how to actually determine if the log has been printed to console. But I can learn from what you do.

bin/templates/cordova/lib/builders/ProjectBuilder.js
breautek: I can add a comment along with the other nit.

brodybits: All of the comments here are exactly what I was looking for, yes.

brodybits:
I just pushed some quick changes to your branch:

- pull from the master branch, which gives us the verbose test reporter among other things
- added the test case, and added a throw statement that I think is needed for `run` to fail when it is called with `--packageType=bundle` (it may be better for the code to reject the promise instead of throwing an error)

The test case is known to fail since I put bogus string values into both the throw statement and the test case. @breautek, I am leaving it up to you to fix it as you feel is most appropriate here.

I am also not 100% confident that the `run` code will do the right thing if the `packageType=bundle` setting is configured in the `build.config` file. I would appreciate it if @breautek or any other experts would double-check this.

I will probably finish my day pretty soon, would be happy to take another look when I start my new day.

janpio:
You both lost me along the way, but that's not a bad thing - progress is made. Pleaes ping me @breautek when you updated the PR description to reflect the full state of this PR (describes all the changes and behavior) and I will look over the code again. 🚀

breautek:
FYI, I'll be pretty busy for the rest of the week/weekend. So I'll probably only come back to this Monday night.

breautek:
@janpio I have updated the original PR description to include the most up to date changes and I have no more pending changes to be made against this branch, unless of course further feedback dictates more changes.

janpio:
There is a conflicted file, then this should be fine for another review.

breautek:
> There is a conflicted file, then this should be fine for another review.

My bad, I thought that conflict had something to do with the PR hold. I resolved the conflict.

breautek:
Pinging the reviewers as requested from the mailing list discussion

@dpogue
@erisu 
@brodybits 
@janpio

brodybits:
Merging now

breautek:
Virtual drinks on me :beers:

brodybits:
> 🍻

should be shared I think, thanks again for the contribution

beevelop:
You guys definitely rock 🍻
Thank you for that feature ✌️

tomriddle1234:
Hi, but can someone tell me how to use, Google play just enhanced 64bit check, cordova is not working at all for Android, all my apk get rejected.

I tried, 

```
cordova build android --release --packageType=bundle

cordova build android --bundle
```

still getting apk instead of aab.

cordova-android installed with 

`ionic cordova platform add android@https://github.com/apache/cordova-android.git`

janpio:
Step 1: Open a new issue instead of commenting on an unrelated, merged and PR.

steebchen:
Thanks for this change, it works great! Would it maybe be possible to create a pre-release for cordova 9, or some other kind of release for this change on npm? It would be a bit easier to use compared to setting up this repo/master branch as a custom android platform.

janpio:
`@nightly` should include this (and all other merged changes of course).

steebchen:
Thanks! However, `npm i cordova-android@nightly` (or yarn) installs `8.1.0-nightly.2019.6.28.4cf3dcfa`, which seems to be a commit 4cf3dcfaae6dc82ddb1ccf439d209cbcc2f474a0 from April 13. What can I do?

janpio:
Ugh, that's true - you should create an issue that cordova-android currently doesn't build nightlies :/

Then your best bet is to install from `master` (which Cordova CLI supports) or fork the repo and publish under your own npm package if you really don't want to use via Git.

steebchen:
Thanks for the info. Created #810

ath0mas:
@tomriddle1234 it should work using `--` before your param
```
cordova build android --release -- --packageType=bundle
```
like explained in [doc about flags](https://cordova.apache.org/docs/en/latest/guide/platforms/android/index.html#signing-an-app) :
> Note: You should use double -- to indicate that these are platform-specific arguments

breautek:
Just leaving a note here for Crosswalk users (like myself, since I haven't migrated away from it yet...)

If you are using the crosswalk webview plugin, building bundles will not give you a bundle that contains the 64bit crosswalk library. This is because the crosswalk plugin downloads a single prebuilt library and it will only download 32bit architecture; unless the `--xwalk64bit` flag is passed in which case, crosswalk will download a 64bit architecture library. The problem is crosswalk plugin in its current state will only ever include 1 architecture.

As a result, using bundles will not work properly with the crosswalk webview. Even if you use the `--xwalk64bit` flag you may pass the Google Play 64bit requirement, but you'll not include the 32bit architecture for older 32bit devices.

This is a problem with the crosswalk webview plugin itself, it will need to be updated so it packages all prebuilt libraries for each architecture when you build a bundle. Note that Intel has stopped maintaining the crosswalk project a couple years ago so you won't see a fix on the official crosswalk repository.

sriharsha0023:
Does this work with cordova platform add android@8.0.0 ?
and this commands looks Correct?
cordova build android --debug  -- --packageType=bundle
and
cordova build android --release  -- --packageType=bundle

breautek:
> Does this work with cordova platform add android@8.0.0 ?

No, this feature was only merged in and released under `8.1.0`

> and this commands looks Correct?
cordova build android --debug -- --packageType=bundle
and
cordova build android --release -- --packageType=bundle

Those two commands looks right.

dfiggtns:
I'm trying to convert to use app bundles, but when I use:

ionic cordova build android --release -- --packageType=bundle

cordova still creates an apk in build/outputs/apk.

I'm using cordova@9.0.0 and cordova-android@8.1.0

How can I get cordova to create an aab in build/outputs/bundle?

breautek:
@dfiggtns 

See https://github.com/ionic-team/ionic-cli/issues/4187#issuecomment-546990372 for how to pass `packageType` to cordova when using ionic.

dfiggtns:
@breautek, yes I saw that post in my trawl through the forums. In fact, that's the one I got the format of the command from that I used above.

Unfortunately, it doesn't appear to work, as it keeps creating an apk, instead of a bundle. Hence my question of how I can get it to create a bundle.

Is there something more I have to do than just adding the extra parameter to the command?

breautek:
@dfiggtns 

For ionic it's:
`ionic cordova build android -- -- --packageType="bundle"`

You have:
`ionic cordova build android --release -- --packageType=bundle`

Note the double `-- --` between android and --packageType.

dfiggtns:
I couldn't get cordova 8.1.0 to build anything other than an apk, so I resorted to using the following step between the ionic build and jarsigner steps:

cd ./platforms/android ; ./gradlew app:bundleRelease ; cd ../.. ;

which worked first time and the resulting signed bundle was accepted by the Play Store.

I hope that helps anyone with the same problem.

mosababubakr:
Hello @breautek 
How can i build my android app for .aab files in phonegap build ?
I am using phonegap build for building iOS and Android apps.

Regards

breautek:
If phonegap supports the `build.json` file, then you can use it to signal that you want a bundle. See https://cordova.apache.org/docs/en/dev/guide/platforms/android/index.html#using-buildjson

If it doesn't work... You'll have to ask the phonegap team. Cordova supports it as of cordova-android 8.1.0.

dfiggtns:
I've now tried using build.json to create an aab and it works!

My build.json is (located in the project root folder):

	{
	    "android": {
	        "debug": {
	            "keystore": "my.keystore",
	            "storePassword": "mypw",
	            "alias": "myalias",
	            "password" : "mypw",
	            "keystoreType": "",
	            "packageType": "apk"
	        },
	        "release": {
	            "keystore": "my.keystore",
	            "storePassword": "mypw",
	            "alias": "myalias",
	            "password" : "mypw",
	            "keystoreType": "",
	            "packageType": "bundle"
	        }
	    }
	}

and the build command is:

    ionic cordova build android --release -- --buildConfig=build.json

I hope that helps.

mosabab:
> I've now tried using build.json to create an aab and it works!
> 
> My build.json is (located in the project root folder):
> 
> ```
> {
>     "android": {
>         "debug": {
>             "keystore": "my.keystore",
>             "storePassword": "mypw",
>             "alias": "myalias",
>             "password" : "mypw",
>             "keystoreType": "",
>             "packageType": "apk"
>         },
>         "release": {
>             "keystore": "my.keystore",
>             "storePassword": "mypw",
>             "alias": "myalias",
>             "password" : "mypw",
>             "keystoreType": "",
>             "packageType": "bundle"
>         }
>     }
> }
> ```
> 
> and the build command is:
> 
> ```
> ionic cordova build android --release -- --buildConfig=build.json
> ```
> 
> I hope that helps.

the problem is that i use phonegap build to build the apps, so i don't know which solution or method can i used to make this codes in phonegap build

breautek:
> the problem is that i use phonegap build to build the apps, so i don't know which solution or method can i used to make this codes in phonegap build

May be best to ask the phonegap team. It's possible that they require some updates on their end to support the latest `cordova-android` version, which adds bundle support.

mosabab:
> > the problem is that i use phonegap build to build the apps, so i don't know which solution or method can i used to make this codes in phonegap build
> 
> May be best to ask the phonegap team. It's possible that they require some updates on their end to support the latest `cordova-android` version, which adds bundle support.

You are right, but unfortunately i put an issue in phoengap build github page about .aab file that not support in current phonegap cli (android 8.0.0), I think they should update the version of android to the latest one, but until now no anyone investigate about the issue

Guy-Sela:
> I couldn't get cordova 8.1.0 to build anything other than an apk, so I resorted to using the following step between the ionic build and jarsigner steps:
> 
> cd ./platforms/android ; ./gradlew app:bundleRelease ; cd ../.. ;
> 
> which worked first time and the resulting signed bundle was accepted by the Play Store.
> 
> I hope that helps anyone with the same problem.

Had a similar problem: I was getting .apk instead of .aab with all the above mentioned commands. What worked for me eventually was running: c:\Cordova\app_name\platforms\android>gradlew.bat bundle

Might also worth mentioning here is that gradle is still looking for ANDROID_HOME environment variable, although it's officially deprecated.

paknecht:
Thanks very much for this feature. But I have a question. Is it somehow possible to build an apk and an abb file at the same time? Or do I have to choose between these two? I would like to have both types built.

breautek:
> Thanks very much for this feature. But I have a question. Is it somehow possible to build an apk and an abb file at the same time? Or do I have to choose between these two? I would like to have both types built.

You can't do it at the same time, but you can run the build twice, one with the `--packageType=bundle` and the other without the `--packageType` (or `--packageType=apk`). There is two folders will appear in `platforms/android/app/build/outputs/`, `apk` for built APKs, and `bundle` for built bundles.

https://cordova.apache.org/docs/en/dev/guide/platforms/android/index.html#using-flags

paknecht:
> > Thanks very much for this feature. But I have a question. Is it somehow possible to build an apk and an abb file at the same time? Or do I have to choose between these two? I would like to have both types built.
> 
> You can't do it at the same time, but you can run the build twice, one with the `--packageType=bundle` and the other without the `--packageType` (or `--packageType=apk`). There is two folders will appear in `platforms/android/app/build/outputs/`, `apk` for built APKs, and `bundle` for built bundles.
> 
> https://cordova.apache.org/docs/en/dev/guide/platforms/android/index.html#using-flags

Thanks for the fast answer, but when I use a build .json file then i can not build both. I would have to switch to the build flags to do the two different builds.

PR #773, opened 2019-07-12
[WIP] lint: (semi)standard lint updates

brodybits:
- eslint-config-semistandard@14.0.0 update
- eslint-config-standard@13 update
- ignore 2 more rules in top-level `.eslintrc.yml`
- ignore yet another rule in `spec/.eslintrc.yml`

These changes seem to be OK back to Node.js 6.

FUTURE TODO items:

- eslint@6 & eslint-plugin-node@9 updates once we can drop Node.js 6 build
- enable some rules that are disabled in `.eslintrc.yml` & `spec/.eslintrc.yml`, and resolve the resulting lint issues

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -8,3 +8,5 @@ rules:
   padded-blocks: off
   operator-linebreak: off
   no-throw-literal: off
+  quote-props: off
+  prefer-const: off
--- a/package.json
+++ b/package.json
@@ -38,8 +38,8 @@
   },
   "devDependencies": {
     "eslint": "^5.12.0",
-    "eslint-config-semistandard": "^13.0.0",
-    "eslint-config-standard": "^12.0.0",
+    "eslint-config-semistandard": "^14.0.0",
+    "eslint-config-standard": "^13.0.1",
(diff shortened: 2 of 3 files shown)


erisu:
I would recommend not updating the `.eslintrc.yml` rules for consistent across all repos.

I would recommend just running the automatic fix first: `npm run eslint --  --fix` In most cases it will be good enough.

brodybits:
If we do want to make the changes for the new StandardJS entries `quote-props` & `prefer-const`, I would like to remove as many other exceptions as practical from `.eslint.yml`, like I just proposed in apache/cordova-common#82. I would like to get some feedback on apache/cordova-common#82 before continuing with this one.

janpio:
If this is a general thing that should be applied to all our repos, best create an issue at apache/cordova with some suggestions what should be changed, then we can discuss it and after that start reviewing PRs on the individual repos.

breautek:
I think this PR is pretty much invalid and stale now since @erisu is building and starting to deploy the eslint config plugin to enforce a standard lint ruleset across repos.

PR #770, opened 2019-07-12
[WIP] Add Node.js 12 to CI WIP with known test failure

brodybits:
This includes a change that @janpio pushed to the `janpio-travis_node_12` branch (draft PR #766).

~~I discovered that there is a hidden failure on Node.js 12, as reported by @breautek in #768. While output shows a single `F` character, the build seems ignore this failure on both Travis CI and AppVeyor CI.~~

~~I think that further discussion of the ignored test failure should continue in #768.~~

~~P.S. For some reason Travis CI still does not work on Node.js 12, which is very strange since this does work on my fork. But we can still see the ignored test failure on Node.js 12 on AppVeyor CI.~~

_Now with known test failure on Node.js 12, as reported in #767. The trick is to add `dist: trusty` setting near the beginning of `.travis.yml`._

PR #777, opened 2019-07-14
ci(travis): set `dist: trusty` in `.travis.yml`

brodybits:
to avoid errored Travis CI build on Node.js 12 (see WIP PR #770)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 sudo: false
 
+dist: trusty
+
 env:
   global:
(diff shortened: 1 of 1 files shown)


janpio:
Should we really merge this? In general this should work without these additions just out of the box :/

brodybits:
This would be needed if we want to test on Node.js 12, unless someone wants to do a bunch of extra investigation.

janpio:
Maybe add a TODO that it should be removed asap?

PR #778, opened 2019-07-14
test: use verbose spec reporter

brodybits:


brodybits:
Now merged. The one thing is that for some reason, I do not see the details in case of any test failures. The workaround is to remove the “helpers” item from a JSON file if I need to see such error details. Any input would be much appreciated.

raphinesse:
@brodybits I could reproduce the described behavior and isolated the cause. Working on a fix for it now.

PR #782, opened 2019-07-16
Do not clobber console.log to spy on it

raphinesse:
Fix for the problem described in https://github.com/apache/cordova-android/pull/778#issuecomment-511229607

Properties that are replaced with `spyOn` are restored after each test. The manually set spies that existed previously clobbered the `console.log` property permanently. This broke the test output with the new reporter.

### DIFF
--- a/spec/unit/android_sdk.spec.js
+++ b/spec/unit/android_sdk.spec.js
@@ -45,16 +45,14 @@ describe('android_sdk', () => {
     describe('print_newest_available_sdk_target', () => {
         it('should log the newest version', () => {
             const sortedIds = ['android-27', 'android-24', 'android-23', 'android-19'];
-            const logSpy = jasmine.createSpy('log');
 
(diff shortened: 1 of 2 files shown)


raphinesse:
> do you think we need to make this kind of fix in other Cordova packages?

The tests are often quite poorly written, so I wouldn't rule it out. But I would probably just fix these kinds of things when they pop up instead of starting to hunt for it now. :man_shrugging:

raphinesse:
Curiously, the previous code should never have caused any problems. See the last section _Dot notation_ in the [`rewire` docs](https://www.npmjs.com/package/rewire#caveats). :confused: So I'll take back my _poorly written code_ judgement from above :disappointed_relieved: 

Would be good to know what is going on here. I think I wrote similar code in other parts of Cordova.

brodybits:
Interesting. I would still favor merging this change since it at least *seems* to be cleaner and does seem to solve the observed problem for me.

It would certainly be nice if we can understand what was going on, though.

PR #784, opened 2019-07-16
Properly handle promise in create script

raphinesse:
This seemingly fixed a spurious error in `spec/unit/create.spec.js` that was occurring on my system.

### DIFF
--- a/bin/lib/create.js
+++ b/bin/lib/create.js
@@ -272,7 +272,7 @@ exports.create = function (project_path, config, options, events) {
     // Make the package conform to Java package types
     return exports.validatePackageName(package_name)
         .then(function () {
-            exports.validateProjectName(project_name);
+            return exports.validateProjectName(project_name);
         }).then(function () {
         // Log the given values for the project
             events.emit('log', 'Creating Cordova project for the Android platform:');
--- a/spec/unit/create.spec.js
+++ b/spec/unit/create.spec.js
@@ -213,6 +213,17 @@ describe('create', function () {
                     expect(err.message).toContain('Project already exists!');
                 }).done(done);
             });
+            it('should fail if validateProjectName rejects', () => {
+                const fakeError = new Error();
+                create.validateProjectName.and.callFake(() => Promise.reject(fakeError));
+
+                return create.create(project_path, config_mock, {}, events_mock).then(() => {
+                    fail('Expected promise to be rejected');
+                }, err => {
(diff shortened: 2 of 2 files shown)


bin/lib/create.js
```diff
@@ -272,7 +272,7 @@ exports.create = function (project_path, config, options, events) {
     // Make the package conform to Java package types
     return exports.validatePackageName(package_name)
         .then(function () {
-            exports.validateProjectName(project_name);
+            return exports.validateProjectName(project_name);
```
brodybits:
```suggestion
            return validateProjectName(project_name);
```

(maybe)

raphinesse: We can't do that, since calling the function on the `exports` object is necessary for stubbing via `spyOn` to be possible. The tests actually make use of that too.

`spyOn(create, 'validateProjectName')` in the spec file would stub `exports.validateProjectName` whereas `validateProjectName` would still refer to the original unstubbed function.

raphinesse:
> I do find it _really_ interesting that the tests, including the `validateProjectName` failure tests, did pass with this promise not properly handled. I am suspecting that this would not have been the case when using standard JavaScript promises.

The `validateProjectName` failure tests never execute that code. It's part of `exports.create`. The tests for `exports.create` OTOH stub `validateProjectName` and only verify that it has been called. So this couldn't have failed any of the existing tests.

brodybits:
> The `validateProjectName` failure tests never execute that code. It's part of `exports.create`. The tests for `exports.create` OTOH stub `validateProjectName` and only verify that it has been called. So this couldn't have failed any of the existing tests.

Hmm, I think we should add a test to cover this case.

If I would ignore this PR and inject the following change onto the `master` branch in my workarea:

```diff
diff --git a/bin/lib/create.js b/bin/lib/create.js
index 2fa457fa..623b655b 100755
--- a/bin/lib/create.js
+++ b/bin/lib/create.js
@@ -273,6 +273,7 @@ exports.create = function (project_path, config, options, events) {
     return exports.validatePackageName(package_name)
         .then(function () {
             exports.validateProjectName(project_name);
+            return Q.resolve();
         }).then(function () {
         // Log the given values for the project
             events.emit('log', 'Creating Cordova project for the Android platform:');
```

then `npm test` seems to succeed for me. What do you think?

raphinesse:
I didn't quite get your comment but I have added a regression test.

PR #788, opened 2019-07-18
Simplify apkSorter using compare-func package

raphinesse:
### Motivation and Context

When reviewing #779 I had a hard time understanding what the behavior of `apkSorter` was. I find that is a common issue with complex sort callbacks in JavaScript.

IMHO it's often makes for way more readable and less repetitive code to have a `sortBy` style approach. That's what I implemented in this PR for `apkSorter`.

### Description

Implement `apkSorter` in a `sortBy` style using the `compare-func` module.

### Testing

I used the existing unit tests. One of them had to be fixed, since it stubbed `fs.statSync` incorrectly (`mtime` is a `Date`, not a timestamp)

### DIFF
--- a/bin/templates/cordova/lib/builders/ProjectBuilder.js
+++ b/bin/templates/cordova/lib/builders/ProjectBuilder.js
@@ -27,6 +27,7 @@ var spawn = require('cordova-common').superspawn.spawn;
 var events = require('cordova-common').events;
 var CordovaError = require('cordova-common').CordovaError;
 var check_reqs = require('../check_reqs');
+const compareFunc = require('compare-func');
 
 const MARKER = 'YOUR CHANGES WILL BE ERASED!';
 const SIGNING_PROPERTIES = '-signing.properties';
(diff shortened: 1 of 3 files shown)


bin/templates/cordova/lib/builders/ProjectBuilder.js
```diff
@@ -300,27 +301,19 @@ class ProjectBuilder {
...
-    // De-prioritize arch-specific builds & unsigned builds
-    const lower = (fileName) => {
-        return archSpecificRE.exec(fileName)
-            ? -2
-            : unsignedRE.exec(fileName)
-                ? -1
-                : 0;
-    };
+    // Sort by file modification time, latest first
+    apkPath => -fs.statSync(apkPath).mtime.getTime(),
```
erisu: Is the `-` in `-fs.statSync(apkPath).mtime.getTime()` correct?

janpio: if it's not, would be "funny" that the tests pass anyway...

raphinesse: Yes, the minus creates a descending sort by `mtime`, which means bigger/later values will come before smaller/earlier ones.

PR #792, opened 2019-07-24
[MAJOR] Gradle upgrade from 4.10.3 to 6.x

### LINKED ISSUE #718: cordova's configured gradle version unsupported in Android Studio 3.4
# Bug Report

Android Studio 3.4 (just released) require a minimum Gradle version of 5.1.1, and so Gradle sync fails:
```
> Failed to apply plugin [id 'com.android.application']
   > Minimum supported Gradle version is 5.1.1. Current version is 4.10.3. If using the gradle wrapper, try editing the distributionUrl in /Users/jacob/workspace/f-tc/src-cordova/platforms/android/gradle/wrapper/gradle-wrapper.properties to gradle-5.1.1-all.zip
```

## Problem

See above

cordova android automatically configured for Gradle 5.1.

uses Gradle 4.10

## Checklist

- [ ] I searched for existing GitHub issues
- [ ] I updated all Cordova tooling to most recent version
- [ ] I included all the necessary information above

### LINKED ISSUE #754: gradle wrapper broken
# Bug Report
gradle wrapper broken

## Problem
cordova platform add android
cordova platform add browser
gradle wrapper --gradle-version 5.1.1
cordova prepare

app\build.gradle : line:42

```
task wrapper(type: Wrapper) {
    gradleVersion = '5.1.1'
}
```

app\build.gradle : line:42

```
task wrapper(type: Wrapper) {
    gradleVersion = '4.10.3'
}
```

## Information

`9.0.0 (cordova-lib@9.0.1)`

## Checklist

### LINKED ISSUE #780: [MAJOR] Update for Gradle version 5
From discussion in #718 it sounds like we should update cordova-android to use Gradle 5.

I think this should be considered a MAJOR BREAKING CHANGE, for a new major release.

Any contributions would be much appreciated.

This may also be related to #754.

### LINKED ISSUE #834: Gradle Sync Issues: "NdkCompile is no longer supported"
# Bug Report
INFO: The following project options are deprecated and have been removed: 
android.useDeprecatedNdk
NdkCompile is no longer supported
Affected Modules: CordovaLib, app

### environment：
Android Studio3.5
cordova  version 9.0.0 (cordova-lib@9.0.1)
classpath 'com.android.tools.build:gradle:3.5.0'
distributionUrl=https\://services.gradle.org/distributions/gradle-5.4.1-all.zip

breautek:
### Platforms affected
Android

### Motivation and Context

Updates gradle version to the latest release (currently 6.1). It solves a few issue tickets that supposedly was caused after upgrading Android Studio.

Fixes #780 
Fixes #754 
Fixes #718 
Fixes #834 

### Description
Changed the desired gradle version from 4.10.3 to 6.1

### Testing

I have tested using `npm test` and all tests passed. Note that if you have ran `npm test` before, you will need to remove the `/test/gradlew` file so the test project will download the new gradle version.

I have also tested by creating a brand new cordova app, and adding the `cordova-android` platform by using `cordova platform add https://github.com/breautek/cordova-android.git#gradle-5.x`
Then running the `cordova build android` command.

I'd also like to make a note that gradle 5.5.1 does currently produce a warning stating:

> Deprecated Gradle features were used in this build, making it incompatible with Gradle 6.0.

Running gradlew manually with `--verbose --stacktrace`  shows the code triggering the warning is android internal code.

### Checklist

### DIFF
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -29,7 +29,7 @@ buildscript {
     dependencies {
         apply from: '../CordovaLib/cordova.gradle'
 
-        classpath 'com.android.tools.build:gradle:3.5.2'
+        classpath 'com.android.tools.build:gradle:3.5.3'
 
         if(cdvHelpers.getConfigPreference('GradlePluginGoogleServicesEnabled', 'false').toBoolean()) {
             String defaultGradlePluginGoogleServicesVersion = '4.2.0'
@@ -61,7 +61,7 @@ allprojects {
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '6.0'
+    gradleVersion = '6.1'
 }
 
 // Configuration properties. Set these via environment variables, build-extras.gradle, or gradle.properties.
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -61,7 +61,7 @@ allprojects {
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '4.10.3'
+    gradleVersion = '5.5.1'
 }
 
 // Configuration properties. Set these via environment variables, build-extras.gradle, or gradle.properties.
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -29,7 +29,7 @@ buildscript {
     dependencies {
         apply from: '../CordovaLib/cordova.gradle'
 
-        classpath 'com.android.tools.build:gradle:3.3.0'
+        classpath 'com.android.tools.build:gradle:3.5.2'
 
         if(cdvHelpers.getConfigPreference('GradlePluginGoogleServicesEnabled', 'false').toBoolean()) {
             String defaultGradlePluginGoogleServicesVersion = '4.2.0'
@@ -61,7 +61,7 @@ allprojects {
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '5.6.2'
+    gradleVersion = '6.0'
 }
 
 // Configuration properties. Set these via environment variables, build-extras.gradle, or gradle.properties.
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -61,7 +61,7 @@ allprojects {
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '5.5.1'
+    gradleVersion = '5.6.2'
 }
 
 // Configuration properties. Set these via environment variables, build-extras.gradle, or gradle.properties.
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "e2e-tests": "jasmine --config=spec/e2e/jasmine.json",
     "java-unit-tests": "node test/run_java_unit_tests.js",
     "eslint": "eslint . \"bin/**/!(*.*|gitignore)\"",
-    "clean": "gulp clean"
+    "clean-tests": "node bin/clean_test.js"
   },
   "author": "Apache Software Foundation",
   "license": "Apache-2.0",
@@ -46,7 +46,6 @@
     "eslint-plugin-node": "^8.0.1",
     "eslint-plugin-promise": "^4.0.1",
     "eslint-plugin-standard": "^4.0.0",
-    "gulp": "^4.0.2",
     "jasmine": "^3.5.0",
     "jasmine-spec-reporter": "^4.2.1",
     "nyc": "^14.1.1",
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "e2e-tests": "jasmine --config=spec/e2e/jasmine.json",
     "java-unit-tests": "node test/run_java_unit_tests.js",
     "eslint": "eslint . \"bin/**/!(*.*|gitignore)\"",
-    "clean": "git clean -fdx -e node_modules"
+    "clean": "gulp clean"
   },
   "author": "Apache Software Foundation",
   "license": "Apache-2.0",
@@ -46,6 +46,7 @@
     "eslint-plugin-node": "^8.0.1",
     "eslint-plugin-promise": "^4.0.1",
     "eslint-plugin-standard": "^4.0.0",
+    "gulp": "^4.0.2",
     "jasmine": "^3.5.0",
     "jasmine-spec-reporter": "^4.2.1",
     "nyc": "^14.1.1",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,8 @@
     "cover": "nyc jasmine --config=spec/coverage.json",
     "e2e-tests": "jasmine --config=spec/e2e/jasmine.json",
     "java-unit-tests": "node test/run_java_unit_tests.js",
-    "eslint": "eslint . \"bin/**/!(*.*|gitignore)\""
+    "eslint": "eslint . \"bin/**/!(*.*|gitignore)\"",
+    "clean": "git clean -fdx -e node_modules"
   },
   "author": "Apache Software Foundation",
   "license": "Apache-2.0",
--- a/bin/clean_test.js
+++ b/bin/clean_test.js
@@ -1,4 +1,3 @@
-
 /*
        Licensed to the Apache Software Foundation (ASF) under one
        or more contributor license agreements.  See the NOTICE file
@@ -20,21 +19,10 @@
 
 const fs = require('fs');
 
-const cleanFiles = [
-    './test/gradlew',
-    './test/gradlew.bat'
-];
+if (fs.existsSync('./test/gradlew')) {
+    fs.unlinkSync('./test/gradlew');
+}
 
-exports.clean = () => {
-    for (let i = 0; i < cleanFiles.length; i++) {
-        try {
-            let file = cleanFiles[i];
-            if (fs.existsSync(file)) {
-                fs.unlinkSync(file);
-            }
-        } catch (ex) {
-            return Promise.reject(ex);
-        }
-    }
-    return Promise.resolve();
-};
+if (fs.existsSync('./test/gradlew.bat')) {
+    fs.unlinkSync('./test/gradlew.bat');
+}
--- a/bin/templates/cordova/lib/builders/ProjectBuilder.js
+++ b/bin/templates/cordova/lib/builders/ProjectBuilder.js
@@ -226,7 +226,7 @@ class ProjectBuilder {
                 // If it's not set, do nothing, assuming that we're using a future version of gradle that we don't want to mess with.
                 // For some reason, using ^ and $ don't work.  This does the job, though.
                 var distributionUrlRegex = /distributionUrl.*zip/;
-                var distributionUrl = process.env['CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL'] || 'https\\://services.gradle.org/distributions/gradle-5.5.1-all.zip';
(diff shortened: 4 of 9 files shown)


breautek:
This PR is intended for the android 9.x milestone.

janpio:
> Note that if you have ran `npm test` before, you will need to remove the `/test/gradlew` file so the test project will download the new gradle version.

We should open an issue for this so this can get fixed.

> I'd also like to make a note that gradle 5.5.1 does currently produce a warning stating:
> 
> > Deprecated Gradle features were used in this build, making it incompatible with Gradle 6.0.
> 
> Running gradlew manually with `--verbose --stacktrace` shows the code triggering the warning is android internal code.

So nothing we can do about?

breautek:
> So nothing we can do about?

Not really. There is an open bug ticket on [Android Issue Tracker](https://issuetracker.google.com/issues/127538936).

erisu:
`git clean -fdx` should be enough as well. This is typically what I use to clean out any left over files and folder that are untracked.

breautek:
> I think `git clean -fdx` should be enough as well. This is typically what I use to clean out any left over files and folder that are untracked.

Actually ran into an issue on "cleaning" `gradlew` as I was writing the script I remembered that not everybody is going to be running in a bash environment...

I don't think windows have different commands for using `rm` and I don't think they support the `&&` syntax in their command line to run another command immediately after the first command... (could be wrong, been forever since I used windows...)

using `git clean -fdx` will also remove `node_modules`, which will require using `npm install` again, so I used this, I'd still have to run a second command.

So I think the best way to handle this is to use a task system such as `gulp` or `grunt` so the clean code can be ran in Node, allowing more cross-platform flexibility. Currently the repository doesn't have either and I'm not sure if such systems are okay in Cordova, or if it is, if one is preferred over the other.

erisu:
> I don't think they support the && syntax in their command line to run another command immediately after the first command

`&&` is supported and runs only if the first command was successful. This worked since Windows NT I believe.

> using git clean -fdx will also remove node_modules, which will require using npm install again, so if I used this, I'd still have to run a second command and run into the issue described above.

```git clean -nfdx -e node_modules```

**Comment by breautek** [comment]
I've added a clean script that does `git clean -fdx -e node_modules`. I run linux and it works excellent for this case as well. Thanks for the suggestion!

So anybody with existing repositories can now simply run `npm run clean` and it will remove gradle and other build artefacts. I don't call this automatically as I think that might be rather annoying having gradle be redownloaded/redeployed everytime running `npm test`.

**Comment by Kevin-Hamilton** [comment]
This assumes 1) that the project is currently being managed under git, and 2) that the developer does not have any untracked or uncommitted work anywhere in the current git repository (which may be tracking files and folders beyond the scope of the package.json that this command is in).

I think this is a bad idea.  Unless I am misunderstanding the target user for the `npm run clean` command?

**Comment by breautek** [comment]
The clean command is something that a developer would use when working in the `cordova-android` repo. It's not intended to be used as a part of a broad script.

However, there is some annoying dangers. I've incorporated using git clean in my workflow in another project and accidentally deleted newly added but untracked files.

So maybe just writing a gulp script might be better, to avoid accidents like that.

**Comment by breautek** [comment]
@erisu FYI, I opted away from the `git clean` method as you suggested. I originally liked the idea and even used it my own personal project. However, I had a situation where I accidentally deleted untracked files that was intended to added to the repo because `git clean` was ran as a part of a script.

Due to this danger I decided to replace it with a `gulpfile` implementation. I see that `gulp` is already used in the `cordova-docs` library, so I assume it wouldn't be a big deal adding a dev dependency here. The gulpfile contains a `clean` task to clean files that we need to specifically clean, which is currently `./test/gradlew` && `./test/gradlew.bat`.

File: package.json [hunk 1]
```diff
@@ -24,7 +24,8 @@
     "cover": "nyc jasmine --config=spec/coverage.json",
     "e2e-tests": "jasmine --config=spec/e2e/jasmine.json",
     "java-unit-tests": "node test/run_java_unit_tests.js",
-    "eslint": "eslint bin spec test"
+    "eslint": "eslint bin spec test",
+    "clean": "gulp clean"
```
piotr-cz: Perhaps this may be achieved without adding gulp as a dependency (via `clean.js`)?

breautek: Project search for `clean.js` yielded no results.

I have found `templates/cordova/lib/clean` however but this is for cleaning the project directory, not the test directory.

To be clear, the `clean` npm script is only for development on the `cordova-android` to reset gradle. It's not for anybody who uses `cordova-android` as part of their project via `cordova platform add`.

Maybe I should write a comment explaining as such? I believe this is the second time there was a misunderstanding.

breautek: Also worth mentioning that gradle doesn't need to be "cleaned" for cordova apps because to upgrade the `cordova-android` platform, you need to remove the platform (and thus destroying the gradle installation with it) and then re-adding the platform, which will install whatever gradle version as defined in `cordova-android`.

piotr-cz: What I meant is that the contents of [gulpfile.js](https://github.com/apache/cordova-android/pull/792/files#diff-b9e12334e9eafd8341a6107dd98510c9) could be moved to separate script - for example `bin/clean.js` which could be run via npm.

`package.json`:

```
{
  "scripts": {
    "clean": "node bin/clean.js"
  }
}
```

breautek: Ahh gotcha. I can move the `gulpfile` to `bin/`. Although gulp is a task system, so you would run the file with `gulp`, and not node directly and I would favour keeping the filename `gulpfile.js` because of this. Reason for adding this tool is to have easier script to clean the test gradle stuff, without relying on `git` commands. It's a dependency that some other cordova repos already use so it should be somewhat familiar.

Reason this is being added as part of this PR is because it does require gradle to be removed for the test directory to use the new version of gradle. This would only apply to other developers who already have this repository cloned and they have ran `npm test` before.

erisu: I am also not in favor of adding the `gulp` dependency to perform a simple task that could have been achieved with just node.

I will let someone else though review this.

breautek: > I am also not in favor of adding the `gulp` dependency to perform a simple task that could have been achieved with just node.
> 
> I will let someone else though review this.

Fair enough. My thought process was to use something existing for familiarity. But it is true that this isn't solving a complex problem so `gulp` isn't really as necessary compared to `cordova-docs` (where `gulp` is used).

I can remove `gulp` dependency and write a simple node script to remove that files that needs to be removed if you want gradle to reinstall for `npm test` and I can place this node script in the `bin/` directory which I believe will solve both critiques.

Does this sound right?

timbru31: Sounds good to me!

breautek: I've made the changes as discussed. Resolving this conversation.

brodybits: > I've made the changes as discussed.

18a9e7d72615e748be01dde3a65a4a7c3ee14383 - rare case where removing a dependency results in *less* script code (sorry if this is too far off-topic)

breautek:
> LGTM! I'd like another maintainer to have a look, too, though. :)

This PR is intended for `9.x`, perhaps a branch should be created on the apache repo for `9.x`? Currently this PR is actually targeting apache's `master`.

brodybits:
LGTM for 9.0. I did verify from [[1]](https://services.gradle.org/distributions/) that this proposal is referencing the most recent Gradle version to date.

Here are my *personal* feelings:

* I would downvote making a new branch for 9.x due to the potential for confusion and extra merging that may be needed as a consequence.
* I would favor committing this change now and *maybe* using cordova-coho to set the version to 9.0.0-dev. Then 8.x patch releases could still be made from the existing `8.1.x` branch, if needed.

I wonder if we should discuss in <dev@cordova.apache.org> before merging?

[1] https://services.gradle.org/distributions/

breautek:
Note to myself: https://github.com/apache/cordova-android/issues/834#issuecomment-553973463 may be affected by this PR. I should double check this.

piotr-cz:
Gradle 6.0 is out.
I wonder if there is a B/C coming anyway maybe it's worth to skip to latest Gradle version?

breautek:
> Gradle 6.0 is out.
> I wonder if there is a B/C coming anyway maybe it's worth to skip to latest Gradle version?

Definitely worth looking into.

bin/templates/project/app/build.gradle
```diff
@@ -40,7 +40,7 @@ allprojects {
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '4.10.3'
+    gradleVersion = '6.0'
```
timbru31: aaaand 6.0.1 is out: https://github.com/gradle/gradle/releases/tag/v6.0.1

breautek: Yah, I've been passively keeping this PR up to date with the latest version and before the merge I'd suggest we do one last final check.

timbru31: I know, was just a friendly reminder :)

breautek: Note to myself mainly but before I commit to 6.x, I need to double check things still works fine in android studio, because according to the [docs](https://developer.android.com/studio/releases/gradle-plugin) the latest gradle plugin only supports up to 5.6.4.

Been awhile since I looked at this but in my testing I probably only used the cordova command line.

erisu: Bumped Gradle to `6.1`
https://github.com/gradle/gradle/releases/tag/v6.1.0

wilywork:
@erisu command "cordova platform add github:breautek/cordova-android\#gradle-5.x" not work for me..

Any solution ?

```
C:\android_build>cordova platform add github:breautek/cordova-android\#gradle-5.x
Using cordova-fetch for github:breautek/cordova-android\#gradle-5.x
Failed to fetch platform github:breautek/cordova-android\#gradle-5.x
Probably this is either a connection problem, or platform spec is incorrect.
Check your connection and platform name/version/URL.
Error: npm: Command failed with exit code 1 Error output:
npm ERR! Error while executing:
npm ERR! C:\Program Files\Git\cmd\git.EXE ls-remote -h -t ssh://git@github.com/breautek/cordova-android%5C.git
npm ERR!
npm ERR! Host key verification failed.
npm ERR! fatal: Could not read from remote repository.
npm ERR!
npm ERR! Please make sure you have the correct access rights
npm ERR! and the repository exists.
npm ERR!
npm ERR! exited with error code: 128

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\xandy\AppData\Roaming\npm-cache\_logs\2020-01-21T12_36_03_938Z-debug.log```

erisu:
There might a difference between the Windows and Linux/macOS syntax when passing in the GitHub URL.

For example, I escape the `#` which is needed. But for Windows maybe that is not required and causes your problem.

Additionally, since this PR had been merged, you can use the Apache repo instead. I would not recommend using his branch as he could delete the branch at any time.

Try what @RafaelKr suggested.

Lastly, please note that we do not consider or suggest using the master branch in production. Master is still an active development branch and would not go through an official release vote.

Testing and reporting anything is always welcomed.

breautek:
You can only use `ssh` urls if you have write access to the repo, which in this case he does not.

Additionally, because this PR is merged in, I'll be deleting my `gradle-5.x` branch, so using the Apache repo will now be required.

Thanks @erisu for picking this up.

RafaelKr:
@breautek No, I also thought the problem comes from using SSH. But @erisu mentioned the escaped # and that's right.

His log says `Files\Git\cmd\git.EXE ls-remote -h -t ssh://git@github.com/breautek/cordova-android%5C.git` and `%5C` is the url encoded backslash.

@wilywork try this: `cordova platform add github:apache/cordova-android`
@breautek just deleted his branch, so all the other mentioned methods won't work anymore.

wilywork:
i find the problem, for windows use ... android\\#gradle-5.x without \ :

cordova platform add github:breautek/cordova-android#gradle-5.x

Thank's

PR #800, opened 2019-08-06
Stop webview from restarting when activity resizes

### LINKED ISSUE #799: App is restarted on split-screen
# Bug Report

App keeps running in split screen, just resized.

Activity is restarted.

### Command or Code

Press and hold the "square" button on any modern Android device to trigger split screen mode. Notice how the Cordova app is restarted.

This is fixed if I add the following to my `config.xml`:

    <edit-config file="AndroidManifest.xml" target="/manifest/application/activity[@android:name='MainActivity']" mode="merge">
      <activity android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|smallestScreenSize|screenLayout|uiMode" />
    </edit-config>

This is the default list of `configChanges` with `|smallestScreenSize|screenLayout|uiMode` appended. See docs: https://developer.android.com/guide/topics/manifest/activity-element#config

I propose that at least `smallestScreenSize|screenLayout` are added to the default `AndroidManifest.xml`. `uiMode` is not necessary to fix this issue, but should maybe be in the default list as well.

Android 7.1.2

### Version information

Cordova 9.0.0  
Cordova Android 8.0.0

## Checklist

- [x] I searched for existing GitHub issues (#744 is a duplicate, but folks explicitly requested a new issue)

breautek:
### Platforms affected
Android

### Motivation and Context

This change stops the webview from restarting when the activity window resizes while using Android's split-screen feature.

Fixes #799

### Description
This change adds some new attributes to the default list of `configChanges` including:

#### smallestScreenSize

> The current available screen size has changed.
This represents a change in the currently available size, relative to the current aspect ratio, so will change when the user switches between landscape and portrait.

#### screenLayout

> The screen layout has changed — a different display might now be active.

#### uiMode

>The user interface mode has changed — the user has placed the device into a desk or car dock, or the night mode has changed. For more information about the different UI modes, see [UiModeManager](https://developer.android.com/reference/android/app/UiModeManager.html).

All of these will allow the webview to respond to screen changes more gracefully allowing the webview to adapt to the new change without restarting the webview or reloading the page.

### Testing
I've ran `npm test` and tested on a physical Samsung Galaxy Note 9 device running Android 9.

Special thanks to @ttencate for not only opening up an issue, but also providing the solution to the issue.

### Checklist

### DIFF
--- a/bin/templates/project/AndroidManifest.xml
+++ b/bin/templates/project/AndroidManifest.xml
@@ -37,7 +37,7 @@
                 android:launchMode="singleTop"
                 android:theme="@android:style/Theme.DeviceDefault.NoActionBar"
                 android:windowSoftInputMode="adjustResize"
-                android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale">
+                android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|smallestScreenSize|screenLayout|uiMode">
             <intent-filter android:label="@string/launcher_name">
                 <action android:name="android.intent.action.MAIN" />
                 <category android:name="android.intent.category.LAUNCHER" />


breautek:
Pinging for review

@brodybits
@erisu 

And I'm not sure who else usually reviews android PR. Please ping anybody else as necessary. Thanks.

breautek:
> I think this should be considered for a minor or major

I'm find with that, the current master is `8.1.0-dev` so that's a pending minor release I assume?

> I wonder if we want to find a way to configure these items in config.xml

If we wanted more flexibility sure, but I think this would be more appropriate for a feature request ticket.

timbru31:
No veto from any others in over one week, merging! :)

PR #831, opened 2019-09-24
Ignore auto-generated eclipse buildship files

breautek:
### Platforms affected
None

### Motivation and Context

Eclipse buildship `.project` files are being added to several directories automatically from my code editor (VS Code) when working on `cordova-android` tasks. I think it makes sense for these files to be git ignored.

### Description

Added `**/.project` line in `.gitignore`

### Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,5 @@ node_modules/
 coverage/
 .nyc_output/
 package-lock.json
+# Eclipse Buildship files
+**/.project
--- a/.gitignore
+++ b/.gitignore
@@ -47,4 +47,4 @@ coverage/
 .nyc_output/
 package-lock.json
 # Eclipse Buildship files
-**/.project
+.project


.gitignore
```diff
@@ -46,3 +46,5 @@ node_modules/
 coverage/
 .nyc_output/
 package-lock.json
+# Eclipse Buildship files
+**/.project
```
raphinesse: This will match the same files:
```suggestion
.project
```
From https://git-scm.com/docs/gitignore#_pattern_format:
> If there is a [directory] separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular `.gitignore` file itself. Otherwise the pattern may also match at any level below the `.gitignore` level.

timbru31: I'm with Raphael, to avoid accidentally merging, I'm pressing the "request changes" button.

PR #854, opened 2019-10-20
Ensure to lint as many files as possible

raphinesse:
- Ensure to lint as many files as possible
  - Binaries w/out file extension and `cordova-js-src/*` hadn't been linted before
- Apply fixes where necessary


BBosman:
Won’t the changes to ‘bin/templates/cordova/version’ be reverted the next time coho runs?

Or is there a PR in the works to update coho as well?

raphinesse:
@BBosman Good point! I hadn't thought of that. No, there's nothing in the works, but it would definitely make sense to update `coho` to conform to our style guide. If you happen to know where the magic happens, a PR would be welcome.

PR #859, opened 2019-11-02
[Major] Removed unnecessary project name restriction

### LINKED ISSUE #584: Error validating project name. Project name must not begin with a number.
Why is there a restriction to this?
Google Play does not have this restriction.

Is there anything that is affected by this?

### LINKED ISSUE #701: Project cannot name cannot start with a number
Hello everyone,

I already commented here: https://github.com/apache/cordova-android/issues/584, however I was instructed to open a new issue to ask for support - so here we are.

I've adopted this project from someone else, so my knowledge of Cordova isn't all that great, although I have spent considerable amounts of time in the documentation to understand what I am supporting. Anyway, as described in my other comment, here is my issue:

> I am really struggling with this currently. I need to build some apps whose names begin with a number - however when I run the Cordova build command it seems to try and name my project after whatever is in the name field in my config.xml. If I change this to not be a string beginning with a number, it also renames the app this as well which I need to figure out how to avoid.
> 
> I should mention I am using this in a pipeline so manual intervention isn't necessarily ideal. Is there anything I can change in config.xml or anywhere else that will allow me to specify a project name that doesn't begin with a number, but an app name that does? I tried using "<name short=..." to set the name as one beginning with a number - am I on the right track? Am I missing something?
> 
> Please help me!

Thanks in advance!

breautek:
### Platforms affected
Android

### Motivation and Context

There was a condition in `validateProjectName` that prevented the project name from started with a number. It was found that this was added approximately 5 years ago, and it was assumed that once upon a time, this was because project names were used to generate a java class, based on the comments. However, that no longer appears to be the case.

Being able to support project names that starts with a number is important and perfectly valid (e.g: [8 Ball Pool](https://play.google.com/store/apps/details?id=com.miniclip.eightballpool)).

It fixes #584 and fixes #701  where lots of people requested this change.

### Description

Simply removed the condition preventing project names from starting with a number. `validateProjectName` has two other checks which were left untouched. I've also updated the comments to remove any references suggesting that successful `validateProjectName` calls means its safe to use the string as a java class name.

There was a unit test test for failure for project names starting with numbers which have been removed. Two items were added for the `valid` list for valid name checks which includes names starting with numbers.

### Testing

Manual testing using:
- `cordova platform add`
- `cordova build android`
- `cordova run android`

### Checklist


breautek:
This PR does alter a function that is exported. While I saw no evidence that this PR will be a breaking change, there is still a possibility. Therefore this PR should only be applied on the next major (9.x)

bin/lib/create.js
```diff
@@ -225,11 +225,6 @@ function validateProjectName (project_name) {
         return Q.reject(new CordovaError(msg + 'Project name cannot be CordovaActivity'));
```
raphinesse: If this name is not used as a class name, then this check would probably be superfluous too, right? And that would mean this whole function basically degenerates to a check if `project_name` is truthy. Which means it could be removed altogether.

_Note_: I'm just extrapolating based on the information you gave in the PR description.

breautek: You're probably right there. I'll do some testing with that removed.

raphinesse: Great! :+1:

breautek: Removed that condition, kept the empty name check as that is still a perfectly valid check. Still able to build and run the sample cordova app. :+1:

bin/lib/create.js
```diff
@@ -209,7 +209,7 @@ function validatePackageName (package_name) {
 }
 
 /**
- * Test whether a project name is acceptable for use as an android class.
+ * Test whether a project name is acceptable for use as a name
```
raphinesse:
```suggestion
 * Test whether given string is acceptable for use as a project name
```

breautek:
Just a reminder that this PR should only be merged once master has been upgraded to 9.x

timbru31:
Yes, as the [major] title suggests - side note: I only approve but leave the merging to the author of the PR (unless it's no Apache org member)

erisu:
@breautek I think this PR is ready for merging after applying a rebase to fix the conflict.

PR #890, opened 2020-01-05
chore: bump version to 9.0.0-dev

erisu:
### Motivation and Context

Prepare repo for next major development and release. 

### Description

Bump repo's version to `9.0.0-dev` in:
* `package.json`
* `VERSION`
* `bin/templates/cordova/version`
* `bin/templates/project/assets/www/cordova.js`
* `framework/build.gradle`
* `framework/src/org/apache/cordova/CordovaWebView.java`

### Testing

- `npm t`

### Checklist


erisu:
@brodybits I am sure we could use coho since it contains already a step that does the version bumping for the releases, but I was not investing the time to include it into my development process. For now, this manual process was much easier and faster.

Also, as a side note, I had been looking into the idea of removing all these version locations and condense it down between one or two file locations only.

PR #894, opened 2020-01-15
(android) Additional WebView preferences: 'UseWideViewport', 'SaveFormData', 'SavePassword'

AndreyZM:
### Platforms affected
android

### Motivation and Context
Now its not possible to set some useful android WebView settings such UseWideViewport, that give a developer define meta viewport tag in cordova apps. Also 'SaveFormData' & 'SavePassword' give more browser like user experience.

### Description
Values for WebView settings are loads from config.xml preferences: 
- UseWideViewport 
- SaveFormData 
- SavePassword

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/engine/SystemWebViewEngine.java
+++ b/framework/src/org/apache/cordova/engine/SystemWebViewEngine.java
@@ -148,6 +148,8 @@ public class SystemWebViewEngine implements CordovaWebViewEngine {
         webView.setVerticalScrollBarEnabled(false);
         // Enable JavaScript
         final WebSettings settings = webView.getSettings();
+        settings.setUseWideViewPort(preferences.getBoolean("UseWideViewPort", false));
+
         settings.setJavaScriptEnabled(true);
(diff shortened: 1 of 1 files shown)


dpogue:
`setSaveFormData` and `setSavePassword` are deprecated methods, so the best thing to do is probably to remove them from the code entirely. The behaviour of saving (or not) can be managed by the app author on a more specific level using HTML attributes.

I think we always want `setUseWideViewPort(true)` so that it obeys the viewport as specified by the app author.

As much as possible, we should allow things to be controlled by the app author via their HTML as opposed to via config preferences.

NiklasMerz (changes requested):
`setSaveFormData` and `setSavePassword` should not be part of this PR as they are deprecated.

See #1274

erisu:
I am going to close out this PR in favor of the privded feedback from the others:
* Per Android docs, the `setSaveFormData` & `setSavePassword` is deprecated.
* The `setUseWideViewPort` should remain default as @dpogue points out to obeys the viewport specified by the app author.

PR #895, opened 2020-01-15
test: move ci to gh-actions

erisu:
### Motivation and Context

Use GitHub Actions instead of third-party CI services.

### Description

* Created GitHub Actions Task
* Move all Travis CI & AppVeyor jobs to GitHub Actions
* Run only one Travis CI job for `codecov` reports. (Job Environment: Linux + Node 12)

Codecov has a better UI but requires setting secrets in GH which we do not have the direct ability to set. If we want to continue supporting and using Codecov, we would need to community with INFRA to set the secret token.

### Testing

- https://github.com/erisu/cordova-android/commit/0e3bda75114f84788ff6b5319b65dd62e0deda02/checks?check_suite_id=401547626

### Checklist


erisu:
Please ignore that the AppVeyor CI service is failing. AppVeyor would be turned off as this would replace it.

PR #896, opened 2020-01-16
feat: add kotlin support

erisu:
### Motivation and Context

To finish off the change requests and close out PR #441

closes: #441

### Description

What this PR does:

* Cherry-picks commit changes from #441
* Rename preference option `EnableKotlin` to `GradlePluginKotlinEnabled` (Default: `false`)
  * Since the preference has not been released or used in any other repos, the name change is safe. This change is only an update from the original PR's proposal.
* Added preference options `GradlePluginKotlinCodeStyle` (Default: `official`)
* Added preference option `GradlePluginKotlinVersion` (Default: `1.3.50`)
* Apply `kotlin-android-extensions` plugin as well when `GradlePluginKotlinEnabled` is `true`
* Changed `compile` to `implementation`

### Testing

- `npm t`

### Checklist

- [ ] I've updated the documentation if necessary (*Pending Work for Cordova Docs Repo*)


erisu:
@timbru31

I went ahead and created a new Kotlin support PR so we can quickly apply all the changes and improvements to get Kotlin prepared and released hopefully in the next day or 2.

Per your change request in the previous PR, I have:

* Updated Kotlin to the latest version

* Replaced `mavenCentral` with `jcenter` in the `test/app/build.gradle`
  
**NOTICE**: 
This change is not applied to the actual `build.gradle` that is used when building a Cordova project. The actual file, `bin/templates/project/app/build.gradle`, still uses in the following order:

    mavenCentral -> google -> jcenter

If these are to be updated, it should be taken care of in a separate PR

* Per your comment about `kotlin:kotlin-stdlib-jdk7` I believe it is best to keep it as `kotlin:kotlin-stdlib-jdk7`. A newly created Android project with the official Android Studio uses `kotlin:kotlin-stdlib-jdk7`. Probably best to not do something different.

erisu:
@timbru31 I just added the Kotlin source set as well, per your change request's overview comment. The only thing I was not sure is if it was necessary to wrap that behind a flag too.

dpogue:
> @timbru31 I just added the Kotlin source set as well, per your change request's overview comment. The only thing I was not sure is if it was necessary to wrap that behind a flag too.

IMO no need for a flag for adding the src/main/kotlin folder. Note that currently all plugin source files will end up in src/main/java because there's no special handling to detect kotlin files.

bin/templates/project/build.gradle
```diff
@@ -19,16 +19,16 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
+    ext.kotlin_version = '1.3.50'
```
timbru31: is this the Android Studio default?

erisu: Yes, if you create a blank "No Activity" project in Android Studio, it has in the project's `build.gradle`:

```
ext.kotlin_version = '1.3.50'
```

PR #900, opened 2020-01-16
refactor: simplify doFindLatestInstalledBuildTools

raphinesse:
### Motivation and Context

Just minor code cleanup

### Description

- simplify `doFindLatestInstalledBuildTools`
- remove `getAvailableBuildTools`
- update com.g00fy2:versioncompare
- use string interpolation for error messages

### Testing

`npm t`

### DIFF
--- a/framework/cordova.gradle
+++ b/framework/cordova.gradle
@@ -194,6 +194,6 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.g00fy2:versioncompare:1.3.1@jar'
+        classpath 'com.g00fy2:versioncompare:1.3.4@jar'
     }
 }
--- a/framework/cordova.gradle
+++ b/framework/cordova.gradle
@@ -61,16 +61,18 @@ String doFindLatestInstalledBuildTools(String minBuildToolsVersionString) {
         .max()
 
     if (highestBuildToolsVersion == null) {
-        throw new RuntimeException(
-            "No installed build tools found. Install the Android build tools version " +
-            minBuildToolsVersionString + " or higher.")
+        throw new RuntimeException("""
+            No installed build tools found. Install the Android build tools
+            version ${minBuildToolsVersionString} or higher.
+        """.replaceAll(/\s+/, ' ').trim())
     }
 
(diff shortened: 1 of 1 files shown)


framework/cordova.gradle
```diff
@@ -40,40 +40,40 @@ String doGetProjectTarget() {
...
         throw new RuntimeException(
             "No installed build tools found. Install the Android build tools version " +
             minBuildToolsVersionString + " or higher.")
     }
+
+    if (highestBuildToolsVersion.isLowerThan(minBuildToolsVersionString)) {
+        throw new RuntimeException(
+            "No usable Android build tools found. Highest installed version is " +
+            highestBuildToolsVersion.getOriginalString() + "; minimum version required is " +
+            minBuildToolsVersionString + ".")
```
erisu: For informational purposes, since Gradle scripts are written in Groovy, using double quotations gives you string interpolation. String literals works 👍 

These three lines could be written in one as,

```
"No usable Android build tools found. Highest installed version is ${highestBuildToolsVersion.getOriginalString()}; minimum version required is ${minBuildToolsVersionString}."
```

If this way is preferred just because the string is long, I would suggest single quotes so there is no string interpolation.

framework/cordova.gradle
```diff
@@ -40,40 +40,40 @@ String doGetProjectTarget() {
...
+    def highestBuildToolsVersion = buildToolsDirContents
+        .collect { new Version(it) }
+        // Invalid inputs will be handled as 0.0.0
+        .findAll { it.isHigherThan('0.0.0') }
+        .max()
+
+    if (highestBuildToolsVersion == null) {
         throw new RuntimeException(
             "No installed build tools found. Install the Android build tools version " +
             minBuildToolsVersionString + " or higher.")
```
erisu: Even though this like is not touched, same could apply here as described below.

framework/cordova.gradle
raphinesse: I pushed a commit that uses string interpolation for the error messages while still avoiding line lengths of > 80 characters. What do you think of it?

PR #945, opened 2020-04-06
Add fix to gradle out of memory issue

brramos:
- Discovered by @Captain-Ahax

### Platforms affected
android

### Motivation and Context
TeamCity reported out of memory errors from gradle

### Description
Doubled memory in gradle.properties

### Testing
Executed npm run test

### Checklist

- [ √] I've run the tests to see all new and existing tests pass
- [ √] I added automated test coverage as appropriate for this change
- [ √ ] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [ √ ] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [ √ ] I've updated the documentation if necessary

PR #948, opened 2020-04-11
Feature: JVM Args flag

breautek:
### Platforms affected
Android

### Motivation and Context

Provides a `--jvmargs` flag to pass through JVM args to gradle. This is mostly useful for changing the `-Xmx` flag, allowing the JVM to use more memory on larger projects.

There was no feature request made, but it does closes https://github.com/apache/cordova-android/pull/945 ; a PR that changes the default memory size to 4096m.

I believe this is a better alternative as I believe it is better to keep the `-Xmx` as low as possible and for most users, our default value of 2048m is probably fine, but allows the flexibility for users to change the value based on their own unique needs. Additionally, this accepts any [JVM args](https://docs.gradle.org/current/userguide/build_environment.html#sec:configuring_jvm_memory), so not necessary limited to `-Xmx` flag.

The flag works on `prepare`, `build`, and `run` cordova commands. Example is given below:

`cordova prepare android -- --jvmargs='-Xmx4g'`

Do note the double set of dashes. Value memory size units are `k`, `m`, and `g` and is case-insensitive. This matches what JVM supports.

### Description

This PR looks at the option vargs in `prepare` and sets the appropriate key for `GradlePropertiesParser` for jvm args.

Some changes were made to the `Api` class, which was required to get unit tests configured properly. (The entire `prepare` function was not unit tested at all). Unit tests changes are made in a separate commit.

Lastly, there is an informational log that was printed to the console if the user changes the `-Xmx` option to something other than the cordova default. I've changed this to only produce that log if the numerical size value is smaller than the cordova default. Code was added to parse the byte value for comparison. Unit tests were also added. This particular part is made in a separate commit from the jvmargs flag implementation.

This PR is accompanied by a `cordova-docs` PR: https://github.com/apache/cordova-docs/pull/1078

### Testing

`npm test` was ran with all existing and new tests passing.
I have also manually tested these changes in a dummy project, ensuring that the `gradle.properties` file produced expected results both with and without the flag.

### Checklist

- [] I've updated the documentation if necessary

### DIFF
--- a/bin/templates/cordova/lib/config/GradlePropertiesParser.js
+++ b/bin/templates/cordova/lib/config/GradlePropertiesParser.js
@@ -90,8 +90,15 @@ class GradlePropertiesParser {
                 this.gradleFile.set(key, properties[key]);
             } else if (value !== properties[key]) {
                 if (this._defaults[key] && this._defaults[key] !== properties[key]) {
-                    // Since the value does not match default, we will notify the discrepancy with Cordova's recommended value.
-                    events.emit('info', `[Gradle Properties] Detected Gradle property "${key}" with the value of "${properties[key]}", Cordova's recommended value is "${this._defaults[key]}"`);
+                    let shouldEmit = true;
+                    if (key === 'org.gradle.jvmargs') {
+                        shouldEmit = this._isJVMMemoryLessThanRecommended(properties[key], this._defaults[key]);
+                    }
+
+                    if (shouldEmit) {
+                        // Since the value does not match default, we will notify the discrepancy with Cordova's recommended value.
+                        events.emit('info', `[Gradle Properties] Detected Gradle property "${key}" with the value of "${properties[key]}", Cordova's recommended value is "${this._defaults[key]}"`);
+                    }
                 } else {
                     // When the current value exists but does not match the new value or does matches the default key value, the new value it set.
                     events.emit('verbose', `[Gradle Properties] Updating Gradle property "${key}" with the value of "${properties[key]}"`);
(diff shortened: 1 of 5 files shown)


breautek:
After looking at http://localhost:3000/docs/en/dev/guide/platforms/android/index.html#setting-gradle-properties

I'm wondering if I should rename the `jvmargs` flag to `cdvJvmArgs` to keep things consistent with the other flags we expose.

bin/templates/cordova/lib/config/GradlePropertiesParser.js
```diff
@@ -103,6 +110,34 @@ class GradlePropertiesParser {
...
+        const MEGABYTE = 1048576;
+        const GIGABYTE = 1073741824;
+
+        switch (unit.toLowerCase()) {
+        case 'k': return size * KILOBYTE;
+        case 'm': return size * MEGABYTE;
+        case 'g': return size * GIGABYTE;
+        }
+
+        events.emit('warn', '[Gradle Properties] Unknown memory size unit (' + unit + ')');
```
timbru31, line 137: I'd prefer a template string instead.

breautek:
Hi @timbru31 

> Should we move the KILOBYTE, MEGABYTE and GIGABYTE constants to something more global

We can, and will likely be more appropriate. Where would you suggest? Something like this I would think would be best suited in `cordova-common` in my opinion, but if that's the route we take, I'd prefer to add it to `cordova-common` and eventually refactor `cordova-android` to use it instead of waiting for a `cordova-common` release (unless if we can get that quick enough...)

> The recommendedCapture and recommendedBase are (or at least should) be the same, too - can we move this to something more global, too?

This uses a regex that specifically parses the jvm arg `-Xmx` flag, so I'm not sure how useful it would be in a global setting. `_getBaseJVMSize` could be something that moved to something more global (with a function rename of course) as it's responsible for converting a memory value to bytes. (e.g. 1k to 1024 bytes).

timbru31:
> We can, and will likely be more appropriate. Where would you suggest? Something like this I would think would be best suited in `cordova-common` in my opinion, but if that's the route we take, I'd prefer to add it to `cordova-common` and eventually refactor `cordova-android` to use it instead of waiting for a `cordova-common` release (unless if we can get that quick enough...)

+1

> The recommendedCapture and recommendedBase are (or at least should) be the same, too - can we move this to something more global, too?

Convinced :)

PR #950, opened 2020-04-12
fix: Removed redundent logcat print

### LINKED ISSUE #914: Cleaner Logging for Android Studio
# Feature Request
Only print a console.log message once to the Android Studio Logcat

## Motivation Behind Feature
Cleaner logs allow for easier searching (no need to divide by 2 to figure out how many times something has printed), and easier readability while it's running.

## Feature Description
I've been mildly annoyed by each console.log message being duplicated in the logs displayed by Logcat and Run tabs in Android Studio for a while. Today, I noticed it's coming from from SystemWebChromeClient.java, onConsoleMessage(). I found that I can enable just one or the other by either only return the super call, or modified to return true instead of the super call. I was hoping it would optionally be controlled by a config item, assuming there's customers for each type of log message. Or maybe no one needs both messages and it can be updated to only log one or the other.

I've heard of some people using a custom regex filter to get to cleaner logs, or using Chrome's debugger.

breautek:
### Platforms affected
Android

### Motivation and Context

The chromium webview will print an informational log already,
we don't need to override the method to do what the webview will already.
Closes #914 

### Description

Simply removes the `onConsoleMessage` message override, since we don't do anything other than to print a logcat message out, which the super class already does, resulting in duplicate logcat prints.

### Testing

Ran npm test and ensured all existing tests passes.

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
+++ b/framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
@@ -151,14 +151,6 @@ public class SystemWebChromeClient extends WebChromeClient {
     }
 
     @Override
-    public boolean onConsoleMessage(ConsoleMessage consoleMessage)
-    {
-        if (consoleMessage.message() != null)
-            LOG.d(LOG_TAG, "%s: Line %d : %s" , consoleMessage.sourceId() , consoleMessage.lineNumber(), consoleMessage.message());
-        return super.onConsoleMessage(consoleMessage);
-    }
-
-    @Override
     /**
      * Instructs the client to show a prompt to ask the user to set the Geolocation permission state for the specified origin.
      *


stochmalm:
After upgrade to cordova android 9.0.0 I'm unable to see any logs from `console.log`. Isn't this related to this fix? Upgrading cordova was the only thing that I have done and now I must use alert() for debugging ._.

breautek:
> After upgrade to cordova android 9.0.0 I'm unable to see any logs from `console.log`. Isn't this related to this fix? Upgrading cordova was the only thing that I have done and now I must use alert() for debugging ._.
> 
> Tested on Android 10, Xiaomi MI 10 SE.

`return super.onConsoleMessage(consoleMessage);` would have produced console log messages similar to what Cordova did in the override. But it may be in a different format and definitely under a different log tag.

> now I must use alert() for debugging ._.

Using logcat is mainly used for native debugging. Debugging cordova apps you should use the web inspector tools provided by Chrome (for android), or Safari for ios apps. See [Remote Debugging webviews](https://developers.google.com/web/tools/chrome-devtools/remote-debugging/webviews) for more information how to do this for the Chrome webview.

lomchik:
Hi. This commit breaks logging in release mode which is very useful for bug reports. 

Reason:

webView print logs only when `WebView.setWebContentsDebuggingEnabled(true)`. It is enabled [here](https://github.com/apache/cordova-android/blob/80b7a7f6acb43dfc2dfd4fd87261052496f37756/framework/src/org/apache/cordova/engine/SystemWebViewEngine.java#L176-L178) 


I propose to add instead of this method something like this.
```
    @Override
    public boolean onConsoleMessage(ConsoleMessage consoleMessage)
    {
        if (!ApplicationInfo.FLAG_DEBUGGABLE && consoleMessage.message() != null)
            LOG.d(LOG_TAG, "%s: Line %d : %s" , consoleMessage.sourceId() , consoleMessage.lineNumber(), consoleMessage.message());
        return super.onConsoleMessage(consoleMessage);
    }
```

stochmalm:
> Using logcat is mainly used for native debugging. Debugging cordova apps you should use the web inspector tools provided by Chrome (for android), or Safari for ios apps. It provides much more robust console API than logcat, + you get your JS breakpoints and other debugging tools. See [Remote Debugging webviews](https://developers.google.com/web/tools/chrome-devtools/remote-debugging/webviews) for more information how to do this for the Chrome webview.

I had an error only on release build for android in which remote debugging is not available ;). Finally I reverted crodova-android back to 8.1.0 and now logging works properly.

breautek:
@lomchik 

I see. I'd support a PR that re-adds console logging if the app is in release mode and if the app is in debug mode, then fall back to the super.call.

PR #962, opened 2020-04-25
breaking: Bump android gradle plugin to 3.6.0

breautek:
### Platforms affected
Android

### Motivation and Context

Upgrades android gradle plugin to 3.6.0.

afaik there are no issues that are dependant on this change, but this is usually considered a breaking change, so it would be great to be included in the 9.0.0 release.

### Description

Updated the framework and the two android test projects version pins.

### Testing

ran `npm test`

### Checklist

### DIFF
--- a/framework/build.gradle
+++ b/framework/build.gradle
@@ -31,7 +31,7 @@ buildscript {
     dependencies {
         // The gradle plugin and the maven plugin have to be updated after each version of Android
         // studio comes out
-        classpath 'com.android.tools.build:gradle:3.5.3'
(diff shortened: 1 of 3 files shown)


timbru31:
Release Notes are over here: https://developer.android.com/studio/releases/gradle-plugin#3-6-0

erisu:
`3.6.3` I believe is actually the latest release from April.

breautek:
I used https://developer.android.com/studio/releases/gradle-plugin to determine the latest, they don't mention 3.6.3 O.o

Issue #963, opened 2020-04-29
Android detectArchitecture timeout is too short


GimpArm:
# Bug Report

### What is expected to happen?

Android debugging to connect

### What does actually happen?

Android debugging fails to connect

## Information

The timeout for detechArchitecture() is sometimes too low when devices are on wifi network connections and even sometimes over USB. The command can take up to 3 seconds to execute and return. Currently the timeout is set to 1000 ms and setting it to 5000 ms seems to be a good compromise.

[/bin/templates/cordova/lib/build.js line213](https://github.com/apache/cordova-android/blob/master/bin/templates/cordova/lib/build.js#L213)

`Command finished with error code 0: adb devices
Running adb shell command "cat /proc/cpuinfo" on target ce071827f219c13701...

Command finished with error code ENOENT: killall adb
Device communication timed out. Try unplugging & replugging the device.
CordovaError: Device communication timed out. Try unplugging & replugging the device.
    at Object.module.exports.detectArchitecture (D:\workspace\Git\store-logistics-app\build-tmp\cordova\cordova\platforms\android\cordova\lib\build.js:209:35)
    at D:\workspace\Git\store-logistics-app\build-tmp\cordova\cordova\platforms\android\cordova\lib\device.js:64:22
    at _fulfilled (D:\workspace\Git\store-logistics-app\node_modules\q\q.js:854:54)
    at D:\workspace\Git\store-logistics-app\node_modules\q\q.js:883:30
    at Promise.promise.promiseDispatch (D:\workspace\Git\store-logistics-app\node_modules\q\q.js:816:13)
    at D:\workspace\Git\store-logistics-app\node_modules\q\q.js:624:44
    at runSingle (D:\workspace\Git\store-logistics-app\node_modules\q\q.js:137:13)
    at flush (D:\workspace\Git\store-logistics-app\node_modules\q\q.js:125:13)
    at processTicksAndRejections (internal/process/task_queues.js:79:11)
Command finished with error code 0: adb -s,ce071827f219c13701,shell,cat,/proc/cpuinfo
'cordova.cmd run android --device --verbose --no-update-notifier' failed with exit code 1`

### Command or Code

`cordova.cmd run android --device --verbose --no-update-notifier`

### Environment, Platform, Device

Windows 10 building Android

### Version information

VS Code: 1.44.2
Cordova: 9.0.0
Cordova-Android: 8.1.0

## Checklist
- [x] I searched for existing GitHub issues
[...]

PR #965, opened 2020-04-29
Increased detectArchitecture() timeout

GimpArm:
The timeout for detectArchitecture() is sometimes too low when devices are on wifi network connections and even sometimes over USB. The command can take up to 3 seconds to execute and return. Currently the timeout is set to 1000 ms and setting it to 5000 ms seems to be a good compromise.

### Platforms affected

Android

### Motivation and Context

https://github.com/apache/cordova-android/issues/963

### Description

Increased the detectArchitecture() method timeout from 1000 ms to 5000 ms

### Testing

Changing the timeout does not affect and unit tests.

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/build.js
+++ b/bin/templates/cordova/lib/build.js
@@ -210,7 +210,7 @@ module.exports.detectArchitecture = function (target) {
     // To fix it, either unplug & replug device, or restart adb server.
     return Promise.race([
         helper(),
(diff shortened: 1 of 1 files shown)


breautek:
Thank you for your contribution. I'd like to see one other PMC member vote positive before I merge.

breautek:
Thanks Tim!

PR #975, opened 2020-05-15
ci: Added Node 14.x

breautek:
Adds Node 14.x to the CI matrix, which is the new current and will be the LTS later this year.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.x, 12.x]
+        node-version: [10.x, 12.x, 14.x]
         os: [ubuntu-latest, windows-latest, macos-latest]
 
     steps:


timbru31:
This is the first addition of a non LTS to the testing matrix (if I'm correct).

Do we want to test those versions, too? Only the new, soon to be, LTS ones? I think we should discuss that :)

erisu:
I think it is OK to include 14.x since it is a stable release version. But only as part of the test environment to begin with. No public announcement needed.

Officially, I believe we only recommend LTS. We might sometimes mention stable release is acceptable but with caution as there could be an issue due to the lack of testing on our end.

The benefit of adding it now is to try and catch the issues quickly and fix them before node reaches LTS. Which is always recommended to do. (Note: never the odd releases of node)

I would prefer we plan on adding node 14 after the major releases is pushed out and only to the CIs.

Currently I been only focused on getting node 12 out because it was already planned and announced. 

* https://github.com/apache/cordova/issues/79
* [Node 6.x and 8.x Deprecation Timeline for Apache Cordova](https://cordova.apache.org/news/2019/04/11/nodejs-6.x-8.x-deprecation-timeline.html)

We also already released a few repos with just node 12.

breautek:
I'd support waiting until all the planned out releases are actually released. Then if issues arise afterwards, we can patch fixes in.

I'd even support waiting until NodeJS 14 actually becomes in LTS state. While in "current" I think new features and rapid changes are still be applied, creating an uncertain environment. Node 14 LTS starts October 20, 2020.

erisu:
Pinning Track Issue: https://github.com/apache/cordova/issues/223

PR #988, opened 2020-06-05
[major] feat: Gradle 6.5 & Android Gradle plugin 4.0.0 updates

brodybits:
### Platforms affected

Android

### Motivation and Context

- Android Studio 4.0 recommended updating to multiple `build.gradle`
- Gradle 6.1.1 update seems to be required to build with Android Gradle plugin 4.0.0 update, may as well go with Gradle 6.5 update

I think it would be ideal if we could make this update before shipping the new major 9.0.0 release.

### Description

- update to use Gradle 6.5
- update the following `build.gradle` files as suggested by Android Studio 4.0:
  - `bin/templates/project/app/build.gradle`
  - `bin/templates/project/build.gradle`
  - `framework/build.gradle`

Note that this proposal would effectively overwrite some or all changes from the following PRs:

- #962
- #792

/cc @erisu @breautek

### Testing

- tested with command-line build from Cordova CLI
- tested with Android Studio 4.0

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/builders/ProjectBuilder.js
+++ b/bin/templates/cordova/lib/builders/ProjectBuilder.js
@@ -303,7 +303,7 @@ class ProjectBuilder {
                 // update/set the distributionUrl in the gradle-wrapper.properties
                 const gradleWrapperPropertiesPath = path.join(self.root, 'gradle/wrapper/gradle-wrapper.properties');
                 const gradleWrapperProperties = createEditor(gradleWrapperPropertiesPath);
-                const distributionUrl = process.env.CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL || 'https://services.gradle.org/distributions/gradle-6.1-all.zip';
+                const distributionUrl = process.env.CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL || 'https://services.gradle.org/distributions/gradle-6.5-all.zip';
                 gradleWrapperProperties.set('distributionUrl', distributionUrl);
                 gradleWrapperProperties.save();
 
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -53,7 +53,7 @@ buildscript {
     dependencies {
         apply from: '../CordovaLib/cordova.gradle'
 
-        classpath 'com.android.tools.build:gradle:3.5.3'
+        classpath 'com.android.tools.build:gradle:4.0.0'
 
         if (cdvHelpers.getConfigPreference('GradlePluginKotlinEnabled', 'false').toBoolean()) {
             classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -89,7 +89,7 @@ allprojects {
 }
 
 task wrapper(type: Wrapper) {
-    gradleVersion = '6.1'
+    gradleVersion = '6.5'
 }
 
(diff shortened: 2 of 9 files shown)


breautek:
Should also update the two gradle wrapper files in the tests at:

https://github.com/apache/cordova-android/blob/master/test/android/wrapper.gradle
https://github.com/apache/cordova-android/blob/master/test/androidx/wrapper.gradle

mosabab:
> Should also update the two gradle wrapper files in the tests at:
> 
> https://github.com/apache/cordova-android/blob/master/test/android/wrapper.gradle
> https://github.com/apache/cordova-android/blob/master/test/androidx/wrapper.gradle
> 
> Also their respective `build.gradle` files for the android gradle plugin, which currently is still at version `3.6.0`

@breautek 
Could you please check this message when I build my app?

Untitled-2

Does this message normal ?
Because I can see some warnings for Gradle .....etc
The build success but i need to know if this normal or there may be some issues?

To generate the previous message, in command CMD on windows I use the following steps:
1- cordova create test com.test TestApp
2- cd test (I added all my files after this step including my config.xml ...etc)
3- cordova platform add https://github.com/apache/cordova-android.git#master
4- cordova build android --release

Please note that i am using:
The latest cordova cli v 9.0.0
Gradle 6.5
Android Studio Latest Version v4.0 with updates.
cordova-android.git#master version.
JDK 8 (jdk1.8.0_251).
----
Thanks
Regards

breautek:
Yes, that's normal. Gradle always appear to have deprecation warnings, but if you use gradle directly with the stacktrace flag, then you'll see that the faulting code comes from gradle itself.

mosabab:
> Yes, that's normal. Gradle always appear to have deprecation warnings, but if you use gradle directly with the stacktrace flag, then you'll see that the faulting code comes from gradle itself.

Thanks for the clarification

brodybits:
> Should also update the two gradle wrapper files in the tests at:
> 
> https://github.com/apache/cordova-android/blob/master/test/android/wrapper.gradle
> https://github.com/apache/cordova-android/blob/master/test/androidx/wrapper.gradle

I was actually wondering why keep updating those 2 files while `bin/templates/project/wrapper.gradle` is kept blank with a comment.

> Also their respective `build.gradle` files for the android gradle plugin, which currently is still at version `3.6.0`

The following spec & test `build.gradle` files also seemed to be out-of-date with respect to the `build.gradle` files that I had updated.

From a search I did also find `bin/templates/project/legacy/build.gradle`, wonder if we really want to keep supporting it or not.

I would definitely be happy to update the test Gradle files if needed, just wanted to understand what we should be keeping up-to-date and if there is anything we should not be supporting any longer.

breautek:
The `project/legacy` files is from beyond my time, I'm not sure the purpose, but it doesn't appear to be used.

> I would definitely be happy to update the test Gradle files if needed, just wanted to understand what we should be keeping up-to-date and if there is anything we should not be supporting any longer.

The test gradle files should be kept in sync, otherwise the unit tests won't be ran under the same configuration as the runtime when the project is installed.

brodybits:
I tried copying the updated Gradle files, now leads to a build error. Any ideas?

brodybits:
I have updated the test Gradle files as requested. There seems to be quite a major difference between the template Gradle files and test Gradle files, and I hope this can be fixed someday.

brodybits:
- build is still green
- tested with my `cordova-sqlite-storage` test app with build & run from Cordova CLI
- tested adding to my `cordova-sqlite-storage` test app but with build & run from Android Studio 4.0

Merging now, thanks to @breautek for the quick reviews!

PR #997, opened 2020-06-10
cleanup: remove trailing spaces from project template

brodybits:
### Platforms affected

Android

### Motivation and Context

see title

### Description

- remove trailing spaces from bin/templates/project/AndroidManifest.xml
- remove trailing spaces from bin/templates/project/assets/www/css/index.css

**open question:** I was wondering if we should really keep `www` assets committed in cordova-android or any other repos? Shouldn't we just use the `www` assets from the app-hello-world template?

### Testing

see below

### Checklist

- [ ] I've run the tests to see all new and existing tests pass - check GitHub CI
- ~~I added automated test coverage as appropriate for this change~~

### DIFF
--- a/bin/templates/project/AndroidManifest.xml
+++ b/bin/templates/project/AndroidManifest.xml
@@ -17,7 +17,7 @@
        specific language governing permissions and limitations
        under the License.
 -->
-<manifest xmlns:android="http://schemas.android.com/apk/res/android" 
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
   package="__PACKAGE__" android:versionName="1.0" android:versionCode="1" android:hardwareAccelerated="true">
     <supports-screens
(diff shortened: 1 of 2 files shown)


dpogue:
> **open question:** I was wondering if we should really keep `www` assets committed in cordova-android or any other repos? Shouldn't we just use the `www` assets from the app-hello-world template?

I believe these are required for the platform-centred workflow where users download cordova-android and then use the `bin/create` script to create new projects as opposed to using the cordova CLI.

brodybits:
> I believe these are required for the platform-centred workflow

Makes sense to me. I wonder if we should consider doing anything to keep these in sync, like we do with cordova.js.

I definitely recall discussions of removing platform-centered workflow. I would personally favor keeping it whenever possible, don't really like the added support burden though. I wonder if we could find some way to automate this.

PR #999, opened 2020-06-10
cleanup: remove trailing spaces from Java sources

brodybits:
### Platforms affected

Android

### Motivation and Context

see title

### Description

see title

### Testing

✅ `npm test` in my work area - OK
✅ tried using in `cordova-sqlite-storage` test app - OK

### Checklist

- [x] I've run the tests to see all new and existing tests pass - both in work area and in GitHub CI
- ~~I added automated test coverage as appropriate for this change~~


brodybits:
In general, I think we should do this cleanup through all of our repositories and look for a way to avoid trailing whitespaces in the future. I learned about `git grep` pretty recently, looks like a good friend, and discovered this for finding tabs: https://stackoverflow.com/questions/4262150/is-it-possible-to-git-grep-for-tabs

breautek:
Could even look at implementing java/android lint

https://developer.android.com/studio/write/lint

PR #1013, opened 2020-07-02
test(pluginHandlers/common): better setup & teardown

raphinesse:
### Motivation and Context

- test file setup was overly verbose
- teardown was faulty (exceptions in tests could lead to files being left behind)
- I wanted to add a regression test to the suite

### Description

- use `fs.outputFileSync` to create files and their parent dirs
- use `afterEach` to reliably clean up after tests

### Testing

Affected unit tests still pass

### DIFF
--- a/spec/unit/pluginHandlers/common.spec.js
+++ b/spec/unit/pluginHandlers/common.spec.js
@@ -21,6 +21,7 @@ var common = rewire('../../../bin/templates/cordova/lib/pluginHandlers');
 var path = require('path');
 var fs = require('fs-extra');
 var osenv = require('os');
+
 var test_dir = path.join(osenv.tmpdir(), 'test_plugman');
 var project_dir = path.join(test_dir, 'project');
 var src = path.join(project_dir, 'src');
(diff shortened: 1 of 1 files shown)


spec/unit/pluginHandlers/common.spec.js
```diff
@@ -35,38 +36,39 @@ var deleteJava = common.__get__('deleteJava');
...
-            fs.writeFileSync(java_file, 'contents', 'utf-8');
+            fs.outputFileSync(java_file, 'contents');
 
             // This will fail on windows if not admin - ignore the error in that case.
             if (ignoreEPERMonWin32(java_file, symlink_file)) {
                 return;
             }
 
             copyFile(test_dir, symlink_file, project_dir, dest);
-            fs.removeSync(project_dir);
```
timbru31, line 64: Won't this leave traces of the test behind? It's not cleared via the afterEach hook.

raphinesse: No. `project_dir` is contained in `test_dir` which is removed in `afterEach`.

raphinesse:
Thanks for the review, @timbru31

PR #1015, opened 2020-07-02
Remove unnecessary stuff

raphinesse:
### Motivation and Context

Keep Cordova maintainable.

### Description

See individual commits.

AFAICT no public interfaces are affected.

### Testing

- Automated tests pass
- Verified that removed stuff was not used or needed


LICENSE
```diff
@@ -200,68 +200,3 @@
...
-THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
-WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
-MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
-ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
-WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
-ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR
-IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
-
-================================================================================
-bin/node_modules/which
```
breautek, line 250: `which` is still used I believe...

bin/clean_test.js
```diff
@@ -1,28 +0,0 @@
...
-       software distributed under the License is distributed on an
-       "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
-       KIND, either express or implied.  See the License for the
-       specific language governing permissions and limitations
-       under the License.
-*/
-
-const fs = require('fs');
-
-if (fs.existsSync('./test/gradlew')) {
```
breautek, line 22: This was added to easily clean gradle wrappers inside the tests, although I think the paths needs to be updated to `test/android` and `test/androidx`.

I think I'd prefer to keep this script and update the paths. The intent was to clean up gradle wrappers so that when you upgrade the gradle version, the new version will be installed. If gradle is already present, it won't install the newer declared version.

LICENSE
erisu: We do not need to maintain these licenses in our LICENSE file since we do not bundle these node modules anymore.
These licenses were referencing modules that we use to bundle in the `bin` dir, but has been removed long ago.

bin/clean_test.js
raphinesse: I wanted to mention `git clean`, but I saw that this was already discussed when this was added. I will remove this file's deletion from this PR.

PR #1022, opened 2020-07-07
WIP: fix: password prompt

### LINKED ISSUE #1021: Keystore Password Prompt is broken
# Bug Report

## Problem

> For release signing, passwords can be excluded and the build system will issue a prompt asking for the password.

The `build.json` password fields when left empty should cause gradle to prompt for the password during build, but an error is thrown instead.

Password prompts to be shown.

An error is thrown:

```
FAILURE: Build failed with an exception.

* Where:
Script '/development/cordova/tests/tp-iri-client/platforms/android/CordovaLib/cordova.gradle' line: 141

* What went wrong:
Failed to create component for 'dialog' reason: java.awt.HeadlessException
> java.awt.HeadlessException (no error message)

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 12s
```

## Information

I've debugged this problem and it's because when gradle daemon is enabled, there is no "console", therefore you cannot use a headless (terminal) password prompt.

So to use password prompts, we must turn off headless mode, which will enable GUI password prompt when invoked.

First brought attention from https://github.com/apache/cordova-android/issues/1020#issuecomment-654524782

I've already debugged the problem and a PR is incoming shortly...

### Command or Code

`cordova build android --release`

While having a `buiild.json` file, with `password` and `storePassword` as empty strings.

Issue was replicated on both Ubuntu & Windows

### Version information

Cordova CLI 9
cordova-android 9

## Checklist

breautek:
### Platforms affected

Android

### Motivation and Context

Password prompt when building release builds is broken and crashes with an exception.
Fixes #1021 

### Description

Adds `System.setProperty("java.awt.headless", "false")` just before building the Swing UI, and resets the property back to true afterwards.

This is required when gradle is set in daemon mode. This [blog](https://www.timroes.de/using-password-prompts-with-gradle-build-files#problem-2--we-dont-have-a-console) explains why.

### Testing

Manual testing and `npm test`

### Checklist

### DIFF
--- a/framework/cordova.gradle
+++ b/framework/cordova.gradle
@@ -134,6 +134,7 @@ def doExtractStringFromManifest(name) {
 def doPromptForPassword(msg) {
     if (System.console() == null) {
         def ret = null
+        System.setProperty('java.awt.headless', 'false')
         new SwingBuilder().edt {
             dialog(modal: true, title: 'Enter password', alwaysOnTop: true, resizable: false, locationRelativeTo: null, pack: true, show: true) {
                 vbox {
@@ -146,6 +147,7 @@ def doPromptForPassword(msg) {
                 }
             }
         }
+        System.setProperty('java.awt.headless', 'true')
         if (!ret) {
             throw new GradleException('User canceled build')
         }


breautek:
This fixes the prompt issue itself, but there seems to be another issue (described at #1023) where the password is always rejected, but only if the password was provided through the prompt.

I'm not sure if this is still an issue with the prompt itself, or if it's something else in the build script.

breautek:
Closing because it looks like the new direction is to remove this feature altogether.

https://lists.apache.org/thread.html/rfd8f4c44acf3fc278dda409d2e677ce24ec3a65e34f5b7db5202d5eb%40%3Cdev.cordova.apache.org%3E

PR #595, opened 2018-12-02
Add a warning about the using of `android.bat`

Songbird0:
Hi,

### Platforms affected

Android.

Like it said:

```
**************************************************************************
The "android" command is deprecated.
For manual SDK, AVD, and project management, please use Android Studio.
For command-line tools, use tools\bin\sdkmanager.bat
and tools\bin\avdmanager.bat
**************************************************************************
```

### What does this PR do?

The conditionnal structure has been removed to use systematically `sdkmanager` instead of `android.bat`.

See [this issue](https://github.com/apache/cordova/issues/61) for more.

### What testing has been done on this change?

None. The method behaviour hasn't been significantly altered, actually.

Feel free to add/remove what you want. :)

### Checklist
- [ ] [Reported an issue](http://cordova.apache.org/contribute/issues.html) in the JIRA database
- [ ] Commit message follows the format: "CB-3232: (android) Fix bug with resolving file paths", where CB-xxxx is the JIRA ID & "android" is the platform affected.
- [ ] Added automated test coverage as appropriate for this change.

Issue #741, opened 2018-12-01
Warning to fix: android.bat shouldn't be recommended anymore


Songbird0:
Hi,

When `cordova requirements` command is typed and any Android target hasn't been assigned, cordova recommends to use `android.bat` to install a fresh Android target. However, sdkmanager.bat should be used instead of android.bat.

I would be very happy to fix it myself, but I don't know the location of this warning.

erisu:
Closing this issue as PR #1083 mentions that it closes this ticket.

PR #1083, opened 2020-10-05
refactor!: drop support for `android` SDK tool

raphinesse:
### Motivation and Context

We have multiple places where we support both the legacy `android` SDK tool as well as [`avdmanager`](https://developer.android.com/studio/command-line/avdmanager) and [`sdkmanager`](https://developer.android.com/studio/command-line/sdkmanager) from current SDK tools. This makes code more complicated and maintenance more difficult.

This PR removes the support for the `android` CLI tool from our code base. After this change, to use `cordova-android` users will need either:
- Revision 25.3.0 (March 2017) or greater of the [Android SDK Tools](https://developer.android.com/studio/releases/sdk-tools)
- any version of the [new command-line tools](https://developer.android.com/studio#cmdline-tools) which are also included in Android Studio

Closes #595 #741

### Description

- `emulator.list_images` now always uses the `avdmanager` binary.
- `android_sdk.list_targets` now always uses the `avdmanager` binary.
- `check_reqs.check_android` does not consider legacy `android` binary for SDK detection anymore
- We do not give error messages that involve using the legacy `android` binary anymore

There are multiple opportunities for further simplification after the code removal in this PR. I already have seized some of these in a local branch but I would prefer to open new PRs for these after this is merged.

### Testing

- Existing unit and E2E tests pass
- The following manual tests give the expected results on my machine:
  - `node -e "require('./bin/templates/cordova/lib/emulator').list_images().then(console.log)"`
  - `node -e "require('./bin/templates/cordova/lib/android_sdk').list_targets().then(console.log)"`
  - `node -e "require('./bin/templates/cordova/lib/check_reqs').check_android().then(console.log)"`


purplecabbage (approved):
Looks good to me!

breautek (approved):
lgtm

Tested adding the platform and building/running on emulator using one of my apps.

erisu (approved):

raphinesse:
Thanks for the review everyone! Since this is a breaking change for next major, how do we want to continue with this PR?

- merge it into master now
- leave it sitting open until we're confident not to release patches/minors for 9 anymore
- integrate it into a long-running `next-major` branch where we collect breaking changes for next major

Side note: I will probably need some integration branch for multiple breaking changes anyway

breautek:
> integrate it into a long-running next-major branch where we collect breaking changes for next major

This might be a good idea for testing PRs against the collection of other breaking changes to be released.

PR #1082, opened 2020-10-05
refactor(lib/run): simplify code to determine install/run target

raphinesse:
### Motivation and Context

This change simplifies the code that determines on which target (device or emulator) the app should be installed and run on.

This refactoring was triggered during a bigger refactoring (still WIP) that has the goal of doing something similar to https://github.com/apache/cordova-electron/pull/142 here.

### Description

The changes made here should not change the observable behavior of the code beyond changing what is logged to the console. The changes included are:

- factor out `resolveInstallTarget` from `lib/run.run`
- improve unit tests for the `lib/run` module
- convert `resolveInstallTarget` to async/await
- DRY logic in `resolveInstallTarget`

The logic for selecting a install/run target is unchanged. Targets are tried to resolve in this order:

- A connected device matching the given ID (if target ID given)
- Any connected device (if `--emulator` option was not given)
- A started emulator matching the given ID (if target ID given)
- An existing emulator matching the given ID (will be started) (if target ID given)
- An existing emulator that best matches the app's target API (will be started) (if `--device` option was not given)

### Testing

The unit tests for this module still pass.

PR #1095, opened 2020-10-17
fix(device.list): do not mess with following `adb devices` runs

raphinesse:
If the `lookHarder` option is true and the initial `adb devices` run
returned no devices, `device.list` will kill the adb server and try
again.

Unfortunately `adb devices` contains incorrect information when it is
run too soon after the adb server has been restarted. Consider below
commands being executed with one running emulator:

    adb devices && killall adb && adb devices && adb devices

While the first invocation of `adb devices` lists the emulator as
running, the second and third one will list it as offline. This is
especially problematic if code contains a sequence such as:

    devs = await device.list(true);
    emus = await emulator.list();

In that case the third invocation of `adb devices` occurs in
`emulator.list()` but because of the `killall` it lists all emulators as
offline, causing `emus` to be empty.

This commit tries to work around that issue by explicitly restarting the
adb server after it was killed, and waiting for 100 ms after that. Local
experimentation has shown this to be sufficient in all instances. YMMV
of course.

Fixes the problem found by @breautek when reviewing #1082

### DIFF
--- a/bin/templates/cordova/lib/device.js
+++ b/bin/templates/cordova/lib/device.js
@@ -25,6 +25,8 @@ var AndroidManifest = require('./AndroidManifest');
 var CordovaError = require('cordova-common').CordovaError;
 var events = require('cordova-common').events;
 
+const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
+
 /**
  * Returns a promise for the list of the device ID's found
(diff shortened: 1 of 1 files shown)


raphinesse:
I will probably fix this as part of a bigger refactoring

raphinesse:
Closing in favor of #1098

PR #1100, opened 2020-10-19
refactor!: remove most platform binaries

raphinesse:
### Motivation and Context

This removes all platform binaries that are not required for compatibility to Cordova CLI.

My work to stop copying `cordova/lib` to the platform folder is based on this PR among others.

### Details
The binaries that remain (relative to a platform project) are:
- `cordova/version` (used by `cordova-lib` to determine the platform version)
- `cordova/android_sdk_version` (used to implement the [`<engine>` requirement](https://cordova.apache.org/docs/en/latest/plugin_ref/spec.html#engines-and-engine) that can be specified in `plugin.xml`)
- `cordova/lib/list-devices` (for [`cordova targets`][1])
- `cordova/lib/list-emulator-images`  (for [`cordova targets`][1])

We could remove all the `*.bat` counterparts though, since `execa` (which is used to run the binaries) also implements shebang-lookup on Windows. This will break compatibility to any CLI version <10, since we did not yet use `execa` then.

There are still places where `nopt` is used to parse arguments passed through from CLI, so we cannot remove it.

[1]: https://github.com/apache/cordova-lib/blob/master/src/cordova/targets.js


dpogue (approved):
Looks good to me so far

erisu:
Everything seems fine but had only 1 question:

* `android_sdk_version`, which was removed, seems to be called from [cordova-lib](https://github.com/apache/cordova-lib/blob/master/src/plugman/util/default-engines.js#L47). Can it be confirmed that this code in lib is not called in the CLI workflow?

raphinesse:
> `android_sdk_version`, which was removed, seems to be called from [cordova-lib](https://github.com/apache/cordova-lib/blob/master/src/plugman/util/default-engines.js#L47). Can it be confirmed that this code in lib is not called in the CLI workflow?

Good catch! Unfortunately this _is_ called during plugin installation as part of the check for the [`<engine>` requirement](https://cordova.apache.org/docs/en/latest/plugin_ref/spec.html#engines-and-engine) that can be specified in `plugin.xml`.

We will have to keep that too for now. In the long run we should probably change how lib delegates this task to the platforms.

erisu:
@raphinesse this PR will need to be rebased and the conflicts resolved.

Other then that, were there other changes needed or is this PR ready and could be merged in?

raphinesse:
> @raphinesse this PR will need to be rebased and the conflicts resolved.
> 
> Other then that, were there other changes needed or is this PR ready and could be merged in?

I'm not sure. I think at least the TODOs in the description still have to be applied. Given that the next major release of this platform is supposed to be independent from a tooling release, this would mean keeping anything that would otherwise break compatibility with current tooling.

breautek (approved):

PR #842, opened 2019-10-13
refactor: remove shelljs dependency

breautek:
### Platforms affected
Android

### Motivation and Context

Replaces `shelljs` with favour for native `fs` APIs, or apis from `fs-extra`
Fixes https://github.com/apache/cordova-android/issues/781

### Description

All shelljs code has been removed and replaced. In some cases, `which` package was used as a replacement.

### Testing

`npm test` passes. Manual testing has been done on linux by using `cordova platform add <my git fork>`

Windows has not been tested. Test are pending.
Mac has has also not been tested. I don't have easy access to a mac machine, so assistance here would help me.

### Checklist

- [x] I've run the tests to see all new and existing tests pass
- [ ] I added automated test coverage as appropriate for this change
- [ ] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [x] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [x] I've updated the documentation if necessary

PR #1102, opened 2020-10-20
refactor(check_reqs): cleanup default Java location detection on Windows

raphinesse:
### Motivation and Context

I was thrown off by [the regex introduced in #842](https://github.com/apache/cordova-android/pull/842/files#diff-0d1afa0d5c88bdb18fbc6ab9ef7fc4bbe10f7047d387bb2964932dbb6eb8c4b7R47) to match JDK folders. While it matches what we want, it is somehow misleading ("does the dir start with an arbitrary long sequence of `jdk` strings followed by any single character?").

### Description

- use `fast-glob` to change the implementation back to the more straightforward "give me all directories that start with `jdk`"
- add a unit test for that feature
- remove a useless `if`
- use `ProgramFiles(x86)` env var to cover all bases

### Testing

New unit test passes

### DIFF
--- a/bin/templates/cordova/lib/check_reqs.js
+++ b/bin/templates/cordova/lib/check_reqs.js
@@ -193,9 +193,7 @@ module.exports.check_java = function () {
                 }
 
                 if (jdkDir) {
-                    if (!javacPath) {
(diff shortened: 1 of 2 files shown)


breautek (approved):
lgtm

PR #1103, opened 2020-10-20
refactor: do not kill adb on UNIX-like systems

### LINKED ISSUE #1098: Killing all adb processes messes with following calls to `adb devices`
# Bug Report

## Problem

In certain programs paths we spawn `killall adb` in an attempt to fix a supposedly invalid state of `adb`. For example: when `device.list` is run with the `lookHarder` option set to `true` and the initial `adb devices` run returns no devices, `device.list` will kill the adb server and try again.

Unfortunately, that `killall adb` causes all `adb devices` calls in a certain time frame after it to list all emulators as `offline`. Since we filter out devices with state `offline`, calls to `emulator.list` within that time frame will have wrong results.

Also, if I had an instance of `adb logcat` running, Cordova would kill that too I guess?

With no devices connected and at least one running emulator:

```js
emus1 = await emulator.list();
devs = await device.list(true);
emus2 = await emulator.list();

expect(emus1).toEqual(emus2);
```

```js
emus1 = await emulator.list();
devs = await device.list(true);
emus2 = await emulator.list();

expect(emus1.length).toBeGreaterThan(0);
expect(emus2.length).toBe(0);
```

### Reproduction on the shell
The JS code `await device.list(true); await emulator.list()` runs this sequence of commands:

```bash
adb devices && killall adb && adb devices && adb devices
```

On my machine the second and third call to `adb devices` lists all emulators as offline. The third one would be the one from `emulator.list`.

## Information

This behavior was introduced in https://github.com/apache/cordova-android/commit/66fa12a to fix [CB-8119](https://issues.apache.org/jira/browse/CB-8119).

## Solution
The problem this `killall` is supposed to solve might not even exist anymore. I remember adb losing devices a few years back but it hasn't happened to me in a long time. It would be great to know if someone still encounters these problems today on a regular basis. If not, we could simply put an end to all that killing :wink: 

If there is still a problem with adb stability that we need or want to solve in Cordova, we need to do better. A robust solution would be introducing a global mutex in our Adb module that is locked when running an adb command or restarting the adb server. After restarting the server we would probably have to introduce a delay of around 100 ms (on my machine) before unlocking that mutex.

I really hope we can just drop that stuff...

raphinesse:
### Motivation and Context

Closes #1098

### Description

This basically reverts 66fa12a.

### Testing

Unit tests pass & works for me on Linux

### DIFF
--- a/bin/templates/cordova/lib/build.js
+++ b/bin/templates/cordova/lib/build.js
@@ -24,8 +24,6 @@ var nopt = require('nopt');
 var Adb = require('./Adb');
 
 var events = require('cordova-common').events;
-const execa = require('execa');
-var CordovaError = require('cordova-common').CordovaError;
 var PackageType = require('./PackageType');
 
 module.exports.parseBuildOptions = parseOpts;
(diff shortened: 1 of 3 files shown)


NiklasMerz:
I think my linux system is set up weirdly. I sometimes get an error when running the adb command that I am using a different version and it restarts. But looking at the log outputs in this PR this is somthing different. I can't really tell if this is still useful :man_shrugging: 😄

raphinesse:
> I sometimes get an error when running the adb command that I am using a different version and it restarts.

Which adb command? It sounds like you are talking about app installation. That has not been touched here.

> But looking at the log outputs in this PR this is somthing different. I can't really tell if this is still useful

I don't think I can follow you. Did you observe any change in behavior with the change from this PR?

NiklasMerz:
I did not test this PR locally, yet. Nevermind I was just thinking loud because I always had strange adb behaviour but that is most likely not caused by Cordova.

From the issue:
> The problem this killall is supposed to solve might not even exist anymore. I remember adb losing devices a few years back but it hasn't happened to me in a long time. It would be great to know if someone still encounters these problems today on a regular basis. If not, we could simply put an end to all that killing 😉

I often have trouble with lost connections to my real device on a Linux machine. I feel this is problem with my setup and it's hard to reproduce properly.

This is probably a good change but I am unsure about how to test this👍

raphinesse:
> This is probably a good change but I am unsure about how to test this

Yeah, it is hard to test in how far adb stability is still an issue. I would suggest we remove the current workaround and in case people report frequent hangs, we add a timeout to short-running adb commands.

breautek:
In my experience, at least on Linux, adb tools is a finicky program that barely works. Often I have to unplug the USB device and replug it in and ADB will start magically working again. I'm not sure if this is a problem we can solve on the Cordova side. Obviously https://github.com/apache/cordova-android/commit/66fa12a091c44d062ce4423e9e413e647895c466 was probably a good attempt but I don't think it solved the issue.

In the long term, I envision that we abandoned adb for most, if not all tasks anyway, and use the [bundletool](https://github.com/google/bundletool) instead. If we're lucky it will fix all ADB related issues, and give us a whole brand new set of challenges :smile:

breautek (approved):
lgtm

raphinesse:
If nobody objects I will merge this into master soon. IMO this is more a fix than anything else.

NiklasMerz (approved):
Good to me. Let's see if this is still a problem 👍

PR #1104, opened 2020-10-20
feat: support gzip encoding requests & use GZIPInputStream

anasofiagribeiro:
This improves the energy efficiency of cordova-android by applying the Reduce Size Energy Pattern for mobile applications.
The energy pattern was applied in CordovaResourceApi.java. The general idea is when receiving data from an URLConnection or subclasses, if possible, receive the data compressed by GZIP scheme. In this specific case, when receiving data from the conn variable, request for the data to be compressed.

### DIFF
--- a/framework/src/org/apache/cordova/CordovaResourceApi.java
+++ b/framework/src/org/apache/cordova/CordovaResourceApi.java
@@ -41,6 +41,7 @@ import java.net.HttpURLConnection;
 import java.net.URL;
 import java.nio.channels.FileChannel;
 import java.util.Locale;
+import java.util.zip.GZIPInputStream;
 
 /**
  * What this class provides:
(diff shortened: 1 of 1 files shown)


timbru31 (approved):
Looks like a valuable addition. I'd like a second pair of eyes, though. According to the Android docs, this is available from API level 1 (https://developer.android.com/reference/java/util/zip/GZIPInputStream)

timbru31:
IMHO this could fit into the 9.1.0 release, too.

breautek:
> IMHO this could fit into the 9.1.0 release, too.

While I think this is okay to be in a minor release, @erisu already made a declaration for 9.1 so I think master should remain frozen.

breautek (approved):

erisu:
> @erisu already made a declaration for 9.1 so I think master should remain frozen.

If this PR can fit in the minor release, then it is still acceptable to be merged into the master branch, at this time.

When a discussion thread is made for a release, PRs can still be merged into the master branch as long as the PRs are fitting for that target release.

Only when a vote thread is created is when the master branch is completely frozen and no PRs should be merged in temporarily. This is so that if the vote fails for any reason, it can be easily corrected and resubmitted. (Even after a vote failure, more PRs could be merged in for the targeted release. hehe)

erisu (approved):

PR #1111, opened 2020-10-27
fix(android): allow file access for existing behavior

erisu:

Fixes SDK 30 related issues

Loading `file://` resources is failing in SDK API 30 because the `setAllowFileAccess` is now defaulted to `false`.

To keep existing behavior, we should set this to `true`, but in future releases we should default it to `false` and have the files loaded not though file. 

### Testing

With the camera plugin, took a photo and loaded the `FILE_URI` in `` tag.

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/engine/SystemWebViewEngine.java
+++ b/framework/src/org/apache/cordova/engine/SystemWebViewEngine.java
@@ -152,6 +152,15 @@ public class SystemWebViewEngine implements CordovaWebViewEngine {
         settings.setJavaScriptCanOpenWindowsAutomatically(true);
         settings.setLayoutAlgorithm(LayoutAlgorithm.NORMAL);
 
+        /**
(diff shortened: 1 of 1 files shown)


dpogue (approved):

ath0mas:
As described in the upper Motivation, here is the Android point on [setAllowFileAccess](https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccess(boolean)) since SDK 30:
```
The default value is true for apps targeting Build.VERSION_CODES.Q and below, and false when targeting Build.VERSION_CODES.R and above.
```

This explicit call should be added next to the `settings.setAllowUniversalAccessFromFileURLs(true);` line?
as both are since API 30 related to deprecated file:// URLs, and suggested use of [androidx.webkit.WebViewAssetLoader](https://developer.android.com/reference/androidx/webkit/WebViewAssetLoader)
(and maybe move this "group" a bit from the use of unrelated `setMediaPlaybackRequiresUserGesture`)

PR #1101, opened 2020-10-19
refactor: unify target resolution for devices & emulators

raphinesse:
### Motivation and Context

This change simplifies and deduplicates the code that, given a user-supplied target specification, finds a target to install and run an app on.

This refactoring was triggered by my ongoing work to stop copying `cordova/lib` to the platform folder.

### Description

- replaced `device.list` and `emulator.list_started` with `target.list`
- merged the virtually identical methods `{device, emulator}.resolveTarget` with the part of the target resolution that was implemented in the `run` module to form the new method `target.resolve`.
- the result of the target resolution remains unchanged, log and error messages are a bit different now.

### Testing

Updated unit tests pass.

PR #1123, opened 2020-11-17
refactor: unify installation on devices & emulators

raphinesse:
### Motivation and Context

`device.install` and `emulator.install` are virtually identical. The only notable difference is that `emulator.install` has some retry logic that is supposed to mitigate emulator hangs. The same goes for the unit tests.

Joining these methods reduces the code we have to maintain and makes upcoming refactors easier.

### Description

This change replaces `device.install` and `emulator.install` with the generic `target.install`.

The old installation methods did also resolve the passed target specification (by calling `{device,emulator}.resolveTarget`) if necessary. The new installation method expects the given installation target to be resolved already. This allows for a better separation of concerns.

### Testing

- Our unit tests pass
- I manually tested the affected platform-centric binaries. They behave like before this change (failing with `TypeError: Cannot read property 'apkPaths' of undefined`)


raphinesse:
This change is actually not independent from #1101. Sorry for the bother. Closing.

raphinesse:
OK, I fixed this PR up and it is now independent and ready to review. Sorry for the inconvenience.

erisu (approved):
Unified code and change LGTM 👍

Issue #1060, opened 2020-08-31
Add java version to getEnvironmentInfo


breautek:
# Feature Request

## Motivation Behind Feature

Completes platform info used by `cordova info` command for cordova troubleshooting and diagnosis.

Related to: https://github.com/apache/cordova-cli/issues/526

Issue #1078, opened 2020-09-28
cordova-android 9.0 problem - Could not initialize class com.android.build.gradle.internal.VariantManager


fjellandermedia:
### Issue Type

- [X ] Bug Report
- [ ] Feature Request
- [ ] Support Question

## Description
Build fails, even on new projects. It works with cordova-android@8.1, but not on cordova-android@9.0. This is the output:

>Gabriels-MBP:test gabriel$ cordova build android
>Checking Java JDK and Android SDK versions
>ANDROID_SDK_ROOT=/Users/gabriel/Library/Android/sdk (recommended setting)
>ANDROID_HOME=/Users/gabriel/Library/Android/sdk (DEPRECATED)
>Using Android SDK: /Users/gabriel/Library/Android/sdk
>Subproject Path: CordovaLib
>Subproject Path: app
>
>FAILURE: Build failed with an exception.
>
>* Where:
>Build file '/Users/gabriel/Documents/cordova/test/platforms/android/app/build.gradle' line: 20
>
>* What went wrong:
>A problem occurred evaluating project ':app'.
> Could not initialize class com.android.build.gradle.internal.VariantManager
>
>* Try:
>Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
>
>* Get more help at https://help.gradle.org
>
>BUILD FAILED in 1s
>Command failed with exit code 1: /Users/gabriel/Documents/cordova/test/platforms/android/gradlew cdvBuildDebug -b >/Users/gabriel/Documents/cordova/test/platforms/android/build.gradle`

I have updated to latest Cordova CLI and latest SDK:s and build tools in Android Studio.

### Command or Code
>cordova create test se.fjellandermedia.test testApp
>cd test
>cordova platform add android
>cordova build android

### Environment, Platform, Device

Macbook Pro, macOS 10.15.6

## Checklist

- [ X] I searched for already existing GitHub issues about this
- [X ] I updated all Cordova tooling to their most recent version
- [ X] I included all the necessary information above

fjellandermedia:
Can someone please help me with this. I've tried countless things: tweaked gradle version, kotlin version, updated everything in Android Studio, removing and reinstalling the cordova platform, reinstalling cordova, and more. I always get the same error message. Everything works on cordova-android@8.1.0 but not on cordova-android@9.0.0.

breautek:
Please provide the output of `cordova info`

[...]

PR #1130, opened 2020-11-22
refactor: java checks / get java version

breautek:
### Platforms affected
Android

### Motivation and Context

Prerequisite PR for https://github.com/apache/cordova-android/issues/1060

This PR abstracts the Java checks and java version retrieval so that it can be reused.

### Description

Created a new Java class with static methods that provides an API to get the version. It also contains the code that finds the java executable that used to live inside `check_reqs`.

`Java.js` is contains 100% line coverage, which is an upgrade from when the code lived in `check_reqs`.

### Testing

`npm test`

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/check_reqs.js
+++ b/bin/templates/cordova/lib/check_reqs.js
@@ -20,30 +20,20 @@
 const execa = require('execa');
 var path = require('path');
 var fs = require('fs-extra');
-var os = require('os');
-var which = require('which');
-const glob = require('fast-glob');
+const { forgivingWhichSync, isWindows, isDarwin } = require('./utils');
+const java = require('./env/java');
 var REPO_ROOT = path.join(__dirname, '..', '..', '..', '..');
 var PROJECT_ROOT = path.join(__dirname, '..', '..');
 const { CordovaError, ConfigParser, events } = require('cordova-common');
 var android_sdk = require('./android_sdk');
 const { createEditor } = require('properties-parser');
+const semver = require('semver');
 
-function forgivingWhichSync (cmd) {
-    const whichResult = which.sync(cmd, { nothrow: true });
+const EXPECTED_JAVA_VERSION = '1.8.x';
 
-    // On null, returns empty string to maintain backwards compatibility
-    // realpathSync follows symlinks
-    return whichResult === null ? '' : fs.realpathSync(whichResult);
-}
-
-module.exports.isWindows = function () {
-    return (os.platform() === 'win32');
-};
-
-module.exports.isDarwin = function () {
-    return (os.platform() === 'darwin');
-};
+// Re-exporting these for backwards compatibility and for unit testing.
+// TODO: Remove uses and use the ./utils module directly.
+Object.assign(module.exports, { isWindows, isDarwin });
 
 /**
  * @description Get valid target from framework/project.properties if run from this repo
@@ -143,72 +133,18 @@ module.exports.check_gradle = function () {
                             'in your path, or install Android Studio'));
 };
 
-// Returns a promise.
-module.exports.check_java = function () {
-    var javacPath = forgivingWhichSync('javac');
-    var hasJavaHome = !!process.env.JAVA_HOME;
-    return Promise.resolve().then(function () {
-        if (hasJavaHome) {
-            // Windows java installer doesn't add javac to PATH, nor set JAVA_HOME (ugh).
-            if (!javacPath) {
-                process.env.PATH += path.delimiter + path.join(process.env.JAVA_HOME, 'bin');
-            }
-        } else {
-            if (javacPath) {
-                // OS X has a command for finding JAVA_HOME.
-                var find_java = '/usr/libexec/java_home';
-                var default_java_error_msg = 'Failed to find \'JAVA_HOME\' environment variable. Try setting it manually.';
-                if (fs.existsSync(find_java)) {
-                    return execa(find_java).then(({ stdout }) => {
-                        process.env.JAVA_HOME = stdout;
-                    }).catch(function (err) {
-                        if (err) {
-                            throw new CordovaError(default_java_error_msg);
-                        }
-                    });
-                } else {
-                    // See if we can derive it from javac's location.
-                    var maybeJavaHome = path.dirname(path.dirname(javacPath));
-                    if (fs.existsSync(path.join(maybeJavaHome, 'lib', 'tools.jar'))) {
-                        process.env.JAVA_HOME = maybeJavaHome;
-                    } else {
-                        throw new CordovaError(default_java_error_msg);
-                    }
-                }
-            } else if (module.exports.isWindows()) {
-                const { env } = process;
-                const baseDirs = [env.ProgramFiles, env['ProgramFiles(x86)']];
-                const globOpts = { absolute: true, onlyDirectories: true };
-                const flatMap = (arr, f) => [].concat(...arr.map(f));
+/**
+ * Checks for the java installation and correct version
+ */
+module.exports.check_java = async function () {
+    const javaVersion = await java.getVersion();
 
-                const jdkDir = flatMap(baseDirs, cwd =>
-                    glob.sync('java/jdk*', { cwd, ...globOpts })
-                )[0];
-
-                if (jdkDir) {
-                    env.PATH += path.delimiter + path.join(jdkDir, 'bin');
-                    env.JAVA_HOME = path.normalize(jdkDir);
-                }
-            }
-        }
-    }).then(function () {
-        return execa('javac', ['-version'], { all: true })
-            .then(({ all: output }) => {
-                // Java <= 8 writes version info to stderr, Java >= 9 to stdout
-                const match = /javac\s+([\d.]+)/i.exec(output);
-                return match && match[1];
-            }, () => {
-                var msg =
-                'Failed to run "javac -version", make sure that you have a JDK version 8 installed.\n' +
-                'You can get it from the following location:\n' +
-                'https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html';
-                if (process.env.JAVA_HOME) {
-                    msg += '\n\n';
-                    msg += 'Your JAVA_HOME is invalid: ' + process.env.JAVA_HOME;
-                }
-                throw new CordovaError(msg);
-            });
-    });
+    if (!semver.satisfies(javaVersion.version, EXPECTED_JAVA_VERSION)) {
+        throw new CordovaError(
+            `Requirements check failed for JDK ${EXPECTED_JAVA_VERSION}! Detected version: ${javaVersion.version}\n` +
+            'Check your ANDROID_SDK_ROOT / JAVA_HOME / PATH environment variables.'
+        );
+    }
 };
 
 // Returns a promise.
@@ -383,13 +319,6 @@ module.exports.run = function () {
 
     return Promise.all([this.check_java(), this.check_android()]).then(function (values) {
         console.log('Using Android SDK: ' + process.env.ANDROID_SDK_ROOT);
-
-        if (!String(values[0]).startsWith('1.8.')) {
-            throw new CordovaError(
-                'Requirements check failed for JDK 8 (\'1.8.*\')! Detected version: ' + values[0] + '\n' +
-                'Check your ANDROID_SDK_ROOT / JAVA_HOME / PATH environment variables.'
-            );
-        }
 
         if (!values[1]) {
             throw new CordovaError('Requirements check failed for Android SDK! Android SDK was not detected.');
--- a/bin/templates/cordova/lib/env/java.js
+++ b/bin/templates/cordova/lib/env/java.js
@@ -1,0 +1,123 @@
+/*
+       Licensed to the Apache Software Foundation (ASF) under one
+       or more contributor license agreements.  See the NOTICE file
+       distributed with this work for additional information
+       regarding copyright ownership.  The ASF licenses this file
+       to you under the Apache License, Version 2.0 (the
+       "License"); you may not use this file except in compliance
+       with the License.  You may obtain a copy of the License at
+
+         http://www.apache.org/licenses/LICENSE-2.0
+
+       Unless required by applicable law or agreed to in writing,
+       software distributed under the License is distributed on an
+       "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+       KIND, either express or implied.  See the License for the
+       specific language governing permissions and limitations
+       under the License.
+*/
+
+const execa = require('execa');
+const fs = require('fs-extra');
+const path = require('path');
+const glob = require('fast-glob');
+const { CordovaError, events } = require('cordova-common');
+const utils = require('../utils');
+const semver = require('semver');
+
+/**
+ * Will be set to true on successful ensureness.
+ * If true, skips the expensive java checks.
+ */
+let javaIsEnsured = false;
+
+const java = {
+    /**
+     * Gets the version from the javac executable.
+     *
+     * @returns {semver.SemVer}
+     */
+    getVersion: async () => {
+        await java._ensure(process.env);
+
+        // Java <= 8 writes version info to stderr, Java >= 9 to stdout
+        let version = null;
+        try {
+            version = (await execa('javac', ['-version'], { all: true })).all;
+        } catch (ex) {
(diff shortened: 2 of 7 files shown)


bin/templates/cordova/lib/Java.js
```diff
@@ -0,0 +1,157 @@
...
+     * }
+     */
+    static _parseVersion (version) {
+        version = version.replace('javac ', '');
+
+        const mainParts = version.split('.');
+
+        const output = {
+            major: parseInt(mainParts[0]),
+            minor: parseInt(mainParts[1]),
```
erisu, line 83:
```suggestion
            major: parseInt(mainParts[0], 10),
            minor: parseInt(mainParts[1], 10),
```

I typically always try to include the `radix` property for reliability. There used to be cases in the past that lead to unexpected results, but I think most of these were resolved since, so providing it might not be as necessary.

bin/templates/cordova/lib/Java.js
```diff
@@ -0,0 +1,157 @@
...
+            patch: null,
+            build: null,
+            string: version
+        };
+
+        const buildParts = mainParts[2].split('_');
+        output.patch = parseInt(buildParts[0]);
+        output.build = parseInt(buildParts[1]);
```
erisu, line 91:
```suggestion
        output.patch = parseInt(buildParts[0], 10);
        output.build = parseInt(buildParts[1], 10);
```

spec/unit/Java.spec.js
```diff
@@ -0,0 +1,156 @@
...
+        it('derive from javac location - error', async () => {
+            spyOn(utils, 'forgivingWhichSync').and.returnValue('/tmp/jdk/bin');
+            process.env.JAVA_HOME = '';
+            Java.__set__('fs', {
+                existsSync: jasmine.createSpy('fs').and.callFake((path) => {
+                    if (/java_home$/.test(path)) {
+                        return false;
+                    } else {
+                        return false;
+                    }
```
erisu, line 149: Always returns `false`? Seems like the `if` condition is not needed.

erisu (commented):
Do we really want the name of the file as `Java.js` and not `java.js`? I believe in almost every repo our filenames are lowercased.

Since this is introducing a new coding style, I think you should document when to use uppercase vs. lowercase and put a document somewhere like the cordova-contribute so we can understand these new cases.

raphinesse:
> Do we really want the name of the file as `Java.js` and not `java.js`? I believe in almost every repo our filenames are lowercased.

We actually have quite a few modules with non-lowercase filenames in this repo that exactly match the exported class' name: `AndroidManifest.js`, `AndroidProject.js`, ...

raphinesse (commented):
Great to see improvements to the `check_reqs` module! Especially the separation of environment checking and inference is something I thought should have been done for a long time.

I made a quick pass over the code, left a few comments and pushed a fix for one of the new tests. Overall, this looks quite good to me. I hope I can take a closer look at it soon.

bin/templates/cordova/lib/Java.js
```diff
@@ -0,0 +1,157 @@
...
+            }
+        } else {
+            if (javacPath) {
+                // OS X has a command for finding JAVA_HOME.
+                const find_java = '/usr/libexec/java_home';
+                const default_java_error_msg = 'Failed to find \'JAVA_HOME\' environment variable. Try setting it manually.';
+                if (fs.existsSync(find_java)) {
+                    try {
+                        process.env.JAVA_HOME = (await execa(find_java)).stdout;
+                        console.log('JAVA HOME', process.env.JAVA_HOME);
```
raphinesse, line 122: This looks like leftover debug output to me
```suggestion
```

bin/templates/cordova/lib/Java.js
```diff
@@ -0,0 +1,157 @@
...
+                    } catch (ex) {
+                        if (ex) {
+                            events.emit('verbose', ex.shortMessage);
+                        }
```
raphinesse, line 126: I know that the original version had this if-clause too, but according to the `execa` docs, `ex` should always be a `childProcessResult` and thus truthy. So I would suggest to use 
```suggestion
                        events.emit('verbose', ex.shortMessage);
```

bin/templates/cordova/lib/Java.js
```diff
@@ -0,0 +1,157 @@
...
+     * @returns {Object}
+     * {
+     *      major: number
+     *      minor: number
+     *      patch: number
+     *      build: number
+     * }
+     */
+    static async getVersion () {
+        await Java._ensure();
```
raphinesse, line 41: To avoid having to restore the global `process.env` properly before/after each test, I would explicitly pass the `env` object to be modified into `Java._ensure` like this:
```suggestion
        await Java._ensure(process.env);
```
Then tests can just pass fake environments to `_ensure`.

spec/unit/Java.spec.js
```diff
@@ -0,0 +1,158 @@
...
+        it('without JAVA_HOME / with javac - Mac OS X - success', async () => {
+            spyOn(utils, 'forgivingWhichSync').and.returnValue('/tmp/jdk/bin');
+            process.env.JAVA_HOME = '';
+            const fsSpy = jasmine.createSpy('fs').and.returnValue(true);
+            Java.__set__('fs', {
+                existsSync: fsSpy
+            });
+            Java.__set__('execa', jasmine.createSpy('execa').and.returnValue(Promise.resolve({
+                stdout: '/tmp/jdk'
+            })));
```
raphinesse, line 97: If we do not use the features of Jasmine spies, I would prefer using plain JS function as test-fakes. It makes for far less verbose code IMHO.
```suggestion
            Java.__set__('execa', async () => ({ stdout: '/tmp/jdk' }));
```

This might be a matter of taste of course.

spec/unit/Java.spec.js
```diff
@@ -0,0 +1,158 @@
...
+            expect(fsSpy).toHaveBeenCalledWith('/usr/libexec/java_home');
+            expect(process.env.JAVA_HOME).toBe('/tmp/jdk');
+        });
+
+        it('without JAVA_HOME / with javac - Mac OS X - error', async () => {
+            spyOn(utils, 'forgivingWhichSync').and.returnValue('/tmp/jdk/bin');
+            process.env.JAVA_HOME = '';
+            Java.__set__('fs', {
+                existsSync: jasmine.createSpy('fs').and.returnValue(true)
+            });
```
raphinesse, line 110: Could also be simplified by using plain functions:
```suggestion
            Java.__set__('fs', { existsSync: () => true });
```

spec/unit/Java.spec.js
```diff
@@ -0,0 +1,158 @@
...
+            Java.__set__('execa', jasmine.createSpy('execa').and.returnValue(Promise.reject({
+                shortMessage: 'test error'
+            })));
```
raphinesse, line 113: Latest jasmine also has special promise strategies:
```suggestion
            Java.__set__('execa', jasmine.createSpy('execa').and.rejectWith({
                shortMessage: 'test error'
            }));
```

spec/unit/Java.spec.js
```diff
@@ -0,0 +1,158 @@
...
+            process.env.JAVA_HOME = '';
+            Java.__set__('fs', {
+                existsSync: jasmine.createSpy('fs').and.callFake((path) => {
+                    if (/java_home$/.test(path)) {
+                        return false;
+                    } else {
+                        return true;
+                    }
+                })
+            });
```
raphinesse, line 135: I would much prefer something a little less verbose here:
```suggestion
            Java.__set__('fs', { existsSync: path => !/java_home$/.test(path) });
```

spec/unit/Java.spec.js
```diff
@@ -0,0 +1,158 @@
...
+            process.env.JAVA_HOME = '';
+            Java.__set__('fs', {
+                existsSync: jasmine.createSpy('fs').and.callFake((path) => {
+                    if (/java_home$/.test(path)) {
+                        return false;
+                    } else {
+                        return false;
+                    }
+                })
+            });
```
raphinesse, line 153: AFAICT this should be the same fake implementation as above
```suggestion
            Java.__set__('fs', { existsSync: path => !/java_home$/.test(path) });
```

raphinesse (commented):
I added some more comments. I hope I didn't go overboard :sweat_smile:

bin/templates/cordova/lib/check_reqs.js
```diff
@@ -143,72 +134,23 @@ module.exports.check_gradle = function () {
...
+/**
+ * Checks for the java installation and correct version
+ */
+module.exports.check_java = async function () {
+    const EXPECTED_JAVA_MAJOR = 1;
+    const EXPECTED_JAVA_MINOR = 8;
+
+    const javaVersion = await Java.getVersion();
+
+    if (javaVersion.major < EXPECTED_JAVA_MAJOR || javaVersion.minor < EXPECTED_JAVA_MINOR) {
```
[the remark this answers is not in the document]
raphinesse: Oh, and the actual version constraint should stay `1.8.x` for this PR. If we want to relax this version requirement, we should do it separately, not hidden in a big refactor.

spec/unit/Java.spec.js
breautek: Yes, this seems like something I missed. I'll re investigate probably on the weekend to determine what this should be. Would have thought I had picked up on this when writing the tests...

bin/templates/cordova/lib/Java.js
```diff
@@ -0,0 +1,157 @@
...
+            'You can get it from the following location:\n' +
+            'https://www.oracle.com/technetwork/java/javase/downloads/jdk8-downloads-2133151.html';
+            if (process.env.JAVA_HOME) {
+                msg += '\n\n';
+                msg += 'Your JAVA_HOME is invalid: ' + process.env.JAVA_HOME;
+            }
+            throw new CordovaError(msg);
+        }
+
+        return Java._parseVersion(version);
```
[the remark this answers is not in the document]
breautek: Using `coerce` will drop the build number, and while I'm not using it in this PR... I think it's something that could be used. [Someone](https://github.com/apache/cordova-android/issues/1078#issuecomment-717944583) had a problem that was caused by using a really old version of java 1.8. Of course if we wanted to add a constraint on the build number, we'd should fine what would be the minimal acceptable number and that could be difficult to do... Thoughts?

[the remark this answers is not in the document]
raphinesse: Huh, I definitely see your point. I guess I would still go with `semver.coerce` and ignore the build part. As you say, we don't even know which build number we should be requiring. Furthermore newer Java versions do not even seem to contain that part in their version numbers. So the number of people that could be affected by such a problem is only diminishing.

bin/templates/cordova/lib/check_reqs.js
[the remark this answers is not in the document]
breautek: I changed the constraint to be `1.8.x`, especially since how https://github.com/apache/cordova-android/pull/928 turned out.

spec/unit/Java.spec.js
breautek: I opted for an earlier suggestion which just set a new function to call rather than using jasmine spies.

raphinesse (approved):
This PR looks pretty good to me :+1:

My suggestions and comments are mainly cosmetic. Feel free to implement or discard those as you see fit.

The few non-cosmetic remarks are not a deal breaker either for me. So if there's no time to address them, that's fine with me too.

Sorry for taking so long with the review but I'm pretty swamped with work currently.

spec/unit/java.spec.js
```diff
@@ -0,0 +1,158 @@
...
+const path = require('path');
+const rewire = require('rewire');
+const { CordovaError } = require('cordova-common');
+const utils = require('../../bin/templates/cordova/lib/utils');
+const glob = require('fast-glob');
+
+describe('Java', () => {
+    const Java = rewire('../../bin/templates/cordova/lib/env/java');
+
+    describe('getVersion', () => {
```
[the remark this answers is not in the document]
breautek: Yah, I can make a spy that makes it call the original function, but with a parameter value other than the actual `process.env` (maybe just a copy of it).

bin/templates/cordova/lib/check_reqs.js
```diff
@@ -143,72 +137,20 @@ module.exports.check_gradle = function () {
...
+    const javaVersion = await java.getVersion();
+
+    if (!semver.satisfies(javaVersion.version, `= ${EXPECTED_JAVA_VERSION}`)) {
+        throw new CordovaError(
+            `Requirements check failed for JDK ${EXPECTED_JAVA_VERSION}! Detected version: ${javaVersion.version}\n` +
+            'Check your ANDROID_SDK_ROOT / JAVA_HOME / PATH environment variables.'
+        );
+    }
+
+    return javaVersion.string;
```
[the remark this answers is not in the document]
breautek: Not really sure what I was thinking at the time of writing, but a "check" method shouldn't really be returning anything, so I'll remove this, which would have matched the old behaviour as well.

bin/templates/cordova/lib/env/java.js
```diff
@@ -0,0 +1,119 @@
...
+                )[0];
+
+                if (jdkDir) {
+                    environment.PATH += path.delimiter + path.join(jdkDir, 'bin');
+                    environment.JAVA_HOME = path.normalize(jdkDir);
+                }
+            }
+        }
+
+        javaIsEnsured = true;
```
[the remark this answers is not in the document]
breautek: I don't think we want to cache errors.

Doing so may make recovery attempts harder (not that there are any recovery attempts currently).

bin/templates/cordova/lib/check_reqs.js
```diff
@@ -20,29 +20,23 @@
...
 
 module.exports.isWindows = function () {
-    return (os.platform() === 'win32');
+    return isWindows();
 };
 
 module.exports.isDarwin = function () {
-    return (os.platform() === 'darwin');
+    return isDarwin();
 };
```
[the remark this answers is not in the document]
breautek: Yes, removing these two functions will require a bigger refactor which probably shouldn't be a part of this PR.

I'll add the shorten version with a comment and a TODO

breautek:
@erisu @raphinesse 

Finally got around into finishing this up, I believe this could be incorporated into 9.1.x release, or do you think it should wait until 10.x?

erisu (approved):
IMO, I think this can be released with the 9.1.0 minor release.

DavidStrausz:
Hi @breautek, I hope it's okay if I add this here, but I can also open a new issue if you need one. 
This causes the requirements check to fail on macOS Big Sur 11.2.3 if custom java options (`_JAVA_OPTIONS`) are set.

For reference in my case I set them in `.zshrc` like this:

```bash
export _JAVA_OPTIONS="-Xms2048M -Xmx4096M"
```

And this is the exact error:
```bash
ANDROID_SDK_ROOT=/Users/username/Library/Android/sdk (recommended setting)
ANDROID_HOME=undefined (DEPRECATED)
Requirements check failed for JDK 1.8.x! Detected version: 2048.0.0
Check your ANDROID_SDK_ROOT / JAVA_HOME / PATH environment variables.
```

The reason for this is happening is because in the new implementation you are using `javac -version` to get the version, this prints the java options before printing the version like this:

```bash
Picked up _JAVA_OPTIONS: -Xms2048M -Xmx4096M
javac 1.8.0_271
```

The result is directly checked via semver without sanitizing the output first. In the old implementation a regular expression was used to only check the relevant parts:

```js
const match = /javac\s+([\d.]+)/i.exec(output);
```
https://github.com/apache/cordova-android/pull/1130/files#diff-0d1afa0d5c88bdb18fbc6ab9ef7fc4bbe10f7047d387bb2964932dbb6eb8c4b7L198

raphinesse:
@DavidStrausz thanks for letting us know. Would you care to try if this problem can be resolved by adding the options `env: {}, extendEnv: false` to the `execa` call in [`getVersion`](https://github.com/breautek/cordova-android/blob/17dbc1d9f8df6ab72b18b946dc6a7ad57d6ff712/bin/templates/cordova/lib/env/java.js#L46)?

If so, we surely would appreciate a pull request with that change.

DavidStrausz:
@raphinesse Thanks for your response! Interesting, didn't think of that way of fixing it yet, I will take a look sometime next week and will open a PR if this indeed resolves the issue!
If it does not work, I will readd the regex in my PR.

PR #1148, opened 2021-01-18
fix: add not null checks to prevent running on destroyed activity

### LINKED ISSUE #1002: NullPointerException: 'void android.app.Activity.runOnUiThread(java.lang.Runnable)' on a null object reference
# Bug Report

Cordova should silently exit.

Cordova crashes because the parent Activity is not alive anymore

## Information

Start a CordovaWebview and load a slow/large website.

`cordovaWebView.loadUrl(url)`

If you return to the previous activity (by using the (hardware) back button), cordova will try to execute `cordova.getActivity().runOnUiThread()`, but the activity is no longer there.

e.g., https://github.com/apache/cordova-android/blob/master/framework/src/org/apache/cordova/CordovaWebViewImpl.java#L187

Stacktrace:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'void android.app.Activity.runOnUiThread(java.lang.Runnable)' on a null object reference
        at org.apache.cordova.CordovaWebViewImpl$2.run(CordovaWebViewImpl.java:179)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:764)
```

### Command or Code

See information above on how to reproduce.

Android / All devices / All OS

### Version information

Android Studio 
Cordova 8.1.0

## Checklist

- [X] I searched for existing GitHub issues:
    Found issue: https://github.com/apache/cordova-android/issues/884 but this was closed due to lack of information
- [X] I updated all Cordova tooling to most recent version:
    Crash occured on 
    'org.apache.cordova:cordova:6.2.3'
    Updated to (crashed as well):
    'org.apache.cordova:framework:8.1.0'
- [X] I included all the necessary information above

secondVISION:
### Platforms affected
Android 

### Motivation and Context

See issue and sample MVP/Videos in : https://github.com/apache/cordova-android/issues/1002
Fixes #1002 

### Description

Added null pointer checks around ```cordova.getActivity()```, to check if the activity hosting Cordova is still alive.

### Testing

Updated MVP (see issue #1002 ) locally with these changes (also updated gradle but is not pushed in this PR).

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/CordovaWebViewImpl.java
+++ b/framework/src/org/apache/cordova/CordovaWebViewImpl.java
@@ -176,22 +176,24 @@ public class CordovaWebViewImpl implements CordovaWebView {
                     e.printStackTrace();
                 }
 
-                // If timeout, then stop loading and handle error
-                if (loadUrlTimeout == currentLoadUrlTimeout) {
+                // If timeout, then stop loading and handle error (if activity still exists)
+                if (loadUrlTimeout == currentLoadUrlTimeout && cordova.getActivity() != null) {
                     cordova.getActivity().runOnUiThread(loadError);
                 }
             }
         };
 
-        final boolean _recreatePlugins = recreatePlugins;
-        cordova.getActivity().runOnUiThread(new Runnable() {
-            public void run() {
-                if (loadUrlTimeoutValue > 0) {
-                    cordova.getThreadPool().execute(timeoutCheck);
+        if (cordova.getActivity() != null) {
+            final boolean _recreatePlugins = recreatePlugins;
+            cordova.getActivity().runOnUiThread(new Runnable() {
+                public void run() {
+                    if (loadUrlTimeoutValue > 0) {
+                        cordova.getThreadPool().execute(timeoutCheck);
+                    }
+                    engine.loadUrl(url, _recreatePlugins);
                 }
-                engine.loadUrl(url, _recreatePlugins);
-            }
-        });
+            });
+        }
     }
 
 
@@ -238,7 +240,9 @@ public class CordovaWebViewImpl implements CordovaWebView {
             } else {
                 intent.setData(uri);
             }
-            cordova.getActivity().startActivity(intent);
+            if (cordova.getActivity() != null) {
+                cordova.getActivity().startActivity(intent);
+            }
         } catch (android.content.ActivityNotFoundException e) {
             LOG.e(TAG, "Error loading url " + url, e);
         }
@@ -553,11 +557,13 @@ public class CordovaWebViewImpl implements CordovaWebView {
                     public void run() {
                         try {
                             Thread.sleep(2000);
-                            cordova.getActivity().runOnUiThread(new Runnable() {
-                                public void run() {
-                                    pluginManager.postMessage("spinner", "stop");
-                                }
-                            });
+                            if (cordova.getActivity() != null) {
+                                cordova.getActivity().runOnUiThread(new Runnable() {
+                                    public void run() {
+                                        pluginManager.postMessage("spinner", "stop");
+                                    }
+                                });
+                            }
                         } catch (InterruptedException e) {
                         }
                     }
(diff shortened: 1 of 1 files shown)


framework/src/org/apache/cordova/CordovaWebViewImpl.java
```diff
@@ -176,22 +176,24 @@ public void run() {
                     e.printStackTrace();
                 }
 
-                // If timeout, then stop loading and handle error
-                if (loadUrlTimeout == currentLoadUrlTimeout) {
+                // If timeout, then stop loading and handle error (if activity still exists)
+                if (loadUrlTimeout == currentLoadUrlTimeout && cordova.getActivity() != null) {
```
breautek, line 180: I wonder if we should add an else condition to log a warning/debug statement when the activity no longer exists. This may help provide better insights for developers when they hit these edge cases unexpectedly.

e.g:

```java
if (loadUrlTimeout == currentLoadUrlTimeout && cordova.getActivity() != null) {
   ...
} else if (cordova.getActivity() == null) {
    LOG.d(TAG, "Cordova activity does not exists.");
}

File: framework/src/org/apache/cordova/CordovaWebViewImpl.java [hunk 2]
```diff
@@ -176,22 +176,24 @@ public void run() {
...
                 }
             }
         };
 
-        final boolean _recreatePlugins = recreatePlugins;
-        cordova.getActivity().runOnUiThread(new Runnable() {
-            public void run() {
-                if (loadUrlTimeoutValue > 0) {
-                    cordova.getThreadPool().execute(timeoutCheck);
+        if (cordova.getActivity() != null) {
```
breautek, line 186: If we do decide if logging debug statements is a good idea, then one should be also printed if this condition fails.

framework/src/org/apache/cordova/CordovaWebViewImpl.java
```diff
@@ -238,7 +240,9 @@ public void showWebPage(String url, boolean openExternal, boolean clearHistory,
             } else {
                 intent.setData(uri);
             }
-            cordova.getActivity().startActivity(intent);
+            if (cordova.getActivity() != null) {
```
breautek, line 243: If we do decide if logging debug statements is a good idea, then one should be also printed if this condition fails.

framework/src/org/apache/cordova/CordovaWebViewImpl.java
```diff
@@ -553,11 +557,13 @@ public void onPageFinishedLoading(String url) {
                     public void run() {
                         try {
                             Thread.sleep(2000);
-                            cordova.getActivity().runOnUiThread(new Runnable() {
-                                public void run() {
-                                    pluginManager.postMessage("spinner", "stop");
-                                }
-                            });
+                            if (cordova.getActivity() != null) {
```
breautek, line 560: If we do decide if logging debug statements is a good idea, then one should be also printed if this condition fails.

breautek (commented):
Thank you for your PR.

I left a few comments, but I'm open minded. Let me know what you think.

Generally speaking I think this PR is :+1: I'll try to give this a more in-depth test later.

framework/src/org/apache/cordova/CordovaWebViewImpl.java
secondVISION: Ah sure, indeed. I'll add those logs!

breautek:
I think the failure is a lie, I'll restart the checks.

secondVISION:
Let me know if anything else is needed!

breautek (approved):

erisu (approved):
Code LGTM

breautek:
Thank you for your contribution @secondVISION

PR #1157, opened 2021-01-21
fix: restore key event handlers when DOM element is fullscreen

### LINKED ISSUE #1156: Key event not reaching JS when a DOM element is fullscreen
Using cordova-android@9.0.0, when a DOM element has requested fullscreen (`Element.requestFullscreen()`), installed key event handler are not executed anymore, until the webview exits fullscreen.

For instance:
1. create a cordova app
2. `document.addEventListener('keydown', x => console.log(x), false)`
3. `document.body.requestFullscreen()`
4. generate keyboard events and inspect JS console

I believe this is due to this change: https://github.com/apache/cordova-android/commit/71f63d7b331af07bdcbb236f28474a874383f4e8. In particular, `dispatchKeyEvent` from base class is not called. I would suggest to use this instead:

```
@Override
public boolean dispatchKeyEvent(KeyEvent event) {
    boolean ret = engine.getView().dispatchKeyEvent(event);
    if (! ret) {
        // If the engine didn't handle the event, handle it normally.
        ret = super.dispatchKeyEvent(event);
    }
    return ret;
}
```

- [ ] I included all the necessary information above

goffioul:
Make sure to call dispatchKeyEvent from base class in WrapperView, if the event hasn't been handled by the engine.

Fixes #1156

### DIFF
--- a/framework/src/org/apache/cordova/CordovaWebViewImpl.java
+++ b/framework/src/org/apache/cordova/CordovaWebViewImpl.java
@@ -255,7 +255,12 @@ public class CordovaWebViewImpl implements CordovaWebView {
 
         @Override
         public boolean dispatchKeyEvent(KeyEvent event) {
-            return engine.getView().dispatchKeyEvent(event);
+            boolean ret = engine.getView().dispatchKeyEvent(event);
+            if (! ret) {
+                // If the engine didn't handle the event, handle it normally.
+                ret = super.dispatchKeyEvent(event);
+            }
+            return ret;
         }
     }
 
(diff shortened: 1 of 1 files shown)


breautek (approved):
lgtm :+1: 

Thank you for your contribution.

framework/src/org/apache/cordova/CordovaWebViewImpl.java
```diff
@@ -255,7 +255,12 @@ public WrapperView(Context context, CordovaWebViewEngine engine) {
 
         @Override
         public boolean dispatchKeyEvent(KeyEvent event) {
-            return engine.getView().dispatchKeyEvent(event);
+            boolean ret = engine.getView().dispatchKeyEvent(event);
+            if (! ret) {
```
erisu, line 259:
```suggestion
            if (!ret) {
```

That space :D

erisu (approved):

PR #1166, opened 2021-02-18
feat: overload PluginEntry constructor to set onload property

### LINKED ISSUE #1164: Provide PluginEntry with plugin instances causes assertion to crash app
# Bug Report

## Problem

After upgrade AGP to 4.1.0 our cordova based solution started to crash on Debug builds.

Not crash ;)

We use Cordova in a way where we provide [PluginEntry](https://github.com/apache/cordova-android/blob/master/framework/src/org/apache/cordova/PluginEntry.java#L51) with existing instance of CordovaServices.

This started to crash after we upgraded to AGP 4.1.0 from assertions [HERE](https://github.com/apache/cordova-android/blob/master/framework/src/org/apache/cordova/CordovaPlugin.java#L52)

Turns out AGP 4.1.0 changed config so all assertions will always throw on debug builds 
>**Assertions in debug builds**
>When you build the debug version of your app using Android Gradle plugin 4.1.0 and higher, the built-in compiler (D8) will >rewrite your app's code to enable assertions at compile time, so you always have assertion checks active.
source: [AGP release notes](https://developer.android.com/studio/releases/gradle-plugin#4.1-debug-assertions)

## Information
So far I found out that plugins provided from outside are initialized twice here [CordovaWebViewImpl:init](https://github.com/apache/cordova-android/blob/3081e5e6e91c32e507423411196b184325baa935/framework/src/org/apache/cordova/CordovaWebViewImpl.java#L97)

First is when PluginManager is created on [L:103](https://github.com/apache/cordova-android/blob/3081e5e6e91c32e507423411196b184325baa935/framework/src/org/apache/cordova/CordovaWebViewImpl.java#L103)
As `plugin.entry` is not null it is initialised here first time
Screen Shot 2021-02-17 at 11 31 02 am

Right arter this in `CordovaWebViewImpl:init` calls [pluginManager.init()](https://github.com/apache/cordova-android/blob/3081e5e6e91c32e507423411196b184325baa935/framework/src/org/apache/cordova/CordovaWebViewImpl.java#L117) which basically initialize plugins again and causes assertion to fail.

What more I found is that if only `PluginEntry.load` could be set to false if plugin instance is provided it would not initialize them twice. Not sure if it breaks something else though.

It seems like a bug to me but if it's not please tell me what is the proper way to provide instances of plugin from outside.

Android, Any

### Version information
8.1.0 - because it's last version that supports android 21

## Checklist

jblejder:
### Platforms affected
Android

### Motivation and Context
resolves https://github.com/apache/cordova-android/issues/1164
This will allow to specify already instantiated PluginEntries with `onload = false`. 
Which will stop Cordova from initialising them multiple time and fix issue on AGP 4.1.2 that enables all asserts in debug 

Is setting onLoad false valid option here?

### Testing
have a look at test exposing issue:
https://github.com/jblejder/cordova-android/pull/1

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/PluginEntry.java
+++ b/framework/src/org/apache/cordova/PluginEntry.java
@@ -47,11 +47,19 @@ public final class PluginEntry {
 
     /**
      * Constructs with a CordovaPlugin already instantiated.
+     *
+     * @param service               The name of the service
+     * @param pluginClass           The plugin class name
      */
     public PluginEntry(String service, CordovaPlugin plugin) {
         this(service, plugin.getClass().getName(), true, plugin);
     }
 
+    /**
+     * @param service               The name of the service
+     * @param plugin                The CordovaPlugin already instantiated
+     * @param onload                Create plugin object when HTML page is loaded
+     */
     public PluginEntry(String service, CordovaPlugin plugin, boolean onload) {
         this(service, plugin.getClass().getName(), onload, plugin);
     }
@@ -65,6 +73,12 @@ public final class PluginEntry {
         this(service, pluginClass, onload, null);
     }
 
+    /**
+     * @param service               The name of the service
+     * @param pluginClass           The plugin class name
+     * @param onload                Create plugin object when HTML page is loaded
+     * @param plugin                The CordovaPlugin already instantiated
+     */
     private PluginEntry(String service, String pluginClass, boolean onload, CordovaPlugin plugin) {
         this.service = service;
         this.pluginClass = pluginClass;
--- a/framework/src/org/apache/cordova/PluginEntry.java
+++ b/framework/src/org/apache/cordova/PluginEntry.java
@@ -52,6 +52,10 @@ public final class PluginEntry {
         this(service, plugin.getClass().getName(), true, plugin);
     }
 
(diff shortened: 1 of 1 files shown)


framework/src/org/apache/cordova/PluginEntry.java
```diff
@@ -52,6 +52,10 @@ public PluginEntry(String service, CordovaPlugin plugin) {
         this(service, plugin.getClass().getName(), true, plugin);
     }
 
+    public PluginEntry(String service, CordovaPlugin plugin, boolean onload) {
```
timbru31, line 55: Could you add a JavaDoc comment for this overload? Thanks!

timbru31 (changes requested):

framework/src/org/apache/cordova/PluginEntry.java
erisu: @jblejder would you be able to add in the JavaDoc comment?

You can use the below `public PluginEntry` overloaded method's comment block as an example for your overloaded method.

After this PR is updated, I can merge it in for next release.

erisu (approved):

Issue #1160, opened 2021-02-05
Migrate away from JFrog Bintray / JCenter


kklyoon:
Here is jfrog Notice

https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

should be updated about it. 

Thanks.

breautek:
We are aware, but we don't know what we are going to do about it yet. If you have any recommendations, please do share.

brodybits:
I took the liberty to rewrite the title. As stated in PR #1160, it seems to be blocked by this issue: https://youtrack.jetbrains.com/issue/IDEA-261387

It looks to me like JCenter will continue serving packages until February 1, 2022.

breautek:
cordova-android@9.1 (currently not released, but planned to be released soon...) will contain https://github.com/apache/cordova-android/pull/1179

While this in itself doesn't remove `jcenter`, it should allow you to override the repositories used so that you can use an alternate source, for as long as the repository contains the dependencies that cordova requires. This should serve as a stopgap measure at the very least I think...

th-jtullgren:
Hi,

We need the framework binary to extend some apis in the CordovaInterface. I see the binary in JCenter but nothing in maven central. Is it possible to publish the framework binary to maven central?

Thanks

erisu:
This should have been resolved in Cordova-Android 10.

th-jtullgren:
@erisu Since you have closed this issue, where can I track the progress for uploading the framework to maven central?  Thanks!

th-jtullgren:
@erisu is there an update on this?  Jcenter deadline is approaching.

erisu:
What do you mean?

It has already been completed. We removed the usage and been uploading to maven central.

th-jtullgren:
@erisu I apologize. I just searched for cordova on maven central and it was not showing up. I put the full coordinate and I have found it. Thank you!

erisu:
Awesome! No Problem.

I will go ahead and share the link here so if in case anyone else comes across this ticket and looking for it.

* https://mvnrepository.com/artifact/org.apache.cordova/framework
* https://search.maven.org/artifact/org.apache.cordova/framework

PR #1170, opened 2021-03-05
remove jcenter from gradles

jamesyoon11:
### Platforms affected

Android

### Motivation and Context
JFrog will sunset Bintray, JCenter, GoCenter, and ChartCenter on May 1st 2021 and it will affect cordova-android project. 

### Description

This PR removes jcenter repository from all gradles and replace it with maven. 

### Testing

Create the android project and check if there is no jcenter in gradle files and it is built successfully.
```
npx cordova create SampleApp com.sample.app SampleApp
cd SampleApp/
npx cordova platform add android
```
### Checklist

### DIFF
--- a/bin/templates/project/build.gradle
+++ b/bin/templates/project/build.gradle
@@ -22,7 +22,8 @@ buildscript {
     ext.kotlin_version = '1.3.50'
     repositories {
         google()
-        maven { url 'https://plugins.gradle.org/m2/' }
+        // org.jetbrains.trove4j:trove4j:20160824.
+        maven { url 'https://repo.appspace.com/list/plugins-release/' }
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:4.0.1'
--- a/bin/templates/project/build.gradle
+++ b/bin/templates/project/build.gradle
@@ -22,7 +22,7 @@ buildscript {
     ext.kotlin_version = '1.3.50'
     repositories {
         google()
-        jcenter()
+        maven { url 'https://plugins.gradle.org/m2/' }
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:4.0.1'
@@ -35,7 +35,6 @@ buildscript {
 allprojects {
     repositories {
         google()
-        jcenter()
     }
 
     //This replaces project.properties w.r.t. build settings
--- a/bin/templates/cordova/lib/plugin-build.gradle
+++ b/bin/templates/cordova/lib/plugin-build.gradle
@@ -21,7 +21,6 @@
 buildscript {
     repositories {
         google()
-        jcenter()
     }
 
     // Switch the Android Gradle plugin version requirement depending on the
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -47,7 +47,8 @@ buildscript {
     repositories {
         google()
         mavenCentral()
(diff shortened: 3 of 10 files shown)


bin/templates/project/build.gradle
```diff
@@ -22,7 +22,8 @@ buildscript {
     ext.kotlin_version = '1.3.50'
     repositories {
         google()
-        jcenter()
+        // org.jetbrains.trove4j:trove4j:20160824.
+        maven { url 'https://repo.appspace.com/list/plugins-release/' }
```
brodybits, line 26: Can you explain this a little?

From a quick search it looks to me like trove4j is under LGPL, which we would rather avoid supporting in the core.

I think it would be ideal if this could be specified from a plugin rather than from cordova-android itself.

brodybits:
Thanks for the contribution, I just requested a review from some other maintainers. I wonder if this should be considered a breaking change?

bin/templates/project/build.gradle
jyoon11: Yeah.. Actually, trove4j is used from com.android.tools.build.
```
> Could not find org.jetbrains.trove4j:trove4j:20160824.
     Searched in the following locations:
       - https://dl.google.com/dl/android/maven2/org/jetbrains/trove4j/trove4j/20160824/trove4j-20160824.pom
     If the artifact you are trying to retrieve can be found in the repository but without metadata in 'Maven POM' format, you need to adjust the 'metadataSources { ... }' of the repository declaration.
     Required by:
         project : > com.android.tools.build:gradle:4.0.1 > com.android.tools.build:builder:4.0.1 > com.android.tools:sdk-common:27.0.1
   
```

https://android.googlesource.com/platform/tools/base/+/refs/heads/mirror-goog-studio-master-dev/sdk-common/build.gradle#34

And, it looks like trove4j has not been migrated to mavenCentral or other popular repository yet.

https://youtrack.jetbrains.com/issue/IDEA-261387

So, please let me know if we should wait until Jetbrain resolves it or if there is any concern to use the appspace maven repo.

brodybits: Aargh. So it looks like this is needed by the Android SDK, nothing to do with Cordova. I would favor making the comment more descriptive, like this for example:

    // This Maven URL is needed to find trove4a artifact for Android SDK ref:
    // https://youtrack.jetbrains.com/issue/IDEA-261387

Or we could just wait for them to resolve the issue. It looks like JCenter will not be gone until February 2022 (next year):

https://jfrog.com/blog/into-the-sunset-bintray-jcenter-gocenter-and-chartcenter/

breautek: Yes, I would prefer to wait. I don't like the idea of adding a whole repository just for one package, which could be difficult to remove later.

jyoon11: I agree with you. @breautek I was feeling that it was a little too early to work on this. Let me close it now and may open it again in the future.

jamesyoon11:
Need to wait until Jetbrain migrates trove4j to mavenCentral.

breautek:
> Need to wait until Jetbrain migrates trove4j to mavenCentral.

For reference: https://youtrack.jetbrains.com/issue/IDEA-261387

brodybits:
FYI discussion already started in #1160. I just added "resolves #1160" to the description of this PR. Thanks again!

PR #1173, opened 2021-03-25
feat(android-studio): display app name as project name

### LINKED ISSUE #1172: Add name of the project in .idea/.name for Android Studio
# Feature Request

## Motivation Behind Feature

Opening multiple Cordova Android projects in Android Studio leads to confusion that which project to open because it does not show the name, instead, it shows the path (that too trimmed) and just **android** as shown below-

image

## Feature Description

While we run the command `cordova platform add android` or whenever we run `cordova platform run android` or `cordova platform build android`, after creating the Android project skeleton in `platform/android`, simply create a new file in `.idea/.name` which can have the name of the project (possibility picked from `config.xml`).

Nothing else!

Now, when we open this Android project in Android Studio, it can show the name instead of just **android** as shown in the above screenshot.

As of now, we do this via Cordova (or Ionic) hooks but we have to do it for every single project.

sagrawal31:
### Platforms affected

- Android

### Motivation and Context

Closes #1172 

### Testing

Tried locally but couldn't run (probably I have to read it more carefully).

### Checklist

### DIFF
--- a/bin/lib/create.js
+++ b/bin/lib/create.js
@@ -206,7 +206,7 @@ function validateProjectName (project_name) {
  * https://github.com/apache/cordova-android/issues/1172
  */
 function writeNameForAndroidStudio (project_path, project_name) {
-    var ideaPath = path.join(project_path, '.idea');
+    const ideaPath = path.join(project_path, '.idea');
     fs.ensureDirSync(ideaPath);
     fs.writeFileSync(path.join(ideaPath, '.name'), project_name);
 }
--- a/bin/lib/create.js
+++ b/bin/lib/create.js
@@ -205,7 +205,7 @@ function validateProjectName (project_name) {
  *
  * https://github.com/apache/cordova-android/issues/1172
  */
-function writeNameForAndroidStudio(project_path, project_name) {
+function writeNameForAndroidStudio (project_path, project_name) {
     var ideaPath = path.join(project_path, '.idea');
     fs.ensureDirSync(ideaPath);
     fs.writeFileSync(path.join(ideaPath, '.name'), project_name);
--- a/bin/lib/create.js
+++ b/bin/lib/create.js
@@ -36,6 +36,7 @@ exports.copyScripts = copyScripts;
 exports.copyBuildRules = copyBuildRules;
 exports.writeProjectProperties = writeProjectProperties;
 exports.prepBuildFiles = prepBuildFiles;
+exports.writeNameForAndroidStudio = writeNameForAndroidStudio;
 
 function getFrameworkDir (projectPath, shared) {
     return shared ? path.join(ROOT, 'framework') : path.join(projectPath, 'CordovaLib');
@@ -198,6 +199,19 @@ function validateProjectName (project_name) {
 }
 
 /**
+ * Write the name of the app in "platforms/android/.idea/.name" so that Android Studio can show that name in the
+ * project listing. This is helpful to quickly look in the Android Studio listing if there are so many projects in
+ * Android Studio.
+ *
+ * https://github.com/apache/cordova-android/issues/1172
+ */
+function writeNameForAndroidStudio(project_path, project_name) {
+    var ideaPath = path.join(project_path, '.idea');
+    fs.ensureDirSync(ideaPath);
+    fs.writeFileSync(path.join(ideaPath, '.name'), project_name);
+}
+
+/**
  * Creates an android application with the given options.
  *
  * @param   {String}  project_path  Path to the new Cordova android project.
@@ -294,6 +308,7 @@ exports.create = function (project_path, config, options, events) {
             // Link it to local android install.
             exports.writeProjectProperties(project_path, target_api);
             exports.prepBuildFiles(project_path);
+            exports.writeNameForAndroidStudio(project_path, project_name);
             events.emit('log', generateDoneMessage('create', options.link));
         }).then(() => project_path);
 };
--- a/spec/unit/create.spec.js
+++ b/spec/unit/create.spec.js
@@ -132,6 +132,7 @@ describe('create', function () {
             spyOn(create, 'copyBuildRules');
             spyOn(create, 'writeProjectProperties');
             spyOn(create, 'prepBuildFiles');
+            spyOn(create, 'writeNameForAndroidStudio');
             revert_manifest_mock = create.__set__('AndroidManifest', Manifest_mock);
             spyOn(fs, 'existsSync').and.returnValue(false);
             spyOn(fs, 'copySync');
@@ -298,6 +299,25 @@ describe('create', function () {
                     expect(create.prepBuildFiles).toHaveBeenCalledWith(project_path);
                 });
             });
+        });
+    });
+
+    describe('writeNameForAndroidStudio', () => {
+        var project_path = path.join('some', 'path');
+
+        it('should call ensureDirSync with path', () => {
+            spyOn(fs, 'ensureDirSync');
+            spyOn(fs, 'writeFileSync');
+            create.writeNameForAndroidStudio(project_path, 'Test Android');
+            expect(fs.ensureDirSync).toHaveBeenCalledWith(path.join(project_path, '.idea'));
+        });
+
+        it('should call writeFileSync with content', () => {
+            var appName = 'Test Cordova';
+            spyOn(fs, 'ensureDirSync');
+            spyOn(fs, 'writeFileSync');
+            create.writeNameForAndroidStudio(project_path, appName);
+            expect(fs.writeFileSync).toHaveBeenCalledWith(path.join(project_path, '.idea', '.name'), appName);
         });
     });
 });
--- a/spec/unit/create.spec.js
+++ b/spec/unit/create.spec.js
@@ -303,19 +303,20 @@ describe('create', function () {
     });
 
     describe('writeNameForAndroidStudio', () => {
-        var project_path = path.join('some', 'path');
+        const project_path = path.join('some', 'path');
+        const appName = 'Test Cordova';
 
-        it('should call ensureDirSync with path', () => {
+        beforeEach(function () {
             spyOn(fs, 'ensureDirSync');
             spyOn(fs, 'writeFileSync');
-            create.writeNameForAndroidStudio(project_path, 'Test Android');
+        });
+
+        it('should call ensureDirSync with path', () => {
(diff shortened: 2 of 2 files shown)


breautek:
> Tried locally but couldn't run (probably I have to read it more carefully).

For testing, I'd advise running `npm test`. This will make sure nothing existing broke by the introduction of your change. PR checks will do more extensive testing in multiple environments.

You can also test locally by adding your custom platform by doing `cordova platform add file:/path/to/your-cordova-android-fork` which will install the current version/state of your cordova-android directory.

sagrawal31:
Oh! thanks again @breautek I was looking for that only to create a platform locally.

sagrawal31:
Tried with `cordova platform add file:/path/to/your-cordova-android-fork`.. worked fine 😎  Now running the test.

Pardon my failing test cases, this is my first-time contribution to `cordova-android`.

sagrawal31:
Ran `npm test`, not sure how to take care of this in test files (digging deeper)-

image

breautek:
The tests uses fake paths cause they don't actually create files, they do this by mocking the real implementations of the filesystem.

Looks like the tests will have to be updated. I'd probably would spy on `fs.ensureDirSync` and `fs.writeFileSync` to prevent actual file IO. (in most of those tests we don't actually care if `writeNameForAndroidStudio` because testing that is out of scope)

Then ideally we will have some specific tests that will test the new `writeNameForAndroidStudio` function itself by ensuring those two filesystem api calls are called with the proper parameters. We don't need to test if the implementation of those two FS calls works because we can assume NodeJS / fs-extra plugin authors have done their own testing, so in theory as long as we call them with the proper parameters, then everything **should** be okay :grin:

sagrawal31:
> We don't need to test if the implementation of those two FS calls works because we can assume NodeJS / fs-extra plugin authors have done their own testing

Make sense. Most of the time (white writing JUnit test cases), my mind keep asking questions and then I used to reach to the same conclusion that the author of XYZ must have done their own testing 😄 

Well, adding `spyOn(create, 'writeNameForAndroidStudio');` started to pass the tests. But it's showing some red mark in the last result. Not sure if those are okay-

image

breautek:
> Well, adding spyOn(create, 'writeNameForAndroidStudio'); started to pass the tests. But it's showing some red mark in the last result. Not sure if those are okay-

In jasmine, by default when you spy on a function, it will not call on the original implementation which is why it seemed to fix everything. I think this is okay since we don't need to test `writeNameForAndroidStudio` in every single test case.

But we should have something like the following somewhere:

```javascript
describe('writeNameForAndroidStudio', () => {
   it('should call ensureDirSync with path', () => {
    spyOn(fs, 'ensureDirSync');
    blah.writeNameForAndroidStudio(path); // Whatever what the interface is... 
    expect(fs.ensureDirSync).toHaveBeenCalledWith("thePathValue that is expected to have been passed into ensureDirSync");
  });

  it('should call wrilteFileSync', () => {
    // basically the same for above but with writeFileSync instead.
  });
});
```
That way the `writeNameForAndriodStudio` is also tested.

spec/unit/create.spec.js
```diff
@@ -300,4 +301,23 @@ describe('create', function () {
             });
         });
     });
+
+    describe('writeNameForAndroidStudio', () => {
+        var project_path = path.join('some', 'path');
```
breautek, line 306: In general we want to avoid using `var`, and instead we use `let` or `const` depending on if we expect the variable to change.

```suggestion
        const project_path = path.join('some', 'path');
```

spec/unit/create.spec.js
```diff
@@ -300,4 +301,23 @@ describe('create', function () {
...
+
+        it('should call ensureDirSync with path', () => {
+            spyOn(fs, 'ensureDirSync');
+            spyOn(fs, 'writeFileSync');
+            create.writeNameForAndroidStudio(project_path, 'Test Android');
+            expect(fs.ensureDirSync).toHaveBeenCalledWith(path.join(project_path, '.idea'));
+        });
+
+        it('should call writeFileSync with content', () => {
+            var appName = 'Test Cordova';
```
breautek, line 316: Same as above

```suggestion
            const appName = 'Test Cordova';
```

bin/lib/create.js
```diff
@@ -197,6 +198,19 @@ function validateProjectName (project_name) {
...
 
+/**
+ * Write the name of the app in "platforms/android/.idea/.name" so that Android Studio can show that name in the
+ * project listing. This is helpful to quickly look in the Android Studio listing if there are so many projects in
+ * Android Studio.
+ *
+ * https://github.com/apache/cordova-android/issues/1172
+ */
+function writeNameForAndroidStudio (project_path, project_name) {
+    var ideaPath = path.join(project_path, '.idea');
```
breautek, line 209:
```suggestion
    const ideaPath = path.join(project_path, '.idea');
```

spec/unit/create.spec.js
breautek, line 306: We can reduce some code duplication by using the `beforeEach` test hook, which gets ran before each test.

```suggestion
        beforeEach(() => {
            spyOn(fs, 'ensureDirSync');
            spyOn(fs, 'writeFileSync');
        });
        var project_path = path.join('some', 'path');
```

Then in each individual test inside this describe block, those two methods are already spied on, so we can remove those `spyOn` lines in the individual tests.

spec/unit/create.spec.js
```diff
@@ -300,4 +301,23 @@ describe('create', function () {
             });
         });
     });
+
+    describe('writeNameForAndroidStudio', () => {
+        var project_path = path.join('some', 'path');
+
+        it('should call ensureDirSync with path', () => {
+            spyOn(fs, 'ensureDirSync');
+            spyOn(fs, 'writeFileSync');
```
breautek, line 310: See `beforeEach` usage above.

```suggestion
```

spec/unit/create.spec.js
```diff
@@ -300,4 +301,23 @@ describe('create', function () {
...
+            create.writeNameForAndroidStudio(project_path, 'Test Android');
+            expect(fs.ensureDirSync).toHaveBeenCalledWith(path.join(project_path, '.idea'));
+        });
+
+        it('should call writeFileSync with content', () => {
+            var appName = 'Test Cordova';
+            spyOn(fs, 'ensureDirSync');
+            spyOn(fs, 'writeFileSync');
```
breautek, line 318: See `beforeEach` usage above

```suggestion
```

breautek (commented):
This is really good, also like that you added the comment explaining why we are adding that file, and the PR checks are green which is a good sign!

I just have a few nitpicks, namely the use of `var` keyword and I'd prefer setting up the spies in a `beforeEach` block since I think we'd want those two functions to be always spied on, even in potential future tests that may be added later.

Once these changes are committed, ping me and I'll give the PR a manual test.

spec/unit/create.spec.js
sagrawal31: Well, I love ES6 (and TypeScripe) but when I always try to follow the pattern of existing code and found that existing code is using `var` mostly and using snake case variables, so used `var`. Thanks anyways. Updated this for new code.

sagrawal31: Glad to see the kind of review you made. You are just like me who can put more than 5 comments on the smallest PR where only a few lines are modified 😄 My team some times gets horrified because of this 😄 

Made the code DRY.

spec/unit/create.spec.js
sagrawal31: Made the code DRY.

spec/unit/create.spec.js
sagrawal31: Done.

sagrawal31:
> This is really good, also like that you added the comment explaining why we are adding that file

Thanks for the appreciation!

> and the PR checks are green which is a good sign!

Kudos to you for making it green.

> I just have a few nitpicks, namely the use of var keyword and I'd prefer setting up the spies in a beforeEach block since I think we'd want those two functions to be always spied on, even in potential future tests that may be added later.

Makes a lot of sense. Thanks for pointing this out.

> Once these changes are committed, ping me and I'll give the PR a manual test.

It's done. By the time you will read this comment, yellow dots will be turned green and then you can test locally.

Thanks again for your time.

spec/unit/create.spec.js
breautek: > Well, I love ES6 (and TypeScripe) but when I always try to follow the pattern of existing code and found that existing code is using var mostly and using snake case variables, so used var. Thanks anyways. Updated this for new code.

Yah, `var` refactoring is a work in progress which I think is why we don't have an eslint config (?) I was kinda surprised the eslint didn't produce an error on it tbh.

erisu (approved):
Code LGTM

breautek:
FYI: I edited the main post to include the [GitHub Keyword](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) so that the main issue ticket will be closed when we merge this in.

breautek:
@sagrawal31 I attempted to try this out by pulling down your fork and can't seem to get it to work, but maybe I'm just not understanding something properly.

I'm not seeing the `.idea/.name` file being created... here are the steps I'm taking:

1. `cordova create projectName`
2. `cd projectName`
3. `cordova platform add file:/path/to/your/fork`
4. `cordova build android`

I've confirmed that my copy contains the code from this PR. Am I missing anything?

breautek (approved):
Ignore my message about things aren't working :sweat_smile: I retried the steps and it started working, I probably just accidentally used the wrong cordova-android fork the first time :sweat_smile:

sagrawal31:
> Ignore my message about things aren't working 😅 I retried the steps and it started working, I probably just accidentally used the wrong cordova-android fork the first time 😅

Thanks, no problem 😄  I was about to comment whether you have used the correct branch or not 😃

breautek:
Thank you @sagrawal31 for your time and effort into preparing this PR. I hope we see more in the future. If you're not already, I invite you to our [Slack](http://slack.cordova.io/) community channel and our [Dev Mailing List](https://cordova.apache.org/contact/) where you can stay up to date or contribute to development decisions.

sagrawal31:
Thanks for your mentoring & guidance @breautek. Happy to contribute more.

PR #867, opened 2019-11-16
feat: add custom repositories support

dirien:
### Platforms affected
android

### Motivation and Context
Since the last, two days long, bintray outage, we desperately need a better way to provide alternate custom repositories locations. Especially when you already have a local repository like Nexus or Artifactory.

Every suggestion on StackOverflow was more like a brittle workaround then a solution.

The good thing is, if you don’t need it, just ignore it :)!

### Description
I just seperated the repositories in the existing build.gradle files. 

With the the cordova config.xml and following task:

```
<resource-file src="extra.gradle" target="repositories.gradle" />
<resource-file src="extra.gradle" target="app/repositories.gradle" />
<resource-file src="extra.gradle" target="CordovaLib/repositories.gradle" />
```

we can easily overwrite the existing repositories without the need of a hook.

### Checklist

- [x] I've run the tests to see all new and existing tests pass
- [ ] I added automated test coverage as appropriate for this change
- [ ] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [ ] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [ ] I've updated the documentation if necessary

PR #1179, opened 2021-03-27
feat: add custom repositories support

breautek:
This is to correct a big rebase booboo.

Original PR and credit goes to https://github.com/apache/cordova-android/pull/867

### DIFF
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -44,11 +44,8 @@ buildscript {
         ext.kotlin_version = gradlePluginKotlinVersion
     }
 
-    repositories {
-        mavenCentral()
-        google()
-        jcenter()
-    }
+    apply from: 'repositories.gradle'
+    repositories repos
 
     dependencies {
         apply from: '../CordovaLib/cordova.gradle'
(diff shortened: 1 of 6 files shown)


erisu (approved):

mattdsteele:
@breautek This looks amazing, I am so excited to use this! Is there any documentation or examples of how a to apply a custom registry? It's unclear to me what I should modify, sorry.

dirien:
Hi @mattdsteele , just add "your" repos to the repositories.gradle. It will pick them up from there.

erisu:
@mattdsteele please also read the description of the original PR, which the link can be found in this PR's description. It will contain the `config.xml` the setup.

goffioul:
> As it has been pushed as a minor release, it would probably have been wiser to make the `apply from:` statement in `allprojects` conditional to the existence of `repositories.gradle`.

E.g. a slightly more robust logic for `allprojects` could maybe be:
1. load `repositories.gradle` from current project directory
2. if file in 1. does not exist, load `repositories.gradle` from root project directory

The template project would then only need to provide a single `repositories.gradle` at the top-level and avoid having to copy the same file at multiple locations.

PR #1189, opened 2021-03-31
Revert "feat: upgrade gradle-plugin@4.1.3"

erisu:
Reverts apache/cordova-android#1175

Sorry, it seems some people do not trust that this follows semver and feels it may not fit a minor release.
We will revert this PR and would like it resubmitted for the major release.

### DIFF
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -50,7 +50,7 @@ buildscript {
     dependencies {
         apply from: '../CordovaLib/cordova.gradle'
 
-        classpath 'com.android.tools.build:gradle:4.1.3'
+        classpath 'com.android.tools.build:gradle:4.0.1'
 
         if (cdvHelpers.getConfigPreference('GradlePluginKotlinEnabled', 'false').toBoolean()) {
             classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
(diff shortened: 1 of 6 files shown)


brodybits:
Could we revert these on a 9.x branch to avoid some extra noise in master?

Or would this be more trouble than its worth?

I do suspect we may want something even newer by the time we are ready to make the 10.0.0 release.

erisu:
These are not on the 9.x branch.

erisu:
I have not experienced any issues with this change. I have even been told that one developer had tested the current master (6.1) and had successful builds.

The versioning states that it is a minor version and Gradle does sometimes mention potential breaking changes when upgrading to each latest release of 6.x.

But in each release note for each version, I don't see them mention clearly that there are breaking changes.

It is a little concerning if we can't rely on the versioning system.

* Google has created breaking changes once before in a minor release, they dropped ASL for Android X in the Firebase library.
* TypeScript introduces breaking changes in minor releases. They do not following semver and is on npm (which we suspect would follow semver)

Really, should we start to consider ALL minor releases from all packages, no matter where it comes from, to be a major release for Cordova?

jcesarmobile:
My concern was about the gradle wrapper as it was greater than the one Android Studio recommends, but the gradle plugin version is the one Android Studio recommends, this version should work with gradle wrapper 6.5

erisu:
We will update all Gradle related plugins/libraries/framework stuff in next major and stick with the current versions that are in `rel/9.0.0` and had been approved. This is to remove any concerns and/or issues that can come from upgrading.

jcesarmobile (approved):

breautek (approved):

Issue #1203, opened 2021-04-13
java checks in check_reqs.js is incompatible with "javac -version" output


rmondegar:
# Bug Report

## Problem
check_reqs.js fails to retrieve the correct version java. When checking the version using `javac -version`, `semver.coerce` always returns version "8.0.0".
`javac -version` returns the following output:

![image](https://user-images.githubusercontent.com/60993939/114612178-2d99e100-9c78-11eb-8e12-24abb2ecca6f.png)

`semver.coerce` uses the first number found and returns this as a version:

![image](https://user-images.githubusercontent.com/60993939/114612207-368ab280-9c78-11eb-8eb6-50cac5e0b4cc.png)

With this, the needed check fails:

![image](https://user-images.githubusercontent.com/60993939/114611339-4fdf2f00-9c77-11eb-916a-e94c13a9d9f7.png)

### What is expected to happen?

Return the correct version, in this case, 1.8.0_282

### What does actually happen?

Always returns 8 (related to UTF8)

## Information
It seems the problem was introduced with the commit 774de786910abd0c83e5b5e285961fcb07c38474 where this check was refactored

### Command or Code
`javac -version`
`cordova build -d`

```
import semver = require('semver');
const version ="Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8\njavac 1.8.0_282"
console.log(semver.coerce(version));
```

### Environment, Platform, Device
OS: Ubuntu 18.04.1

### Version information
Cordova: 10.0.0
Plugin list:
```
cordova-hot-code-push-plugin 1.5.3 "Hot Code Push Plugin"
cordova-plugin-appcenter-analytics 0.5.1 "App Center Analytics for Cordova"
cordova-plugin-appcenter-crashes 0.5.1 "App Center Crashes for Cordova"
cordova-plugin-appcenter-shared 0.5.1 "App Center shared code for Cordova"
cordova-plugin-barcodescanner 0.7.4 "BarcodeScanner"
cordova-plugin-camera 5.0.1 "Camera"
cordova-plugin-compat 1.2.0 "Compat"
cordova-plugin-device 1.1.2 "Device"
cordova-plugin-file 6.0.2 "File"
cordova-plugin-geolocation 4.0.2 "Geolocation"
cordova-plugin-inappbrowser 4.0.0 "InAppBrowser"
cordova-plugin-media-capture 3.0.3 "Capture"
cordova-plugin-media 5.0.3 "Media"
[...]

PR #1220, opened 2021-04-26
fix(requirements check): use regex to get java version from javac output

DavidStrausz:
### Platforms affected
Android

### Motivation and Context
Fixes getting the java version from the `javac -version` output on macOS devices when custom `_JAVA_OPTIONS` are set. A detailed description of the issue can be found im my comment [here](https://github.com/apache/cordova-android/pull/1130#issuecomment-824749304).

Fixes https://github.com/apache/cordova-android/issues/1203

### Description
Use a regex to get just the version string from the `javac` output. Fixes #1130.

Previously `semver.coerce()` would get a wrong version from a string like the following:

```js
Picked up _JAVA_OPTIONS: -Xms1024M -Xmx2048M\njavac 1.8.0_271
```

With the change in this PR, the following regex is used to get the version string (`1.8.0`) from the output:

```js
/javac\s+([\d.]+)/i
```

I also tried removing `_JAVA_OPTIONS` from the version check by adding `env: {}, extendEnv: false` to the options of the `javac` command like suggested [here](https://github.com/apache/cordova-android/pull/1130#issuecomment-826084254):

```js
execa('javac', ['-version'], { all: true, env: {}, extendEnv: false })
```

This changed the output from:

```js
Picked up _JAVA_OPTIONS: -Xms1024M -Xmx2048M\njavac 1.8.0_271
```

to:

```js
javac 12.0.2
```

which subsequently let the requirements check fail again (no idea where `12.0.2` comes from). So I just went with the regex solution.

### Testing
- Add android platform with fix on macOS, run `cordova build android`, check if requirements check succeeds
- Add android platform with fix on Windows, run `cordova build android`, check if requirements check succeeds
- Add unit test for `javac -version` output with `_JAVA_OPTIONS`, run unit tests

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/env/java.js
+++ b/bin/templates/cordova/lib/env/java.js
@@ -43,9 +43,9 @@ const java = {
         // Java <= 8 writes version info to stderr, Java >= 9 to stdout
         let version = null;
         try {
-            javacOutput = (await execa('javac', ['-version'], { all: true })).all;
+            const javacOutput = (await execa('javac', ['-version'], { all: true })).all;
 
-            /*  
+            /*
             A regex match for the java version looks like the following:
 
             version: [
@@ -54,7 +54,7 @@ const java = {
                 index: 45,
                 input: 'Picked up _JAVA_OPTIONS: -Xms1024M -Xmx2048M\njavac 1.8.0_271',
                 groups: undefined
-            ] 
+            ]
 
             We have to use a regular expression to get the java version, because on some environments
             (e.g. macOS Big Sur) javac prints _JAVA_OPTIONS before printing the version and semver.coerce()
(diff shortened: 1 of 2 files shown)


breautek:
> I also tried removing _JAVA_OPTIONS from the version check by adding env: {}, extendEnv: false to the options of the javac command like suggested here:

Yah I don't think we can remove the environment variables, as that may contain details like `JAVA_HOME` used to point to the direction where Cordova should look for a path... otherwise it will find the first java variable in your path and call it a day which may not be what we want. ;)

breautek (approved):

breautek:
Will merge by tomorrow night if there are no objections.

PieterVanPoyer (approved):
Looks good. I did not test it.

PR #1222, opened 2021-04-28
feat: add backwards compatibility mode for WebViewAssetLoader

NiklasMerz:
### Platforms affected
Android

### Motivation and Context

As we discussed on the list we might need a fallback for the new changes in #1137 

With this PR you can fall back to using files instead of https://localhost with this preference:
`
<preference name="AndroidInsecureFileModeEnabled" value="true" />
`

Documentation: https://github.com/apache/cordova-docs/pull/1172

### Testing

Manual

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/ConfigXmlParser.java
+++ b/framework/src/org/apache/cordova/ConfigXmlParser.java
@@ -46,9 +46,14 @@ public class ConfigXmlParser {
     }
 
     public String getLaunchUrl() {
-        if (launchUrl == null) { 
+        if (launchUrl == null) {
             launchUrl = "https://" +  this.prefs.getString("hostname", "localhost");
         }
+
+        if (this.prefs.getBoolean("InsecureFileMode", false)) {
+            launchUrl = "file:///android_asset/www/index.html";
+        }
+
         return launchUrl;
     }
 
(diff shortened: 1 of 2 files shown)


erisu (approved):

framework/src/org/apache/cordova/ConfigXmlParser.java
```diff
@@ -46,9 +46,14 @@ public CordovaPreferences getPreferences() {
     }
 
     public String getLaunchUrl() {
-        if (launchUrl == null) { 
+        if (launchUrl == null) {
```
PieterVanPoyer, line 49: Nice work!
But, just a quick question.

There is also a setStartUrl method. The launchUrl is also prefixed with https there.
https://github.com/apache/cordova-android/pull/1222/files#diff-a32677abc0f97a96abd2948fad9d6b12f1251645a9f330431862dc83dabaeae2R150

Should it be prefixed with file if the AndroidInsecureFileModeEnabled returns true?

NiklasMerz: You are right. I just pushed another change with this.

PieterVanPoyer (approved):

dpogue (approved):

jcesarmobile (approved):
tested and works fine, thanks!

Issue #689, opened 2019-03-13
Build errors about drawable since upgrading to v8.0.0


BorntraegerMarc:
# Bug Report

### What is expected to happen?
Building a cordova app with `ionic cordova build android --prod --release` does not throw errors.

### What does actually happen?
I see this error in the console:
```
/platforms/android/app/src/main/res/drawable-land-hdpi/screen.png: Error: The drawable "screen" in drawable-land-hdpi has no declaration in the base drawable folder or in a drawable-densitydpi folder; this can lead to crashes when the drawable is queried in a configuration that does not match this qualifier [MissingDefaultResource]
```

## Information

This error was not present in v7.X

I do specify a splash screen in `config.xml` like so: `<splash density="land-hdpi" src="resources/android/splash/drawable-land-hdpi-screen.png" />`

And it gets displayed correctly (according to my tests so far)

I'm using this plugin to configure the splash screens: https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-splashscreen/

I only included 1 example of the errors I'm getting to keep the bug report small. But I get it for different screen sizes & densities.

I understand that it has something to do with android linting rules: http://tools.android.com/tips/lint-checks

But I don't fully understand if it's a misconfiguration in the plugin or if cordova-android needs to do something differently.

### Command or Code

`ionic cordova build android --prod --release`

### Environment, Platform, Device

On all. It happens on build time.

### Version information

```
~/Documents/komed/komed-frontend$ ionic info

Ionic:

   ionic (Ionic CLI)  : 4.11.0 (/Users/borntsch/.nvm/versions/node/v8.13.0/lib/node_modules/ionic)
   Ionic Framework    : ionic-angular 3.9.2
   @ionic/app-scripts : 3.2.3

Cordova:

   cordova (Cordova CLI) : 8.0.0
   Cordova Platforms     : android 8.0.0, browser 5.0.4, ios 5.0.0
   Cordova Plugins       : cordova-plugin-ionic-webview 2.3.1, (and 27 other plugins)

System:

   Android SDK Tools : 26.1.1 (/Users/borntsch/Library/Android/sdk)
   ios-deploy        : 1.9.2
   ios-sim           : 7.0.0
   NodeJS            : v8.13.0 (/Users/borntsch/.nvm/versions/node/v8.13.0/bin/node)
   npm               : 6.4.1
   OS                : macOS Mojave
   Xcode             : Xcode 10.1 Build version 10B61
```

## Checklist

[...]

PR #1227, opened 2021-05-04
fix(prepare): delete splash screens if none are used

### LINKED ISSUE #1226: Default splash screens end up in APKs even if no splash screens are used
# Bug Report

The default splash screens are copied into the platform folder during the initial end up in release APK even if no splash screens are used by the app. Aside from about a megabyte of unnecessary clutter in the APK, this triggers the warnings described in #689 during release builds.

Splashs should not be copied to `platforms/` if they are not used.

### Command or Code

This is reproducible using the default app:
1. `cordova create foo && cd foo`
2. `cordova platform add android`

If you run `du -ch platforms/android/app/src/main/res/*/screen.png` now, you will see the default splashs listed. Note that there are not splashs defined (cordova also emits a verbose log message about that during prepare: `This app does not have splash screens defined`). If you proceed to build those files will end up in the APK and cause aforementioned warnings if it is a release build.

### Version information
```
Cordova Packages:
    cli: 10.0.0
        common: 4.0.2
        create: 3.0.0
        lib: 10.0.0
            common: 4.0.2
            fetch: 3.0.1
            serve: 4.0.0

Project Installed Platforms:
    android: 9.1.0

Project Installed Plugins:
    cordova-plugin-whitelist: 1.3.4

Environment:
    OS: Ubuntu 20.04.2 LTS (linux 5.4.0-72-generic) x64
    Node: v14.16.1
    npm: 7.11.2

```

## Checklist

raphinesse:
### Motivation and Context

We copy default splash screens to the platform folder during platform creation. But then we do not delete them when the user uses no splash screens at all. This causes #1226 and what is described in #689.

Fixes #1226

### Description

The new implementation does not return early when updating splashes and none are defined. Instead, we let the cleanup map we create initially take care of deleting all unused splash screens. This also deletes the default splashes during the first prepare after platform creation.

### Testing

Manual testing on a default app.
A regression test would be great but I'm on a tight schedule. Feel free to add one if you feel like it.

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/prepare.js
+++ b/bin/templates/cordova/lib/prepare.js
@@ -334,7 +334,7 @@ function updateSplashes (cordovaProject, platformResourcesDir) {
         // We must not return here!
         // If the user defines no splash screens, the cleanup map will cause any
         // existing splash screen images (e.g. the defaults that we copy into a
-        // new app) to be removed the app folder, which is what we want.
+        // new app) to be removed from the app folder, which is what we want.
     }
 
     // Build an initial resource map that deletes all existing splash screens
--- a/bin/templates/cordova/lib/prepare.js
+++ b/bin/templates/cordova/lib/prepare.js
@@ -328,10 +328,13 @@ function makeSplashCleanupMap (projectRoot, resourcesDir) {
 function updateSplashes (cordovaProject, platformResourcesDir) {
     var resources = cordovaProject.projectConfig.getSplashScreens('android');
 
-    // if there are "splash" elements in config.xml
+    // if there are no "splash" elements in config.xml
     if (resources.length === 0) {
         events.emit('verbose', 'This app does not have splash screens defined');
-        return;
+        // We must not return here!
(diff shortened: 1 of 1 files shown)


breautek (approved):

bin/templates/cordova/lib/prepare.js
```diff
@@ -328,10 +328,13 @@ function makeSplashCleanupMap (projectRoot, resourcesDir) {
...
 
-    // if there are "splash" elements in config.xml
+    // if there are no "splash" elements in config.xml
     if (resources.length === 0) {
         events.emit('verbose', 'This app does not have splash screens defined');
-        return;
+        // We must not return here!
+        // If the user defines no splash screens, the cleanup map will cause any
+        // existing splash screen images (e.g. the defaults that we copy into a
+        // new app) to be removed the app folder, which is what we want.
```
raphinesse, line 337:
```suggestion
        // new app) to be removed from the app folder, which is what we want.
```

jcesarmobile (approved):

Issue #1221, opened 2021-04-27
Cordova isn't detecting java 8 jdk when a more recent version is installed


ldesroch:
Cordova isn't detecting java 8 jdk when a more recent version is installed.
Even when the JAVA_HOME env var is properly set :

```
PS C:\Users\***\workspace\prestas\***\cordova-poc\in-app-browser> cordova run android
Checking Java JDK and Android SDK versions
ANDROID_SDK_ROOT=C:\Users\***\AppData\Local\Android\Sdk\platforms\android-30 (recommended setting)
ANDROID_HOME=C:\Users\***\AppData\Local\Android\Sdk (DEPRECATED)
Requirements check failed for JDK 1.8.x! Detected version: 15.0.1
Check your ANDROID_SDK_ROOT / JAVA_HOME / PATH environment variables.
PS C:\Users\***\workspace\prestas\***\cordova-poc\in-app-browser> $Env:ANDROID_SDK_ROOT
C:\Users\***\AppData\Local\Android\Sdk\platforms\android-30
PS C:\Users\***\workspace\prestas\***\cordova-poc\in-app-browser> $Env:JAVA_HOME
C:\Program Files\Java\jdk1.8.0_291
```

Cordova version : 10.0.0

breautek:
What's your `cordova-android` version? (`cordova platform ls`)

ldesroch:
9.1.0 I guess

```> cordova platform ls
6.0.0
Installed platforms:
  android 9.1.0
  browser
  ios 6.2.0
Available platforms:
  browser ^6.0.0
  electron ^1.0.0
  windows ^7.0.0```

**Comment by breautek** [comment]
Thanks for your response,

9.1 introduced some refactoring changes around java detection, can you confirm if you see the same behaviour on `cordova-android@9.0`?

This will tell us if this was a new problem introduced as part of that refactoring, or if it's just simply an older problem.

**Comment by mcfarljw** [comment]
I don't know if it's related or not but a few days ago there was a big Android Studio update and that is when mine stopped working as well. Mine is spouting a different detected version as I was previously using whatever they provided.

```
export JAVA_HOME="/Applications/Android Studio.app/Contents/jre/jdk/Contents/Home"
export ANDROID_SDK_ROOT="$HOME/Library/Android/sdk"
export PATH="$ANDROID_SDK_ROOT/platform-tools:$PATH"
export PATH="$ANDROID_SDK_ROOT/tools/bin:$PATH"

[...]

PR #1228, opened 2021-05-09
Bugfix/java checks

breautek:
### Platforms affected

Android

### Motivation and Context

Fixes https://github.com/apache/cordova-android/issues/1221
~Fixes https://github.com/apache/cordova-android/issues/1203~ (Addressed by https://github.com/apache/cordova-android/pull/1220)

### Description

This addresses two flaws in Java checking.

1. ~If some Java options are present, `javac -version` will print out additional lines of information. A regex is used to find the `java(c) x.x.x` line for proper version checking.~ Addressed by https://github.com/apache/cordova-android/pull/1220
2. If `JAVA_HOME` is set, we would only set the path if `javac` wasn't findable, and when we do, we set the path at the end of the `PATH` chain. To be smarter, if `JAVA_HOME` is set, we **always** want to set the `JAVA_HOME` bin path to the front of the `PATH` chain so that we can guarantee the proper java executable to be used.

### Testing

Added new test for the `javac -version` issue. All existing tests passes. Manual testing for the `JAVA_HOME` issue.

### Checklist

### DIFF
--- a/bin/templates/cordova/lib/env/java.js
+++ b/bin/templates/cordova/lib/env/java.js
@@ -79,10 +79,6 @@ const java = {
             throw new CordovaError(msg);
         }
 
-        // This parses out additional lines that
-        // the Java executable may print out
-        version = /^javac? .+$/im.exec(version)[0];
-
         return semver.coerce(version);
     },
 
@@ -97,7 +93,6 @@ const java = {
             return;
         }
 
-        const javacPath = utils.forgivingWhichSync('javac');
         const javaHome = environment.CORDOVA_JAVA_HOME || environment.JAVA_HOME;
         if (javaHome) {
             // Ensure that CORDOVA_JAVA_HOME overrides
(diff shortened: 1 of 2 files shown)


bin/templates/cordova/lib/env/java.js
```diff
@@ -96,10 +96,9 @@ const java = {
         const javacPath = utils.forgivingWhichSync('javac');
         const hasJavaHome = !!environment.JAVA_HOME;
         if (hasJavaHome) {
-            // Windows java installer doesn't add javac to PATH, nor set JAVA_HOME (ugh).
-            if (!javacPath) {
```
PieterVanPoyer, line 100: if javacPath is not used here.

The scope of the const can be smaller. Maybe place the `` const javacPath = utils.forgivingWhichSync('javac');`` after the else { (Line 103)?

PieterVanPoyer (approved):
Great!

PR #1229, opened 2021-05-09
feat: CORDOVA_JAVA_HOME env variable

breautek:
### Platforms affected

Android

### Motivation and Context

It's been a requested feature (not formally, but in variety of bug tickets of people trying to use java >8) because often times, we are stuck on using older Java versions
due to the Android SDK. This will make it easier for users to have multiple java versions installed.

### Description

If `CORDOVA_JAVA_HOME` is present, it will override `JAVA_HOME` environment variable for the execution context of Cordova.

### Testing

Unit test added. All existing tests passes. Manual testing.

### Checklist

### DIFF
--- a/spec/unit/java.spec.js
+++ b/spec/unit/java.spec.js
@@ -76,6 +76,19 @@ describe('Java', () => {
             Java.__set__('javaIsEnsured', false);
         });
 
+        it('CORDOVA_JAVA_HOME overrides JAVA_HOME', async () => {
+            spyOn(utils, 'forgivingWhichSync').and.returnValue('');
+
+            const env = {
+                CORDOVA_JAVA_HOME: '/tmp/jdk'
+            };
+
+            await Java._ensure(env);
+
+            expect(env.PATH.split(path.delimiter))
+                .toContain(path.join(env.JAVA_HOME, 'bin'));
+        });
+
         it('with JAVA_HOME / without javac', async () => {
             spyOn(utils, 'forgivingWhichSync').and.returnValue('');
 
--- a/spec/unit/java.spec.js
+++ b/spec/unit/java.spec.js
@@ -85,8 +85,8 @@ describe('Java', () => {
 
             await Java._ensure(env);
 
-            expect(env.PATH.split(path.delimiter))
-                .toContain(path.join(env.JAVA_HOME, 'bin'));
+            expect(env.JAVA_HOME).toBe('/tmp/jdk');
+            expect(env.PATH.split(path.delimiter)).toContain('/tmp/jdk/bin');
         });
 
         it('with JAVA_HOME / without javac', async () => {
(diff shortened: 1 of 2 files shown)


spec/unit/java.spec.js
```diff
@@ -76,6 +76,19 @@ describe('Java', () => {
...
+            spyOn(utils, 'forgivingWhichSync').and.returnValue('');
+
+            const env = {
+                CORDOVA_JAVA_HOME: '/tmp/jdk'
+            };
+
+            await Java._ensure(env);
+
+            expect(env.PATH.split(path.delimiter))
+                .toContain(path.join(env.JAVA_HOME, 'bin'));
```
PieterVanPoyer, line 89: This is maybe nitpicking. (Feel free to ignore or comment it, off course )
But I expect the test to be without env.JAVA_HOME . 
I expect it more like
```
expect(env.PATH.split(path.delimiter)).toContain(path.join('/tmp/jdk', 'bin'));
```
You know exactly what it should be. Assert the exact value.
Now, when someone changes the business logic around JAVA_HOME and doesn't use the CORDOVA_JAVA_HOME in the correct way, the test could still succeed.

PieterVanPoyer (approved):
I left a comment, but it is nitpicking, so feel free to look at it, and address it, ignore it, or give your opinion about my comment.

Minor, and discussable off course.

spec/unit/java.spec.js
breautek: Nope, I think I understand what you mean, this test could be improved. Let me know what you think about my change. I added an explicit assertion for `JAVA_HOME` as we want to ensure `JAVA_HOME` is set for our process (so that sub processes can be set appropriately), and changed it to be hard-coded `/tmp/jdk` as that is our chosen test value.

PieterVanPoyer (approved):
Awesome stuff!

breautek:
Thanks for the review @PieterVanPoyer 

Just going to push a new commit to fix the test faliure that occurs windows (stupid path separator differences....)

paulwwisl:
This is not documented in older v10 https://cordova.apache.org/docs/en/10.x/guide/platforms/android/index.html
or newer v11 https://cordova.apache.org/docs/en/11.x/guide/platforms/android/index.html

but only in v10 release notes
https://cordova.apache.org/announcements/2021/07/20/cordova-android-10.0.0.html

paulwwisl:
As of cordova 11.1 CORDOVA_JAVA_HOME is silently being used over JAVA_HOME

$ cordova --version
11.1.0
MBP-pverest:cordovaapp1 pverest$ cordova requirements

Requirements check results for android:
Java JDK: installed 11.0.15
Android SDK: installed true
Android target: installed android-33-ext5,android-33,android-32,android-31,android-30
Gradle: installed /opt/homebrew/Cellar/gradle/8.0.2/bin/gradle


ANDROID_HOME=/Users/pverest/Library/Android/sdk
ANDROID_SDK_ROOT=/Users/pverest/Library/Android/sdk
CORDOVA_JAVA_HOME=/Applications/Android Studio.app/Contents/jbr/Contents/Home
...
JAVA_HOME=/Users/pverest/Library/Java/JavaVirtualMachines/corretto-17.0.6-1/Contents/Home

PR #1233, opened 2021-05-10
feat: Upgrade gradle to 7.0

jcesarmobile:
### Checklist

### DIFF
--- a/bin/templates/cordova/lib/builders/ProjectBuilder.js
+++ b/bin/templates/cordova/lib/builders/ProjectBuilder.js
@@ -268,7 +268,7 @@ class ProjectBuilder {
                 // update/set the distributionUrl in the gradle-wrapper.properties
                 const gradleWrapperPropertiesPath = path.join(self.root, 'gradle/wrapper/gradle-wrapper.properties');
                 const gradleWrapperProperties = createEditor(gradleWrapperPropertiesPath);
-                const distributionUrl = process.env.CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL || 'https://services.gradle.org/distributions/gradle-6.8.3-all.zip';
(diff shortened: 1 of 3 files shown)


erisu:
IMO, for this major release, upgrading to Gradle 7.x is not recommended.

Following Android documetation and the versioning system, the current released version of Android Gradle Plugin does not support Gradle 7.

https://developer.android.com/studio/releases/gradle-plugin#updating-gradle

* If we upgraded to 7.x can it be guaranteed that we will never face issues caused by using an unsupported version?
* Is there any other documentation provided by Android that contradicts the above page where tehy say 7.x is supported?

jcesarmobile:
It is supported, that table shows the minimum version, it says 6.7.1+
we updated Capacitor to gradle 7 without problems (on the project itself, but plugins or apps can still break depending on their gradle scripts if something they use was removed)

I'm ok if it doesn't get merged, the fact that tests didn't pass means some of our scripts don't work on gradle 7 and need a rewrite.

erisu:
Per Gradle Docs:

https://docs.gradle.org/current/userguide/single_versions.html

> A prefix version range: e.g. 1.+, 1.3.+
> Only versions exactly matching the portion before the + are included.
> The range + on it’s own will include any version.

`6.7.1+` means onlys `6.7.x` is supported, not even the minor.
> > Only versions exactly matching the portion before the + are included.

jcesarmobile:
That is gradle docs, while the other are android docs, so I wouldn't trust google uses that system 😛 

Gradle site says they have tested gradle 7 on gradle plugin 4.1, 4.2 and 7 alpha/beta

https://docs.gradle.org/7.0/userguide/compatibility.html#android

jcesarmobile:
anyway, that would mean we need to downgrade as the PR is updating from 6.8.3 to 7.0.0, and you say we should be using 6.7.1

PR #1241, opened 2021-06-11
feat: remove java 1.8 version check

jcesarmobile:
This is another approach to the java versions handling/requirement proposed here https://github.com/apache/cordova-android/pull/1230

At the moment we require java 8, the linked PR will require java 11, but I think we don't really need to require any specific version (or well, maybe >= java 8, but I don't think people is using java 7 or older at this point).

So this PR just removes the java version check (it still checks that java is installed, but just "ignores" which version, as long as java is installed any version should be fine)

At the moment I have java 11 installed because I use the one that Android Studio includes and Android Studio 4.2 started using java 11 instead of java 8, but I can still run cordova apps from the CLI that supposedly require java 8, and I think that java versions are backward compatible and if an user had java 16 it should also work (but didn't try).

Even if we want to require java 11 and go with the linked PR, I think we should change the version check to >= 11 and not a strict 11, because some people work with java for server/desktop apps and use newer java versions and are forced to downgrade to use Cordova, when it looks like it's not really necessary.


erisu (approved):
This PR will require a rebase, but I had confirmed that the changes will work with current main branch.

I created a seperate branch & cherry-picked this commit for testing: https://github.com/erisu/cordova-android/tree/feat/java-11

**Build Output:**
https://gist.github.com/erisu/d6ee1859d3fecd1b9e8af1e5769b58e3

**Gradle Wrapper Version Output:**

```
$ ./platforms/android/gradlew -v

------------------------------------------------------------
Gradle 7.1.1
------------------------------------------------------------

Build time:   2021-07-02 12:16:43 UTC
Revision:     774525a055494e0ece39f522ac7ad17498ce032c

Kotlin:       1.4.31
Groovy:       3.0.7
Ant:          Apache Ant(TM) version 1.10.9 compiled on September 27 2020
JVM:          11.0.11 (AdoptOpenJDK 11.0.11+9)
OS:           Mac OS X 11.4 x86_64
```

I also confirm that JDK 1.8 is not affected from the changes.

PR #1260, opened 2021-07-06
refactor(check_reqs): drop originalError param from check_android_target

raphinesse:
This refactor removes a strange optional parameter from `check_reqs.check_android_target` in favor of having the sole user of that parameter handle the related logic itself.


erisu (approved):
LGTM

PR #1265, opened 2021-07-07
refactor: do not infer project root from script location

raphinesse:
### Motivation and Context

The goal of this PR is to eliminate all areas where our JS library expects to be located inside the project directory.

This is to prepare for a change that will keep the library in node_modules and not copy it to the platform project.

### Description

Basically this passes the project root that has been passed to `Api` to all places that need it but do not yet have it.

This also extends the E2E tests to check that we do not have to require the Api from the platform project folder.

### Testing

- Updated tests pass
- Manually built default app with updated platform


erisu (approved):
LGTM

PR #1269, opened 2021-07-13
refactor!: do not copy JS lib to platform project

raphinesse:
### Motivation and Context

Finally, the moment we've all been waiting for! :tada: :wink: 

This PR does away with the questionable practice of copying the entire JS lib of `cordova-android` into each generated platform project – which had caused several issues and complications in the past.

### Description

- create: do not copy our JS lib (or `node_modules`) to platform project
- Move files around
  - `bin/templates/cordova/lib/` to `lib/`
  - `bin/templates/cordova/Api.js` to `lib/`
  - `bin/lib/create.js` to `lib/`
  - `bin/templates/cordova/lib/plugin-build.gradle` to `lib/builders`

After these changes, the `cordova/` folder of a generated project only contains the bare minimum to maintain compatibility with existing cordova CLI versions (plus `defaults.xml` which IMHO should be moved in another PR). All of the remaining JS files in that folder require their dependencies via the `cordova-android` module, so that needs to be resolvable somehow (which is the case for a platform project generated by the regular CLI workflow).

### Testing

- existing tests pass
- I successfully created and built a cordova-android project using cordova@10


erisu (approved):
* The changes LGTM
* I confirmed that the tests are passing.
* I created a new cordova app & added the platform.
* I built the default project

PR #1272, opened 2021-07-13
cleanup: delete old ANT & Eclipse files

erisu:

Remove old ANT/Eclipse related files.

### Testing

- `npm t`

### Checklist


raphinesse (changes requested):
This probably should also remove the VERSION file. Seems it is only referenced from `build.xml`

Other than that: would be great to get rid of these seemingly long obsolete files.

erisu:
I will create a separate PR for removing the VERSION file.

I think that the file was not from ANT nor Eclipse. I think it was something we created to easily manage the cordova-android version in one location and use it in places like ANT. I think it was apart of the pre-npm registry era.

Even through after this PR is merged, there should be no references to this file. I just want to make sure we can easily revert the VERSION drop if there was something unexpected that was not in our radar.

As for ANT & Eclipse, these are outdated tools that isnt genrally used or prefered for Android development/release anymore.

raphinesse (approved):

jcesarmobile (approved):

dpogue (approved):

PR #1138, opened 2020-12-01
feat(allow-list): integrate and refactor core plugin

erisu:
### Motivation, Context & Description

Integrate the Allow List core plugin to platform.

### Testing

- `npm t`
- `cordova create`
- `cordova platform add`
- `cordova build android`

Run in emulator though Android Studio.

* Must remove the original plugin that is configured with new projects.

### Checklist

- [x] I've run the tests to see all new and existing tests pass
- [x] I've updated the documentation if necessary

PR #1275, opened 2021-07-14
fix: add WebViewAssetloader to default allow list

### LINKED ISSUE #1217: Application does not initialize correctly on 10.x nightly/dev master
# Bug Report

Application to start correctly

Application fails to initialize probably (deviceReady not fired).

## Information

This is a regression from a newly added feature https://github.com/apache/cordova-android/pull/1137

With the default configuration, plugins will fail to initialise because they do not pass the whitelist checks. This occurs when you have at least one plugin installed.

The default `<content src="index.html" />` seems to produce a white list pattern where `index.html` is the `host`, instead of the `path`, so when comparing against the `https://localhost` origin, it returns false.

Screenshot from 2021-04-24 12-36-24

### Command or Code

```
cordova create test
cd test
cordova platform add android@nightly
cordova plugin add cordova-plugin-file
cordova run android
```

These steps will produce the hello world cordova app that stays on "connecting" because `deviceReady` never fires.

`cordova-android@nightly`
And one other plugin

## Workaround

Setting `<content src="https://localhost/index.html" />` will correct the problem, but the framework should perhaps figure out this out so that `<content src="index.html" />` continues to work properly.

## Checklist

NiklasMerz:
### Platforms affected
Android

### Motivation and Context

Addresses issues in #1217 

Closes #1217

### Description

With the introduction of the WebViewAssetloader #1137 and the Allow List Plugin #1138 we need to add the origin / href the app is running to the allowlist.

### Testing

Create a blank new app with one plugin

```
cordova create test
cd test
cordova platform add https://github.com/apache/cordova-android#webviewassetloader-allowlist
cordova plugin add cordova-plugin-file
cordova run android
```

These steps will produce the hello world cordova app that should show "deviceready" with this pr. See issue.

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/AllowListPlugin.java
+++ b/framework/src/org/apache/cordova/AllowListPlugin.java
@@ -84,7 +84,9 @@ public class AllowListPlugin extends CordovaPlugin {
                 allowedNavigations.addAllowListEntry(startPage, false);
 
                 // Allow origin for WebViewAssetLoader
-                allowedNavigations.addAllowListEntry("https://" + this.prefs.getString("hostname", "localhost") + "/" + startPage, false);
+                if (!this.prefs.getBoolean("AndroidInsecureFileModeEnabled", false)) {
+                    allowedNavigations.addAllowListEntry("https://" + this.prefs.getString("hostname", "localhost"), false);
+                }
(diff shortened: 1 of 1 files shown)


erisu (approved):
LGTM

Issue #1284, opened 2021-07-19
Unknown property gradlePluginGoogleServicesVersion


BmsVisionJo:
# Bug Report

## Problem
When enabling the GradlePluginGoogleServicesEnabled preference the gradle build fails with the following message:
* Where:
Build file 'D:\Repos\MyMES\mmMobileApp\platforms\android\app\build.gradle' line: 49

* What went wrong:
A problem occurred evaluating project ':app'.
> Could not get unknown property 'gradlePluginGoogleServicesVersion' for object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.

### What is expected to happen?
Use the correct property so the build can complete

### What does actually happen?
This code has been changed to be able to pass in the config defined by the preferences.
I think that it is just a forgotten property that needs changing.
Replacing line 49 from:
 if(!cdvHelpers.isVersionValid(gradlePluginGoogleServicesVersion)) {
to:
 if(!cdvHelpers.isVersionValid(cordovaConfig.GRADLE_PLUGIN_GOOGLE_SERVICES_VERSION)) {
 fixes the issue

### Command or Code
Cordova build

### Version information
Cordova android 10.0.0-nightly.2021.7.17.92864963

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

PR #1287, opened 2021-07-20
fix: Google Services Gradle Plugin version check failure

erisu:
### Motivation and Context

fixes: #1284

Build failure when Google Services Gradle Plugin is enabled. It fails because the version checking is calling a variable that was refactored and forgotten.

### Description

Changed old variable name `gradlePluginGoogleServicesVersion` to new variable name `cordovaConfig.GRADLE_PLUGIN_GOOGLE_SERVICES_VERSION`

### Testing

- `npm t`

### Checklist

### DIFF
--- a/templates/project/app/build.gradle
+++ b/templates/project/app/build.gradle
@@ -46,7 +46,7 @@ buildscript {
 
         if(cordovaConfig.IS_GRADLE_PLUGIN_GOOGLE_SERVICES_ENABLED) {
             // Checks if the kotlin version format is valid.
-            if(!cdvHelpers.isVersionValid(gradlePluginGoogleServicesVersion)) {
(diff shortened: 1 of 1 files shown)


pataar:
After applying this fix, I'm still getting errors related to the google-services plugin.

image

erisu:
@pataar 

The error your reporting has nothing to do with the version that is set as default nor the flag is being fixed in this PR.

A plugin that your using appears to require a different version and causing your issues. You should review the plugin requirements and set the version based off what the plugin is requesting.

When I tested the changes from this PR, the Google Services Gradle plugin classpaths is checked out successfuly and build successfully.

pataar:
@erisu Yes, upgrading the plugin worked.
Made an issue here: https://github.com/havesource/cordova-plugin-push/issues/108

PR #1294, opened 2021-07-26
fix: automatic latest build tools finding

breautek:
### Platforms affected
Android

### Motivation and Context

Currently, android will attempt to pull in the latest build tools version available on the system, regardless of major versions.
This is problematic because we test against a specific major, and future majors may have breaking changes resulting in an inconsistent build.

### Description

`doFindLatestInstalledBuildTools` will now limit it's scope to the minimum version major. This means that if our minimum required version is `30.0.3`, it will filter everything that is `>= 31.0.0`.

Likewise, in the future when we support v31, if our minimum version was `31.0.0`, it will filter everything `>= 32.0.0`, etc.

### Testing

Ran npm test successfully.

Manually tested with:
- build tools v31 installed, 30.0.3 uninstalled, 30.0.2 installed, errors as expected (minimum version is 30.0.3)
- build tools v31 installed, 30.0.3 installed, 30.0.2 installed, rans successfully, using build tools version 30.0.3.

### Checklist

### DIFF
--- a/framework/cordova.gradle
+++ b/framework/cordova.gradle
@@ -53,10 +53,13 @@ String doFindLatestInstalledBuildTools(String minBuildToolsVersionString) {
         throw e
     }
 
+    def minBuildToolsVersion = new Version(minBuildToolsVersionString)
+    def maxVersion = new Version((minBuildToolsVersion.getMajor() + 1) + ".0.0")
+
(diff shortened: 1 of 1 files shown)


erisu (approved):
LGTM

Issue #1296, opened 2021-07-27
forced https:// scheme breaks http and websocket APIs


phatpaul:
# Bug Report

## Problem

New forced https:// scheme is incompatible with unsecure APIs.

I think there should be an option to use http:// scheme. 
 (file:// scheme works OK on newer Android devices, but causes issues with loading ES6 modules on older devices)

### What is expected to happen?
Upgraded Cordova should continue to connect to devices over http: and ws:

### What does actually happen?
Unable to connect to unsecure HTTP API or websocket.
```
Error: DOMException: Failed to construct 'WebSocket': An insecure WebSocket connection may not be initiated from a page loaded over HTTPS.
Mixed Content: The page at 'https://localhost/index.html' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://192.168.33.144/websocket/ws.cgi'. This request has been blocked; this endpoint must be available over WSS.
```

### Command or Code to reproduce issue
Upgrade cordova-android platform to 10.0.

```
ws = new WebSocket("ws://192.168.33.144")
Mixed Content: The page at 'https://localhost/index.html' was loaded over HTTPS, but attempted to connect to the insecure WebSocket endpoint 'ws://192.168.33.144/'. This request has been blocked; this endpoint must be available over WSS.
VM1019:1
DOMException: Failed to construct 'WebSocket': An insecure WebSocket connection may not be initiated from a page loaded over HTTPS.
```

```
fetch("http://192.168.33.144")
Mixed Content: The page at 'https://localhost/index.html' was loaded over HTTPS, but requested an insecure resource 'http://192.168.33.144/'. This request has been blocked; the content must be served over HTTPS.
```

### Environment, Platform, Device
Cordova app is connecting to a ESP32 webserver via encrypted WiFi.  (not on the internet)  There shouldn't be a need to upgrade the ESP32 firmware to support HTTPS (and even if I did, would a self-signed SSL cert work?)

### Version information
Cordova: Cordova 10.0
Cordova Platforms: Cordova-Android 10.0
Other Frameworks: none.
Operating System: Win 7 x64
Android Studio 4.2.2
Android 10 on Moto Stylus G

## Checklist

- [x ] I searched for existing GitHub issues
- [x ] I updated all Cordova tooling to most recent version
- [x ] I included all the necessary information above

NiklasMerz:
[...]

Issue #1289, opened 2021-07-21
AndroidInsecureFileModeEnabled breaks config.xml content src


ctayl:
When using `<preference name="AndroidInsecureFileModeEnabled" value="true" />`
the `<content/>` tag's `src` is set to a hardcoded value: `file:///android_asset/www/index.html`. 

https://github.com/apache/cordova-android/blob/015db819aed3f35892d2b6035781ea4bd752149d/framework/src/org/apache/cordova/ConfigXmlParser.java#L54

coolscene:
To add on a similar issue, when "AndroidInsecureFileModeEnabled" is set to false or not set at all, the "hostname" preference is ignored and always tries to run what's set in the content tag (eg: https://localhost/index.html).

PR #1298, opened 2021-07-28
feat: support http w/ content src fix

erisu:
### Motivation and Context

fixes: #1296
fixes: #1289
fixes: #1299

### Description

`setStartUrl` was being called during the parsing of the `config.xml`. This should not be called until after the parsing has been completed. The reason why this was an issue is because `setStartUrl` was called when the `content` tag was being parsed. Some or all `preference` tags may not have been already loaded, so therefore we do not have the information to best determin what the start url should be.

This PR moves the `setStartUrl` in side the `getLaunchUrl`. The first call of `getLaunchUrl` will trigger the `setStartUrl` method. Any calls there after will return immediately the launchUrl as it will be saved from the first call.

With the changes described above, it resolved the reported issue in #1289.

Additionally, this PR allows users to set the scheme between `http` and `https`. These are the only valid schemes for the `WebViewAssetLoader`. If the user supplied anything that is not valid, it will default to `https`

Also to continue to maintain a best secure pratice, the default with no defined `scheme` will also be `https`.

### Testing

- `npm t`
- `cordova build`
- Tested Use Cases
  - `scheme`+`hostname`+`content:src=<file>` Test
    - `http`, `localhost`, `index.html`
    - `http`, `localhost`, `test/index.html`
    - `http`, `localhost`, `page.html`
    - `https`, `localhost`, `index.html`
    - `https`, `localhost`, `test/index.html`
    - `https`, `localhost`, `page.html`
    - `http`, `foobar`, `index.html`
    - `http`, `foobar`, `test/index.html`
    - `http`, `foobar`, `page.html`
    - `https`, `foobar`, `index.html`
    - `https`, `foobar`, `test/index.html`
    - `https`, `foobar`, `page.html`
  - `AndroidInsecureFileModeEnabled=true` (file://) Test
    - `index.html`
    - `test/index.html`
    - `page.html`
  -  `content:src=<URL>`
    - `https://cordova.apache.org/`

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/ConfigXmlParser.java
+++ b/framework/src/org/apache/cordova/ConfigXmlParser.java
@@ -24,16 +24,19 @@ import java.util.ArrayList;
 import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-
 import org.xmlpull.v1.XmlPullParser;
 import org.xmlpull.v1.XmlPullParserException;
-
 import android.content.Context;
 
 public class ConfigXmlParser {
     private static String TAG = "ConfigXmlParser";
 
-    private String launchUrl = null;
+    private static String SCHEME_HTTP = "http";
+    private static String SCHEME_HTTPS = "https";
+    private static String DEFAULT_HOSTNAME = "localhost";
+
+    private String launchUrl;
+    private String contentSrc;
     private CordovaPreferences prefs = new CordovaPreferences();
     private ArrayList<PluginEntry> pluginEntries = new ArrayList<PluginEntry>(20);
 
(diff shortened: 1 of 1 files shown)


breautek (approved):
lgtm

framework/src/org/apache/cordova/ConfigXmlParser.java
```diff
@@ -147,20 +149,40 @@ public void handleEndTag(XmlPullParser xml) {
...
+                    "(Valid Options=" + SCHEME_HTTP + "," + SCHEME_HTTPS + ")");
+
+                scheme = SCHEME_HTTPS;
+            }
+
+            return scheme + "://" + hostname + '/';
+        }
+    }
+
     private void setStartUrl(String src) {
```
AdriVanHoudt, line 171: Just for someone not familiar with this code, I would rename this to `setLaunchUrl`.

AdriVanHoudt (commented):
This seems to fix https://github.com/apache/cordova-android/issues/1299.

Added a small nit but lgtm otherwise (I can only vouch for the workings regarding https://github.com/apache/cordova-android/issues/1299 not scheme stuff :P)

PR #1300, opened 2021-07-29
fix: fall back to project root repositories.gradle

erisu:
### Motivation and Context

Not all subprojects will have the `repositories.gradle` so therefore trying to apply `repositories.gradle` will fail.

### Description

This PR updates the logic to check if the `repositories.gradle` exists and if it does not, fallback and use the `repositories.gradle` that is in the project root directory.

### Testing

- `npm t`
- `cordova build android` with a plugin that has a subproject.

### Checklist

### DIFF
--- a/framework/build.gradle
+++ b/framework/build.gradle
@@ -29,7 +29,13 @@ buildscript {
 }
 
 allprojects {
-    apply from: 'repositories.gradle'
+    def hasRepositoriesGradle = file('repositories.gradle').exists()
+    if (hasRepositoriesGradle) {
+        apply from: 'repositories.gradle'
+    } else {
+        apply from: "${project.rootDir}/repositories.gradle"
+    }
+
     repositories repos
 }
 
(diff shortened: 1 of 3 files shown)


erisu:
@goffioul I saw your https://github.com/apache/cordova-android/pull/1179#issuecomment-820486928 about an issue that appeared in that PR. The issue was where the `repositories.gradle` file is missing for `subproject`.

I went ahead and created this PR that should resolve the issue. I followed based on your comments. If the file is missing, it will fall back to the project's root `repositories.gradle`.

Sorry for the slow response. Since the comment was in a merged PR, I missed it.

If you have a project where you could test with these changes, that would be great.

This PR's branch is based on the current master branch, which means it contains Cordova-Android 10.x. If your project does not support Cordova-Android 10 yet, you might have to cherry-pick the commit. I am not positive if this fix will be back ported to 9.1.x, but there might be a chance.

erisu:
I will merge this in on Monday with or without a review.

The default project build successfully.

I tested with an only plugin that I could find that added a subproject. Without the PR, It showed the repository error, but with this PR, the repository error was fixed/gone. I couldnt fully test the build because the plugin itself was 4+ years old, unmatained, and used libraries that doesnt work anymore with the current Android SDK & Gradle Build.

PR #1304, opened 2021-07-31
feat: bump android-gradle-plugin@7.0.1

mosabab:
### Motivation and Context

Bump Android Gradle Plugin: AGP_VERSION@7.0.1 Using the new centralized configuration defaults.

### Testing

npm t

### Checklist


jacobg:
Is this merge what will enable compilation in Arctic Fox?

breautek:
> Is this merge what will enable compilation in Arctic Fox?

Probably. Artic Fox is the AS release to handle a lot of breaking changes between API 30 and API 31. AGP 7 also introduces a ton of other breaking changes, including requiring Java 11. cordova-android very likely won't work out of the box with AGP 7. AGP 7 I think also requires build tools / API 31.

mosabab:
JDK 11 required to run AGP 7.0
according to https://developer.android.com/studio/releases/gradle-plugin

When using Android Gradle plugin 7.0 to build your app, JDK 11 is now required to run Gradle

https://developer.android.com/studio/releases/gradle-plugin#jdk-11

PR #1326, opened 2021-08-16
8.1.x

Sharavanan7944:
### Checklist

### DIFF
--- a/bin/templates/project/app/build.gradle
+++ b/bin/templates/project/app/build.gradle
@@ -27,7 +27,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.3.0'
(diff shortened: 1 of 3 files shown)


Sharavanan7944:
Changed gradle version to support android 11 package visibility to be configured if necessary
https://developer.android.com/about/versions/11/privacy/package-visibility

To prepare for package visibility need gradle patch as per docs
https://android-developers.googleblog.com/2020/07/preparing-your-build-for-package-visibility-in-android-11.html

breautek:
Thank you for the PR, but Cordova isn't supporting 8.x anymore.

Of course, you can always use the changes in your own projects as necessary but we would recommend upgrading your cordova-android version to the latest if possible. Cordova-android@10 will use gradle 7.1.1 by default, which contains the requirements for android 11 package visibility.

Kind regards,
Norman

Issue #1328, opened 2021-08-26
10.1.0 - deviceready not fired when hostname is set and plugin installed


codeconsole:
**>>> ANY <<< Android 10.1.0** app will break and deviceready will not fire when `<preference name="hostname" value="apache.org" />` is set and **cordova-plugin-device** is installed.

I do not believe this is related to the plugin because simply adding 
`<preference name="AndroidInsecureFileModeEnabled" value="true" />` resolves the issue.

Here is a complete very minimal app replicating the issue:
https://github.com/codeconsole/cordova-bug

However, you can break any app by just adding this preference `<preference name="hostname" value="apache.org" />`  and installing **cordova-plugin-device**.

This is possibly related to this commit:
https://github.com/apache/cordova-android/pull/1275/files#diff-c2eeba9559f1ec42b3e90bd1531f3389a2aab9d199752d4cca70fbf489d92ca0R88

in which https is hard coded.

lirancr:
@erisu 

any chance to share a like to the PR ? just to verify it solve the use case in the meanwhile ?

erisu:
Sorry, I got distracted and forgot to submit the PR. I just linked to PR to this ticket, seen above.

PR #1332, opened 2021-08-31
fix(PluginManager): AllowBridgeAccess default policy to handle scheme & hostname

erisu:
### Motivation and Context

fixes: #1328
fixes: https://github.com/apache/cordova-plugin-device/issues/118

### Description

Update how the PluginManager handles the default policy.

* If `AndroidInsecureFileModeEnabled` is `true`, then the default policy is `file://`
* If `AndroidInsecureFileModeEnabled` is `false`, then the default policy is the combination of the `scheme` & `hostname`.
  * The default scheme is: `https`
  * The default hostname is: `localhost`
  * If the user changes any of these values, it will be used.

### Testing

* Creating Project
* Adding Platform
* Adding Device Plugin
* Changing `scheme` to `http`
* Changing `hostname` to `apache.org`
* Setting `AndroidInsecureFileModeEnabled` to `true`

### Additional Notes

I also tested a solution that added the fix to the `AllowListPlugin` class. https://github.com/apache/cordova-android/commit/fa5b228bb9cefca36d92df3c80bb6bbc42672377

This is where @NiklasMerz originally the code. The issue is that is was written in the config parser step. The config parser has not fully parse everything, e.g. the preferences, and therefore always picked the default values.

If we wanted to keep the login inside the `AllowListPlugin` file, we would have to move it to the `pluginInitialize` method, after the config parser has been completed.

The question really comes down to, where it should reside?

* The `PluginManager`, is where the default policy is defined.
* The `AllowListPlugin` is where we parse the user's custom-defined `allow-navigation`, `allow-intent`, `access`, and `content`.

IMO the `PluginManager` sounded better. Previously the default policy was `file://` but since `WebViewAssetLoader` was added and default should now be `https://localhost`. But also, since users can change the launch URL to either file or default/custom scheme+hostname, it seems maybe the default policy should be a bit flexible.
### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/AllowListPlugin.java
+++ b/framework/src/org/apache/cordova/AllowListPlugin.java
@@ -32,6 +32,11 @@ public class AllowListPlugin extends CordovaPlugin {
     public static final String PLUGIN_NAME = "CordovaAllowListPlugin";
     protected static final String LOG_TAG = "CordovaAllowListPlugin";
 
+    // @todo same as ConfigXmlParser. Research centralizing ideas, maybe create CordovaConstants
+    private static String SCHEME_HTTPS = "https";
+    // @todo same as ConfigXmlParser. Research centralizing ideas, maybe create CordovaConstants
+    private static String DEFAULT_HOSTNAME = "localhost";
+
     private AllowList allowedNavigations;
     private AllowList allowedIntents;
     private AllowList allowedRequests;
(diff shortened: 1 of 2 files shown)


NiklasMerz (approved):
Code LGTM. I agree that this is a better place to set these settings.

I didn't like how this was implemented in the allow list plugin and config parser.

I did a quick test in my test project for https://github.com/apache/cordova-plugin-device/issues/118 and this issues seems to be solved. Similar to #1328

jcesarmobile (approved):

Issue #1322, opened 2021-08-12
The AllowList plugin may throw a NullPointerException in some case


YCKang:
# Bug Report

## Problem
The AllowList plugin may throw a NullPointerException in some case.

### What is expected to happen?
To check the url correctly.

### What does actually happen?
Throw NullPointerException error.

## Information

If you don't set any `<allow-navigation>` in the config.xml, `this.shouldAllowNavigation(url)` will return `null`.
And operator '||' cannot be applied to 'null', 'boolean', it will throw NullPointerException error.

L130 should be `return (Boolean.TRUE.equals(this.shouldAllowNavigation(url)) || this.allowedRequests.isUrlAllowListed(url)) `.

https://github.com/apache/cordova-android/blob/bedb4427f420157ac547588acdf1083e8e916e20/framework/src/org/apache/cordova/AllowListPlugin.java#L122-L133

### Version information

cordova-android@10.0.1

## Checklist

- [X] I searched for existing GitHub issues
- [X] I updated all Cordova tooling to most recent version
- [X] I included all the necessary information above

breautek:
PR is welcome

jskrepnek:
If I understand this right, it would mean that any allowed requests that are not also in allowed navigations will always fail.  I'm able to reproduce this, even if there are some `allow-navigation` tags, for any requests that are also not in allowed navigations, even if they are in allowed requests.

PR #1342, opened 2021-09-08
fix(AllowListPlugin): Safely handle default allow navigation policy in allow request

jskrepnek:
### Motivation and Context

fixes: https://github.com/apache/cordova-android/issues/1322

### Description

`shouldAllowRequest` will throw an exception for all URLS that are not in the navigation allowed list.  This change safely tests the result of `shouldAllowNavigation`, avoiding the exception and allowing for the execution to continue to see if the URL is in the request allowed list.

### Testing

Incorporating this change into our project successfully allows expected network requests that previously were allowed with the whitelist plugin.

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/AllowListPlugin.java
+++ b/framework/src/org/apache/cordova/AllowListPlugin.java
@@ -127,7 +127,7 @@ public class AllowListPlugin extends CordovaPlugin {
 
     @Override
     public Boolean shouldAllowRequest(String url) {
-        return (this.shouldAllowNavigation(url) || this.allowedRequests.isUrlAllowListed(url))
+        return (Boolean.TRUE.equals(this.shouldAllowNavigation(url)) || this.allowedRequests.isUrlAllowListed(url))
             ? true
             : null; // default policy
     }


jskrepnek:
@erisu I'm a bit confused as to why `shouldAllowRequest` calls `shouldAllowNavigation` in the first place.  

My basic expectation is that `shouldAllowNavigation` would map to `allow-navigation` tags and `shouldAllowRequest` would map to `access` tags.

Haven't looked deeply, just wanted to confirm it was intended and curious why.

breautek (approved):

erisu:
I would have to take a look at the code to try and figure it out myself.

I didnt write the original code. All I did was integrate the whitelist plugin as a core feature of the Cordova Android platform and renamed it to AllowList.

jskrepnek:
> I would have to take a look at the code to try and figure it out myself.
> 
> I didnt write the original code. All I did was integrate the whitelist plugin as a core feature of the Cordova Android platform and renamed it to AllowList.

Ah, okay.  It looks like a bit of refactoring led to the bug.  This is the [original code](https://github.com/apache/cordova-plugin-whitelist/blob/master/src/android/WhitelistPlugin.java#L119):

```
    @Override
    public Boolean shouldAllowRequest(String url) {
        if (Boolean.TRUE == shouldAllowNavigation(url)) {
            return true;
        }
        if (allowedRequests.isUrlWhiteListed(url)) {
            return true;
        }
        return null; // Default policy
    }
```

erisu:
Ah yes, nice catch.

That refactoing bug would probally have been my fault :D Sorry

Changes looks correct.

erisu (approved):

Issue #1380, opened 2021-11-24
Error : Invalid character in entity name when &amp; is in the <name>


tpneumat:
# Bug Report

## Problem

Can not include `&amp; ` in the cordova app name without causing a build failure to occur when strings.xml is created.

### What is expected to happen?

If I have properly escaped the & in the config.xml, it should also apply to the strings.xml.  

### What does actually happen?
It seems the `&amp;` is converted back to `&` when strings xml is created, perhaps in prepare.js.  This causes the cordova add command to fail.

## Information

Given the following config.xml
```
<?xml version='1.0' encoding='utf-8'?>
<widget id="xxx.xxx.xxxx" version="6.2.2" android-versionCode="622000062" versionCode="62" xmlns="http://www.w3.org/ns/widgets" xmlns:android="http://schemas.android.com/apk/res/android" xmlns:gap="http://phonegap.com/ns/1.0">

    <name>TIMING &amp; TRACKING</name>
    <description>
        Timing APP
    </description>
```

### Command or Code
Run the command to create the cordova project

```
cordova platform add android@10.1.1
Using cordova-fetch for cordova-android@10.1.1
Adding android project...
Creating Cordova project for the Android platform:
	Path: platforms/android
	Package: me.rtrt.app.tsa
	Name: TIMING & TRACKING
	Activity: MainActivity
	Android target: android-30
Subproject Path: CordovaLib
Subproject Path: app
Android project created with cordova-android@10.1.1
Invalid character in entity name
Line: 3
Column: 37
Char:
```

Then, checking the strings.xml that was created, you can find there is a non-escaped & 

/platforms/android/app/src/main/res/values/strings.xml

```
<?xml version="1.0" encoding="utf-8"?>
<resources>
    <!-- App label shown within list of installed apps, battery & network usage screens. -->
    <string name="app_name">TIMING & TRACKING</string>
    <!-- App label shown on the launcher. -->
    <string name="launcher_name">@string/app_name</string>
    <!-- App label shown on the task switcher. -->
    <string name="activity_name">@string/launcher_name</string>
</resources>
```

### Environment, Platform, Device
N/A

### Version information

Android  Cordova 10.1.1 
MacOs 10.15.7 (catalina)

## Checklist

- [X] I searched for existing GitHub issues
- [X] I updated all Cordova tooling to most recent version
- [X] I included all the necessary information above

tiagoappereira:
[...]

PR #1384, opened 2021-12-07
fix: escape strings.xml app name

tiagoappereira:
### Platforms affected
Android

### Motivation and Context

Error on `cordova platform add` when the Cordova app name contains the character `&`. The issue occurs when parsing the `strings.xml` file. The issue was introduced on cordova-android 10.1.0

Fixes https://github.com/apache/cordova-android/issues/1380

### Description

Simply replaces the `&` characters in the project name (if present) by an `_`. 

### Testing

`npm test` and project creation successful

### Checklist

### DIFF
--- a/lib/create.js
+++ b/lib/create.js
@@ -255,7 +255,7 @@ exports.create = function (project_path, config, options, events) {
             fs.ensureDirSync(activity_dir);
             fs.copySync(path.join(project_template_dir, 'Activity.java'), activity_path);
             utils.replaceFileContents(activity_path, /__ACTIVITY__/, safe_activity_name);
-            utils.replaceFileContents(path.join(app_path, 'res', 'values', 'strings.xml'), /__NAME__/, project_name);
+            utils.replaceFileContents(path.join(app_path, 'res', 'values', 'strings.xml'), /__NAME__/, utils.escape(project_name));
             utils.replaceFileContents(activity_path, /__ID__/, package_name);
 
             var manifest = new AndroidManifest(path.join(project_template_dir, 'AndroidManifest.xml'));
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -66,3 +66,21 @@ exports.forgivingWhichSync = (cmd) => {
 
 exports.isWindows = () => os.platform() === 'win32';
 exports.isDarwin = () => os.platform() === 'darwin';
+
+const UNESCAPED_REGEX = /[&<>"']/g;
+
+const escapes = {
+    '&': '&amp;',
+    '<': '&lt;',
+    '>': '&gt;',
+    '"': '&quot;',
+    "'": '&#39;'
+};
+
+/**
+ * Converts the characters "&", "<", ">", '"' and "'" in the given string to
+ * their corresponding escaped value
+ * @param {string} str the string to be escaped
+ * @returns the escaped string
+ */
+exports.escape = (str) => UNESCAPED_REGEX.test(str) ? str.replace(UNESCAPED_REGEX, (key) => escapes[key]) : str;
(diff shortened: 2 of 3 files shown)


dpogue:
Wouldn't the correct fix be to XML-escape the `&` (or other special characters like `<` and `>`)?

breautek (approved):
Lgtm

jcesarmobile (changes requested):
package.json doesn't include lodash.
It works because there are 3 other dependencies that include lodash as dependency themselves but it could break in the future if those dependencies are replaced/removed or they stop using lodash in a newer version. 
So I think it should be included in the package.json to prevent that.

jcesarmobile (approved):

erisu (changes requested):
I am not infavor for installing the entire lodash package just to use one method.

Either install the exact package, lodash.escape, or write our own escape method please.

**Lodash**
package size:  319.0 kB
unpacked size: 1.4 MB

**Lodash Escape**
package size:  3.9 kB
unpacked size: 9.1 kB

jcesarmobile:
It's already being installed because it's a dependency of other 3 dependencies.
As you can see, the package-lock.json doesn't change because it's already installed.

I would be in favor of not installing lodash, but that would mean updating a lot of our dependencies (including cordova-common as it depends on @netflix/nerror which depends on lodash)

erisu:
We have been removing lodash from our dependencies because of the vuln. report frequencies.

I have discussed with others on dropping `@netflix/nerror` because they neglected to update their package over 2 years now and even with PRs open to resolve some issues.

https://github.com/Netflix/nerror/pull/18

breautek:
> We have been removing lodash from our dependencies because of the vuln. report frequencies.
> 
> I have discussed with others on dropping `@netflix/nerror` because they neglected to update their package over 2 years now and even with PRs open to resolve some issues.
> 
> [Netflix/nerror#18](https://github.com/Netflix/nerror/pull/18)

To add onto this, we can probably solve the same issue using the JS native method [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) so that we can avoid introducing another lodash dependency. I'm pretty sure the method exists in NodeJS environments.

jcesarmobile:
`encodeURIComponent` won't work because it scapes `<Incredible&App>` to `%3CIncredible%26App%3E`, which is not valid on the `strings.xml`.

I solved a similar problem in Capacitor using replace like this
https://github.com/ionic-team/capacitor/pull/5325/files#diff-ed3a1ffe6172324599c40aaeea6832beab33815ec5c28ba7f87e1a37fb55fae1R74-R78

note that it doesn't include `>` replacement because it wasn't needed in Capacitor, but might still be needed in Cordova as it looks like it does some previous conversion
also note that it includes `"`, which is not escaped by lodash

erisu:
Here is what lodash escape does, just for additional details:

```
  '&': '&amp;',
  '<': '&lt;',
  '>': '&gt;',
  '"': '&quot;',
  "'": '&#39;',
  '`': '&#96;'
```

jcesarmobile:
there is some previous conversion somewhere else, because I tested with `'` and `"` and `'` was converted to `\'` and `"` wasn't converted at all.
Note that for them to work in `strings.xml` they have to be `\'` and `\"`, didn't try ` (by work I mean that the app name is displayed on the phone exactly as on the `strings.xml`)

jcesarmobile:
I found where the `'` is converted to `\'`, is not done by lodash

https://github.com/apache/cordova-android/blob/09c75237d9348a944b4946ccb31d2dee90330ce4/lib/prepare.js#L266

erisu (approved):

PR #1410, opened 2022-04-04
feat!: API 31 support

breautek:
### Platforms affected

Android

### Motivation and Context

Changes necessary to support API 31.

### Description

Version updates:

SDK Version 30 -> 31
Gradle 7.1.1 -> 7.4.2
AGP 4.2.2 -> 7.1.0
Build Tools 30.0.3 -> 31.0.0

### Testing

Ran hello cordova app in emulator. Ran npm test successfully.

### Other remarks

To test this PR, requires the following changes:

With AGP 7.x, Java 11 is now required. Ensure that `JAVA_HOME` is set to a jdk 11 path and `PATH` has jdk binaries.

Further instructions assumes that you have `ANDROID_SDK_ROOT` path set appropriately. If you still use the older `ANDROID_HOME`, then `$ANDROID_SDK_ROOT` should be substituted with `$ANDROID_HOME`

With this change, `cmdline-tools` package is required. I tesed with `cmdline-tools` 6.0. If you attempt to use the old tools from `$ANDROID_SDK_ROOT/tools/bin`, it will crash. You will need to add `$ANDROID_SDK_ROOT/cmdline-tools/<version>/bin` to your `PATH` variable.

If you have `$ANDROID_SDK_ROOT/tools/bin` path in your `PATH` variable, it should be removed.

API 31 requires build tools 31. Use Android Studio's SDK Manager to install build tools 31.0.0. And ensure that `$ANDROID_SDK_ROOT/build-tools/31.0.0` is in your `PATH` variable.

These things probably should be mentioned in our release announcement.

### Checklist


erisu (approved):

ath0mas:
Thanks for v11 💯 !!

Just to let you know that the `AGP 4.2.2 -> 7.1.0` is partly blocking for IntelliJ IDEA users for now (not Android Studio).
When opening the generated Android native project, an error during Sync prevents further loading because of AGP version incompatibility.

Details:
- IntelliJ IDEA latest stable version is **2022.1.4**
- its Android plugin is mapped to Android Studio, and current version is **bundled 2020.3.1 Final.221.6008.13** (corresponding to Arctic Fox)
- in practice and given the [compatibility table](https://developer.android.com/studio/releases/gradle-plugin#agp-studio-compatibility), **AGP can only be 7.0 at max** for now to work properly
- this PR sets v7.1, requiring **Bumblebee | 2021.1.1**, and was followed by an update to v7.2 in #1446, requiring **Chipmunk | 2021.2.1**

Workarounds (see their [issue tracker](https://youtrack.jetbrains.com/issue/IDEA-252823/Gradle-Android-Support-Plugin-fails-to-import-project-misleadingly-complains-about-the-version-of-the-Android-Support-Gradle)):
> please use IDEA 2022.2 or downgrade AGP to version 7.0.x
- edit `framework/cdv-gradle-config-defaults.json` to force `AGP_VERSION` to, let say, **7.0.1** ; and force reload/Sync
- try to update IntelliJ IDEA to preview builds of **2022.2** (currently [RC](https://blog.jetbrains.com/idea/2022/07/intellij-idea-2022-2-release-candidate/))

erisu:
@ath0mas Have you tried setting the `preference` flag `AndroidGradlePluginVersion` instead of editing the default file directly?

```xml
<preference name="AndroidGradlePluginVersion" value="7.0.1" />
```
Then run CLI command: `cordova prepare android`

This will update the `platforms/android/cdv-gradle-config.json` that is read by the project.

I guess it depends on the use case, but I think the above config is enough for most cases?

ath0mas:
@erisu You are correct, your solution works great too! Thanks 👍 

I was not aware of this preference as it is not documented.
See https://cordova.apache.org/docs/en/11.x/config_ref/index.html#preference.
I can only find it in cordova-android [source code](https://github.com/apache/cordova-android/blob/e5b7e8ab266c8f9de83923430be77c62499810a2/lib/prepare.js#L98-L112).

breautek:
I wasn't aware there can be some restrictions in the gradle version + android studio, I figured they'd support at least a given major with any minor/patch version. Thanks for the [compatibility](https://developer.android.com/studio/releases/gradle-plugin#agp-studio-compatibility) table, this is definitely good to know.

Regarding the preferences, yah there are a number of preferences that was added (back in v10 if I recall correctly) without actually being documented. Some of these preferences will very likely break builds if you use something other than the default configured. Cordova is only ever tested against the default configuration for build tools, AGP version, etc... I think we can document these but I think some of these new preferences needs to be stressed that they are for advanced usage / cordova development only

I'll create a issue on the docs repo as a reminder. Eventually I'll get to it, if someone else doesn't get to it first. PRs are welcome.

Issue #1373, opened 2021-10-29
[+use case: auto-reset of permissions] Overriding compileSDK version has been removed


shankari:
# Bug Report

## Problem

Overriding compileSDK version has been removed in cordova-android@10.x
https://github.com/apache/cordova-android/issues/1310#issuecomment-892802205

I see that this was an intentional change, but @breautek said that

> The stance that we took is "We'll make the change, release it, and if it becomes a problem, then we'll bring back the independent settings". So I'd like to understand your use case better.

Here's the use case. I wasn't sure if this was a bug since the change was intentional, but it is arguably a regression, and is inconsistent with the documentation, so I went ahead and chose "bug" instead of "feature request".

### What is expected to happen?

I can include 

```
+    <framework src="androidx.core:core:$ANDROIDX_CORE_VERSION"/>
+    <preference name="ANDROIDX_CORE_VERSION" default="1.7.0"/>
```

in my plugin

### What does actually happen?

When I add that plugin and compile the app, I get the error

```
Execution failed for task ':app:checkDebugAarMetadata'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckAarMetadataWorkAction
   > The minCompileSdk (31) specified in a
     dependency's AAR metadata (META-INF/com/android/build/gradle/aar-metadata.properties)
     is greater than this module's compileSdkVersion (android-30).
     Dependency: androidx.core:core:1.7.0.
     AAR metadata file: /Users/kshankar/.gradle/caches/transforms-3/553b5caab8c99faefbf90799c7dc70a3/transformed/core-1.7.0/META-INF/com/android/build/gradle/aar-metadata.properties.
```

## Information

Starting in December, android will start resetting permissions of apps that have not been brought into the foreground "for a while". This will happen irrespective of the phone's android version (https://developer.android.com/topic/performance/app-hibernation).

It only applies to apps whose target SDK is Android 11+ (API 30+), but the google play min API requirements are 30+ now for new apps and will be 30+ for app updates in November (https://developer.android.com/distribute/best-practices/develop/target-sdk)

If such a new or updated app reads location in the background (e.g. my app: https://github.com/e-mission/e-mission-phone/), even if it is installed on an Oreo phone, it will stop working "soon after" December.
https://developer.android.com/topic/performance/app-hibernation#effects

[...]

Issue #1310, opened 2021-08-04
Unable to set targetSdk != compileSdk, fail to compile with targetSdk < 30


goffioul:
I used to have the following in my `config.xml`, while using build tools 29.x:
```
<preference name="android-minSdkVersion" value="23" />
<preference name="android-targetSdkVersion" value="25" />
```

This doesn't work anymore with cordova-android 10. The above config leads to the following in `platforms/android/cdv-gradle-config.json`:
```
"SDK_VERSION": 25
```

And `platforms/android/app/build.gradle` contains:
```
targetSdkVersion cordovaConfig.SDK_VERSION
compileSdkVersion cordovaConfig.SDK_VERSION
```

So it seems now impossible to have targetSdk != compileSdk. This can be annoying as you may need to use newer compile SDK due to the use of support libraries (e.g. appcompat), but your app may not be ready to target the same newer API level.

On the other hand, using a lower compile SDK (like 25 in the example above) actually leads to compilation error, due to the requirement of the support libraries.

I understand the reason for bumping the target SDK to 30, but Play Store is not the only way to deploy apps on Android, especially on systems that do not use the Google ecosystem. And even then, and unsupported `android-targetSdkVersion` value should probably error earlier in the process, or just be ignored when adding the platform.

breautek:
This change introduced in cordova-android@10 was intentional in effort to uniform and simplify the API that Cordova provides.

The rationale behind this change is that we didn't see any **good** reason to keep the two settings as two independent fields and they were often a source of confusion or errors when they are misconfigured.

Having a Target SDK that is > than the Compile SDK is simply invalid. You cannot target something that you're not compiling against.

On the other hand, having a Target SDK that is less than the Compile SDK is valid, but we saw no reason why someone would want to have a mismatch Target/Compile SDK. If they wanted to test against a newer API, then it's best to set the both the target and compile SDK.

Having said that, this decision is by no means final. The stance that we took is "We'll make the change, release it, and if it becomes a problem, then we'll bring back the independent settings". So I'd like to understand your use case better. Perhaps expand on

> ...you may need to use newer compile SDK due to the use of support libraries (e.g. appcompat)...

Isn't these libraries independent of the compile SDK? Looking at the [android docs](https://developer.android.com/jetpack/androidx/versions) at a variety of different androidx packages, I don't ever recall (nor can I find anything that states you need a specific compileSdk) to use them.

Kind regards,
Norman

goffioul:
[...]

PR #1431, opened 2022-05-13
feat: support custom compileSdk setting

erisu:
### Motivation and Context

closes https://github.com/apache/cordova-android/issues/1310
closes https://github.com/apache/cordova-android/issues/1373

Allow setting a custom `compileSdk`version, different from targetSdk.

Some libraries might require a higher compileSdk while it can still target a lower Sdk.

Generally it is recommended that the targetSdk and compileSdk match.

If a user decides to increase the `targetSdk`, they will now need to also manually bump the `compileSdk` or it will remain behind.

### Description

Can set custom `compileSdk` version with the follow `preference`:

```xml
<preference name="android-compileSdkVersion" value="31" />
```

A warning will display during the `platform add` and `prepare` step when the `targetSdk` is higher then the `compileSdk`.

Even if the build passes, there could be potential errors further down. 

### Testing

`npm t`

### Checklist

### DIFF
--- a/framework/build.gradle
+++ b/framework/build.gradle
@@ -26,6 +26,8 @@ buildscript {
         // Android Gradle Plugin (AGP) Build Tools
         classpath "com.android.tools.build:gradle:${cordovaConfig.AGP_VERSION}"
     }
+
+    cdvHelpers.verifyCordovaConfigForBuild()
 }
 
 allprojects {
--- a/framework/cordova.gradle
+++ b/framework/cordova.gradle
@@ -191,9 +191,11 @@ def doApplyCordovaConfigCustomization() {
             but got Android Build Tools version ${cordovaConfig.BUILD_TOOLS_VERSION}. Please use version ${cordovaConfig.MIN_BUILD_TOOLS_VERSION} or later.
         """.replaceAll(/\s+/, ' ').trim())
     }
+}
 
+def doVerifyCordovaConfigForBuild() {
     if (cordovaConfig.COMPILE_SDK_VERSION < cordovaConfig.SDK_VERSION) {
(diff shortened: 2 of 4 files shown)


erisu:
Downside of allowing custom compileSdk..

- Plugins can take control of defining the version and users may not notice it changed.. Might lead to issues
- If the user bumps the targetSdk, the compileSdk will no longer be in sync. The users will now have to manually control the compileSdk.

breautek:
> If the user bumps the targetSdk, the compileSdk will no longer be in sync. The users will now have to manually control the compileSdk.

This might be difficult since we use a JSON file to create our defaults... but wonder if there's a way to default `compileSdk` to the value of `targetSdk`.

breautek (commented):
I've made a collection of suggestions that if accepted... probably should be applied together.

The attempt is rather than explicitly declaring the `COMPILE_SDK_VERSION` default to a hard-coded value, is to make it's default the same as `SDK_VERSION`.

The suggested changes I believe will work (but untested) so that unless explicitly declared, compile SDK will be whatever what target SDK is set to (even if the user changes the target SDK).

e.g.:

### Use Case 1

User does not declare either target or compile

Target defaults to 32, as defined in defaults json.
compile will default to target, which is set to 32.

### Use Case 2

User explicitly sets target sdk to 33

Target will be set to 33 from config preferences.
Compile will default to target, which is set to 33.

### Use Case 3

User explicitly sets compile sdk to 33

Target sdk will be set to 32, which comes from the defaults json.
Compile sdk will be set to 33, which comes from config preferences.

### Use Case 4

User explicitly sets both target to 30 & compile sdk to 32

Target sdk will be set to 30, which comes from config preferences.
compile sdk will be set to 32, which comes from the config preferences.

Let me know of your thoughts, or whether if this is even a good or bad idea.

breautek (approved):

breautek (approved):

breautek (approved):

dpogue (approved):
LGTM

erisu:
Thank you for the votes!

PR #1444, opened 2022-06-27
fix!: set & use ANDROID_HOME as default

### LINKED ISSUE #1425: ANDROID_HOME is un-deprecated, and ANDROID_SDK_ROOT is now deprecated
As of April 27, 2022, the page at https://developer.android.com/studio/command-line/variables says:

> ANDROID_HOME: Sets the path to the SDK installation directory. Once set, the value does not typically change, and can be shared by multiple users on the same machine. ANDROID_SDK_ROOT, which also points to the SDK installation directory, is deprecated. If you continue to use it, Android Studio and the Android Gradle plugin will check that the old and new variables are consistent.

So Cordova is saying the opposite of what the Android team is saying.

_Originally posted by @fortuna in https://github.com/apache/cordova-android/issues/949#issuecomment-1111400691_

erisu:
### Motivation and Context

Resolves #1425

### Description

* Check `ANDROID_HOME` before `ANDROID_SDK_ROOT`
* Set `ANDROID_HOME` based on paths
* Flag `ANDROID_SDK_ROOT` as deprecate

### Testing

`npm t`

### Checklist

### DIFF
--- a/lib/check_reqs.js
+++ b/lib/check_reqs.js
@@ -179,15 +179,15 @@ module.exports.check_android = function () {
         function maybeSetAndroidHome (value) {
             if (!hasAndroidHome && fs.existsSync(value)) {
                 hasAndroidHome = true;
-                process.env.ANDROID_SDK_ROOT = value;
+                process.env.ANDROID_HOME = value;
             }
         }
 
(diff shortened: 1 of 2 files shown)


breautek (approved):
lgtm

Issue #1080, opened 2020-10-02
Security issue with setAcceptFileSchemeCookies


systemlord01:
# Bug Report

### What is expected to happen?
We expect cordova to use safe methods that are not depricated to save cookies.

### What does actually happen?
Cordova uses a depricated API that is a known security issue.

## Information
In the following file a deprecated api is used with a known security issue (setAcceptFileSchemeCookies).
https://github.com/apache/cordova-android/blob/master/framework/src/org/apache/cordova/engine/SystemCookieManager.java

The following docs mention it is deprecated and a security issues.
https://developer.android.com/reference/android/webkit/CookieManager#setAcceptFileSchemeCookies(boolean)

### Command or Code

Use cordova.

### Environment, Platform, Device

Android

### Version information
Cordova latest. Corodava-android.

## Checklist
- [ x] I searched for existing GitHub issues
- [ x] I updated all Cordova tooling to most recent version
- [ x] I included all the necessary information above

breautek:
I believe this is only used in the current version of cordova if `AndroidInsecureFileModeEnabled` is set to true, which makes cordova not to use the `WebViewAssetLoader` system and return back to the original file-based system.

By default, cordova will use the `WebViewAssetLoader` and file system paths are disabled, as recommended by Google, which renders the use of this flag obsolete even if it is used unconditionally (I believe).

While potentially insecure (by potentially data leaking) it is required for users using `AndroidInsecureFileModeEnabled` and depends on the use of cookies. The configuration is opt-in and I think the name speaks for itself that developers are opting into an insecure method.

Eventually Cordova will completely drop support for file-based webview assets and force WebViewAssetLoader and when that does happen we'll be able to remove `setAcceptFileSchemeCookies`. I'll also re-iterate that I believe this is only an issue if you have `AndroidInsecureFileModeEnabled` enabled.

PR #1449, opened 2022-06-28
fix: accept file cookies only if AndroidInsecureFileModeEnabled

jcesarmobile:
### Platforms affected
android

### Motivation and Context

we are enabling file cookies even if AndroidInsecureFileModeEnabled is not enabled

### Description

accept file cookies only if AndroidInsecureFileModeEnabled is true
also suppress the warning caused by using that method 

removes two unused imports
closes https://github.com/apache/cordova-android/issues/1080

### Checklist


erisu (commented):
Changes LGTM 👍

PR #1451, opened 2022-06-30
chore: display warning on deprecated <splash> tag usage

erisu:
### Motivation and Context

Users may still use old tag.

### Description

Display a warning to migate to the new preference tag when `<splash>` tag is used.

This only checks for slash tags in `platform[name="android` as global tags could still be used in other platforms.

### Testing

`cordova build`

### Checklist


NiklasMerz (approved):
This works.

Looks like I get these warnings twice but I think this is ok

grafik

Issue #1475, opened 2022-08-13
Error when adding AndroidWindowSplashScreenBrandingImage


fortunella:
# Bug Report

### What is expected to happen?
Branding image is shown in the splash screen

### What does actually happen?
Error during build of the project.

## Information

I migrated my app project to cordova-android 11 and got the splash screen working with the new API. Therefor I added these lines to config.xml:
```
<preference name="AndroidWindowSplashScreenAnimatedIcon" value="resources/android/splash-android.png" />
<preference name="AndroidWindowSplashScreenBackground" value="#FF0000" />
```
This works fine. I then wanted to add a branding image to the bottom of the splash screen with this setting:
```
<preference name="AndroidWindowSplashScreenBrandingImage" value="resources/android/brand-image.png" />
```
When I then build the project I get this error:
```
Execution failed for task ':app:processDebugResources'.
> A failure occurred while executing com.android.build.gradle.internal.res.LinkApplicationAndroidResourcesTask$TaskAction
   > Android resource linking failed
     ERROR:/home/ralf/myAppBuild/platforms/android/app/build/intermediates/incremental/debug/mergeDebugResources/merged.dir/values/values.xml:2210: AAPT: error: style attribute 'attr/windowSplashScreenBrandingImage (aka de.foonax.myapp:attr/windowSplashScreenBrandingImage)' not found.
```
The specified image file exists at the specified location.

### Version information

cordova-android:  11.0.0

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

expcapitaldev:
looks like error here:
`<resources>
    <style name="Theme.App.SplashScreen" parent="Theme.SplashScreen.IconBackground">
        <item name="windowSplashScreenBackground">@color/cdv_splashscreen_background</item>
        <item name="windowSplashScreenAnimatedIcon">@drawable/ic_cdv_splashscreen</item>
        <item name="windowSplashScreenAnimationDuration">1000</item>
        <item name="postSplashScreenTheme">@style/Theme.AppCompat.NoActionBar</item>
        <item name="windowSplashScreenIconBackgroundColor">@color/cdv_splashscreen_icon_background</item>
        <item name="**android:windowSplashScreenBrandingImage**">@drawable/ic_cdv_splashscreen_branding</item>
    </style>
</resources>`
 because after cordova prepare I see 
[...]

PR #1487, opened 2022-09-11
fix: Add android prefix to windowSplashScreenBrandingImage

jcesarmobile:
closes https://github.com/apache/cordova-android/issues/1475

`windowSplashScreenBrandingImage` is not valid, it should be `android:windowSplashScreenBrandingImage`.

All other values should also include the `android:`, but the support library allows to use them without the `android:` part, but since `windowSplashScreenBrandingImage` is not supported by the support library apps fail to compile.

This PR adds the `android:` prefix to the value so it gets added to the styles file with the prefix.
Needed to patch how the cdvConfigPrefKey is generated to strip the android: part of the themeKey

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -382,14 +382,10 @@ function updateProjectSplashScreen (platformConfig, locations) {
         'windowSplashScreenIconBackgroundColor',
         'postSplashScreenTheme'
     ].forEach(themeKey => {
-        let index = 0;
-        if (themeKey.indexOf(':') !== -1) {
-            index = themeKey.indexOf(':') + 1;
-        }
+        let index = themeKey.indexOf(':') + 1;
         const cdvConfigPrefKey = 'Android' + themeKey.charAt(index).toUpperCase() + themeKey.slice(index + 1);
         const cdvConfigPrefValue = platformConfig.getPreference(cdvConfigPrefKey, this.platform);
         let themeTargetNode = splashScreenTheme.find(`item[@name="${themeKey}"]`);
-        console.log('hola', cdvConfigPrefKey);
         switch (themeKey) {
         case 'windowSplashScreenBackground':
             // use the user defined value for "colors.xml"
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -378,14 +378,18 @@ function updateProjectSplashScreen (platformConfig, locations) {
         'windowSplashScreenAnimatedIcon',
         'windowSplashScreenAnimationDuration',
         'windowSplashScreenBackground',
-        'windowSplashScreenBrandingImage',
+        'android:windowSplashScreenBrandingImage',
         'windowSplashScreenIconBackgroundColor',
         'postSplashScreenTheme'
     ].forEach(themeKey => {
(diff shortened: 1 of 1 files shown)


lib/prepare.js
```diff
@@ -378,14 +378,18 @@ function updateProjectSplashScreen (platformConfig, locations) {
...
-        const cdvConfigPrefKey = 'Android' + themeKey.charAt(0).toUpperCase() + themeKey.slice(1);
+        let index = 0;
+        if (themeKey.indexOf(':') !== -1) {
+            index = themeKey.indexOf(':') + 1;
+        }
+        const cdvConfigPrefKey = 'Android' + themeKey.charAt(index).toUpperCase() + themeKey.slice(index + 1);
         const cdvConfigPrefValue = platformConfig.getPreference(cdvConfigPrefKey, this.platform);
         let themeTargetNode = splashScreenTheme.find(`item[@name="${themeKey}"]`);
-
+        console.log('hola', cdvConfigPrefKey);
```
dpogue, line 392: stray console.log

erisu (changes requested):
I think we also need to add `tools:targetApi="31"` to the node that gets created for this one property.

I remember having issues without this.

lib/prepare.js
```diff
@@ -378,14 +378,18 @@ function updateProjectSplashScreen (platformConfig, locations) {
...
+        'android:windowSplashScreenBrandingImage',
         'windowSplashScreenIconBackgroundColor',
         'postSplashScreenTheme'
     ].forEach(themeKey => {
-        const cdvConfigPrefKey = 'Android' + themeKey.charAt(0).toUpperCase() + themeKey.slice(1);
+        let index = 0;
+        if (themeKey.indexOf(':') !== -1) {
+            index = themeKey.indexOf(':') + 1;
+        }
+        const cdvConfigPrefKey = 'Android' + themeKey.charAt(index).toUpperCase() + themeKey.slice(index + 1);
```
erisu, line 389: Not sure if this is better performance or eaiser to read, but what about this?

```js
const themeKeyDescoped = themeKey.replace('android:', '');
const cdvConfigPrefKey = 'Android' + themeKeyDescoped.charAt(0).toUpperCase() + themeKeyDescoped.slice(1);
```

jcesarmobile: I have removed the if as it's not really necessary, so the code is simpler now, let me know if you still prefer to use the replace.

lib/prepare.js
jcesarmobile: wops, sorry, shouldn't commit so late, I was pretty sure I deleted all the console messages

jcesarmobile:
Android Studio underlines the `android:windowSplashScreenBrandingImage` in red warning about it being a target SDK 31+ feature, but still compiles, so not sure if we need to fix it.
While at the moment the `windowSplashScreenBrandingImage` is red and fails to compile.

lib/prepare.js
erisu: What you have is good. Dont need to use replace.

leMaik:
It's been a while; is there any chance we can get this released? :pray:

PR #1490, opened 2022-09-12
fix(splash-screen): support navigator.splashscreen.hide w/ AutoHideSplashScreen set to true

### LINKED ISSUE #1463: navigator.splashscreen.hide() issue in cordova-android v11.0.0
# Bug Report

## Problem

I recently updated Cordova and the Android platform plugin both to v11.0.0, adjusted my config.xml to use `AndroidWindowSplashScreenAnimatedIcon`, removed old splash screen entries (except the ones mentioned below) and made a **clean build** of my project (Node.js 14).
My app was previously able to close the splash screen manually so I had `AutoHideSplashScreen` set to `true`, `SplashScreenDelay` set to `10000` and simply used `navigator.splashscreen.hide()` to close the splash screen when my app is ready (depending on the device and connection quality 1-4s). In the worst case it would close after 10s.

`navigator.splashscreen.hide()` is supposed to close the splash screen before `SplashScreenDelay` expires.

The splash screen always stays for the full `SplashScreenDelay` time.

## Information

- I've tried different settings for `AutoHideSplashScreen` and `SplashScreenDelay` but nothing worked
- I can see log entries of my app trying to call `navigator.splashscreen.hide()` after a while with no effect
- I've done some remote debugging of the app using Chrome and tried to call the hide function manually with no effect (and no error message). Interestingly the remote screen is blank as long as the splash screen is open, but the dev console clearly shows that my app finished loading (including login of the user).
- The exact same code was working in Cordova 9 with Android platform plugin v8.1.0

### Command or Code

Set the `SplashScreenDelay` to 60s and call `navigator.splashscreen.hide()` after device-ready but while the splash screen is still showing.

I've tested the Android Emulator with Android 12.1, (real) Samsung mobile with Android 8, (real) Samsung tablet with Android 11.
Every device has the same issue.

### Version information

Cordova: "Pure" Cordova v11.0.0 with Android platform v11.0.0, no Ionic, no Capacitor
Windows 11, Android Studio 2021.2.1 Patch 1

## Checklist

erisu:

Allow developers to call the `.hide()` method while also using automatic hide settings

Closes #1463

### Testing

None

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/SplashScreenPlugin.java
+++ b/framework/src/org/apache/cordova/SplashScreenPlugin.java
@@ -67,6 +67,11 @@ public class SplashScreenPlugin extends CordovaPlugin {
      */
     private boolean keepOnScreen = true;
 
+    /**
+     * @param boolean delayHandlerCondition flag determins if the handler should continut to run
+     */
+    private boolean delayHandlerCondition = true;
+
     @Override
     protected void pluginInitialize() {
         // Auto Hide & Delay Settings
@@ -92,11 +97,8 @@ public class SplashScreenPlugin extends CordovaPlugin {
         JSONArray args,
         CallbackContext callbackContext
     ) throws JSONException {
-        if (action.equals("hide") && autoHide == false) {
-            /*
-             * The `.hide()` method can only be triggered if the `splashScreenAutoHide`
-             * is set to `false`.
-             */
+        if (action.equals("hide")) {
+            delayHandlerCondition = false;
             keepOnScreen = false;
         } else {
             return false;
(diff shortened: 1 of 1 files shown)


breautek (approved):

framework/src/org/apache/cordova/SplashScreenPlugin.java
```diff
@@ -127,8 +129,11 @@ private void setupSplashScreen(SplashScreen splashScreen) {
 
         // auto hide splash screen when custom delay is defined.
         if (autoHide && delayTime != DEFAULT_DELAY_TIME) {
-            Handler splashScreenDelayHandler = new Handler();
-            splashScreenDelayHandler.postDelayed(() -> keepOnScreen = false, delayTime);
+            Handler delayHandler = new Handler();
+            delayHandler.postDelayed(
+                () -> delayHandlerCondition ? keepOnScreen = false : null,
```
breautek, line 134: This is causing a build to fail

```
/home/runner/work/cordova-android/cordova-android/framework/src/org/apache/cordova/SplashScreenPlugin.java:134: error: incompatible types: lambda body is not compatible with a void functional interface
                () -> delayHandlerCondition ? keepOnScreen = false : null,
> Task :CordovaLib:compileDebugJavaWithJavac
```

I don't really understand Java's lambda syntax but I think it might be trying to return a value, when it expects a `void` return value.

breautek (changes requested):

PR #1493, opened 2022-09-19
chore: Add lint:fix script for fixing lint errors

jcesarmobile:
At the moment we have the lint script that tells you all the lint errors the code has, but you have to fix all the problems manually, which is tedious.
Since we are already using eslint for the linting, we can use the `--fix` parameter to automatically fix the errors (or at least most of them).

With this PR users will be able to run `npm run fmt` (I'm open for using other name), and fix the lint errors without having to manually change the code.

We could probably add this to all repositories that are already using eslint.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,6 +15,7 @@
     "prepare": "cordova-js build > templates/project/assets/www/cordova.js",
     "test": "npm run lint && npm run cover && npm run java-unit-tests",
     "lint": "eslint lib spec test \"templates/cordova/**/!(*.*)\"",
+    "fmt": "eslint lib spec test \"templates/cordova/**/!(*.*)\" --fix",
     "unit-tests": "jasmine --config=spec/unit/jasmine.json",
     "cover": "nyc jasmine --config=spec/coverage.json",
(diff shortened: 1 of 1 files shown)


erisu:
`npm run lint -- --fix` is what I am using.

erisu:
If you must have a command added to `package.json`, then I suggest

```json
"lint:fix": "npm run lint -- --fix",
```

So the command becomes `npm run lint:fix`.

breautek:
On windows command line / Powershell, you might have to encapsulate the intermediate `--` with double quotations so it's `npm run lint "--" --fix`. I vaguely recall that being an issue for windows users trying to pass through the `--packageType` flag to the cordova build command.

Not sure if the same syntax works for bash shells, but if we do add a command, it probably should work in both windows cmd line as well as bash.

jcesarmobile:
tested the command erisu proposed and worked fine on macOS, windows CMD and windows powershell, the quotation might be needed for when you type the commands yourself, but seem to work fine if the command is in a npm script

erisu (approved):

PR #1538, opened 2023-01-11
feat!: bump target sdk & build tools for SDK 33 support

erisu:
### Motivation and Context

Set latest Target SDK & Build Tools Version.

### Description

- Target SDK Version 33
- Minimum Build Tools Version 33.0.2

### Testing

n/a

### Checklist


breautek (approved):

jcesarmobile (approved):

erisu:
@jcesarmobile It seems that build tools `33.0.2` is out. Should I bump?

Or do you think it would be better to keep it as is or even lower the minimum to `33.0.0` instead?

jcesarmobile:
Probably

jcesarmobile:
I don’t know, since it’s the minimum allowed maybe 33.0.0, but both 33.0.0 and 33.0.1 present bugs that were fixed in 33.0.2, so not sure if we should require 33.0.2 as minimum

breautek:
We generally discourage people from updating the build tools version unless absolutely necessary/know what they are doing so we probably should use the latest patch available, so that cordova by default will use the latest patch available (at the time of this PR) obviously.

33.0.1 resolves a compilation issue:

> AIDL fails with build-tools 33.0.0. (See [issue #236167971](https://issuetracker.google.com/issues/236167971))

It probably doesn't affect the cordova base but could affect other native libraries or third-party plugins.

33.0.2 addresses what appears to be a minor issue:

> core-lambda-stubs.jar version 33.0.0 differs on macOS versus Linux/Windows. (See [issue #237299698](https://issuetracker.google.com/issues/237299698).)

Nonetheless, I think we should just use the latest patch available, so that on release we can cleanly write an update guide stating they must have `33.0.2` installed, etc. It may save us having to explain that they need to set a preference to upgrade build tools to address `AIDL` issue if that issue affects them.

Just my own opinion. If they absolutely need to use 33.0.0 for some reason, I believe they can still set the `android-buildToolsVersion` themselves.

jcesarmobile:
no, they can't use an older version since this value is for the minimum version allowed and the code has a check that prevents from using older versions that the one we set there

but I'm ok making the minimum the latest at the moment of the release

PR #1543, opened 2023-01-11
feat: bump kotlin 1.9.24 & drop kotlin-android-extensions when kotlin >= 1.8.0

### LINKED ISSUE #1642: Configure the need of `kotlin-android-extensions` plugin to support Kotlin 1.9
# Feature Request
Kotlin 1.9 does not make use of `kotlin-android-extesions` plugin which is still being generated by the latest cordova-android 12.0.0 version and thus android app build fails.
It would be nice to add another preference apart of `{ xmlKey: 'GradlePluginKotlinEnabled', gradleKey: 'IS_GRADLE_PLUGIN_KOTLIN_ENABLED', type: Boolean }` to be able to configure the need of `kotlin-android-extensions` plugin for the android app

## Motivation Behind Feature

Kotlin 1.9 does not work out of the box with the latest Cordova 12.0.0 framework. It requires manual interference to remove the line `apply plugin 'kotlin-android-extensions'` from the /platforms/android/app/build.gradle file.

## Feature Description
Add new xml key and preference { xmlKey: 'GradlePluginKotlinExtEnabled', gradleKey: 'IS_GRADLE_PLUGIN_KOTLIN_EXT_ENABLED', type: Boolean } and apply the `kotlin-android-extesions` plugin only in case the property is set to true.

### LINKED ISSUE #1661: Support AGP 8.1+ by using JVM 17
# Feature Request
Support latest build tool set like Gradle 8.2+, AGP 8.1+, Kotlin 1.9 (already addressed in https://github.com/apache/cordova-android/issues/1642)

## Motivation Behind Feature

Starting from AGP 8.1.0 Android ships with JDK 17 bytecode.
https://developer.android.com/build/releases/gradle-plugin#8-1-0

Current cordova-android build.gradle template hardcodes the JVM toolchain to the version 1.8 which is incompatible with the latest AGP plugin:

```groovy
android {
...
    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }
...
```

## Feature Description
There are multiple ways to support AGP 8.1+:
- Make JVM version customisable so a developer can set it from `config.xml` (a bit more work, I am not sure whether it's going to be beneficial). Or
- Make a hard switch to JVM 17:
```groovy
    compileOptions {
        sourceCompatibility JavaVersion.VERSION_17
        targetCompatibility JavaVersion.VERSION_17
    }
```

### LINKED ISSUE #1692: Will Cordova Android 13 support Kotlin 1.9.x ?
In my company, we integrate vendor's SDK to our cordova application. The vendor request us to raise the version of Kotlin to 1.9.x but Cordova Android has not support is now. 
If possible, could you tell me whether Cordova Android 13 will support Kotlin 1.9.x  or not.

erisu:
### Motivation and Context

Use latest Kotlin

resolves #1692
resolves #1661 (Kotlin)
resolves #1642

### Description

- Bump Kotlin version `1.9.24`
- Do not apply `kotlin-android-extensions` plugin when Kotlin version is `>= 1.8.0`

Default projects do not enable Kotlin. If Kotlin is enabled, with the config `preference` `GradlePluginKotlinEnabled`, the default version `1.9.24` is applied.

```xml
<preference name="GradlePluginKotlinEnabled" value="true" />
```

App developers can change the Kotlin version with the following config `preference` `GradlePluginKotlinVersion`.

Example:

```xml
<preference name="GradlePluginKotlinVersion" value="1.7.22" />
```

* If the Kotlin version is lower than 1.8.0, the `kotlin-android-extensions` plugin is applied to maintain previous implementation.
* If Kotlin version is greater than or equal to 1.8.0, then the `kotlin-android-extensions` plugin is not applied as it was deprecated.

**Kotlin Release Notes**

[Kotlin's Release](https://kotlinlang.org/docs/releases.html#update-to-a-new-release) says:

> Starting from IntelliJ IDEA 2023.3 and Android Studio Iguana (2023.2.1) Canary 15, the Kotlin plugin is automatically updated. All you need to do is update the Kotlin version in your projects.

> For Android Studio Giraffe and Hedgehog, the Kotlin plugin 1.9.10 will be delivered with upcoming Android Studios updates.

### Testing

- `npm t`
- Cordova create, add platform, and build
- Changed `GradlePluginKotlinVersion` to `1.7.22` to test ensure `kotlin-android-extensions` is applied.

### Checklist


breautek:
Kotlin has a compatibility guide at https://kotlinlang.org/docs/compatibility-guide-18.html

It seems like they don't follow the "semver" rules and breaking changes may be introduced in "minor" versions.

As an example, some things changed with the `when` syntax as documented: https://kotlinlang.org/docs/compatibility-guide-18.html#deprecate-confusing-grammar-in-when-with-subject

```
1.6.20: introduce a deprecation warning on the affected expressions

1.8.0: raise this warning to an error, -XXLanguage:-ProhibitConfusingSyntaxInWhenBranches can be used to temporarily revert to the pre-1.8 behavior

>= 1.9: repurpose some deprecated constructs for new language features
```

There's a lot there, so given this information, if the Android SDK was built with compatibility of a specific kotlin version, we should try to follow that as our default as much as possible.

breautek:
Not sure how much this is relevant but there is also a Jetpack Compose Compiler package that does have explicit compatibility notes regarding kotlin

https://developer.android.com/jetpack/androidx/releases/compose-kotlin

For example, their latest release v1.4.0 introduces support for Kotlin 1.8.0.

kbrownlees:
As a data point I recently had to upgrade my kotlin version to 1.7 because our builds started failing with a version mismatch issue. 

Trying to use 1.8 failed to build, with the error:

```
e: The Android extensions ('kotlin-android-extensions') compiler plugin is no longer supported. Please use kotlin parcelize and view binding. More information: https://goo.gle/kotlin-android-extensions-deprecation
```

For context, to change you kotlin version set `<preference name="GradlePluginKotlinVersion" value="1.7.22"/>` in `config.xml`.

petercutting:
I ran in to this build error today

Class 'kotlin.Unit' was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.7.1, expected version is 1.5.1.

Really frustrating when these errors appear for no apparent reason. The kotlin fix above worked so it was atleast an easy fix, when I eventually found it here.

I wonder if this fix will cause a new problem down the road when 1.7.22 isnt compatible with something else

breautek:
I found another compatibility matrix in regards to gradle: https://docs.gradle.org/current/userguide/compatibility.html

As it stands right now, kotlin 1.8 is only officially supported on Gradle 8, which we aren't using yet (because it will require AGP 8).

I know folks already stated compatibility issues with kotlin 1.8, this is just a data point that might explain why there are compatibility issues.

breautek (approved):
Looks good, but I think this PR might be blocked by https://github.com/apache/cordova-android/pull/1678

Because that PR bumps Gradle to 8.7 which I think is a [requirement](https://docs.gradle.org/current/userguide/compatibility.html#kotlin) for >= kotlin 1.9.22.

PR #1571, opened 2023-03-15
feat!: Bump min SDK to 24

breautek:
Rationale:
API 22 & API 23 both account for an insignificant part of the market share. While API 24 - API 26 has similar market share, we felt that bumping to API 26 from API 22 is too large of a jump.

Legacy devices may be completely out of support by Google and may not be able to receive the latest webview version. As of writing, Chromium's latest tag shows they are using a Min SDK version of 24. (Ref: https://chromium.googlesource.com/chromium/src/+/refs/tags/113.0.5653.1/build/config/android/config.gni#46)

Based on AOSP emulators, API 24 (Android 7.0) will ship with Chrome 52 webview, which has good support for ECMAscript 2015 (ES6) (Ref: https://caniuse.com/?search=es6)

While in most cases, app users will likely have a modern webview vesion installed, this means you can be confident that the app user will have a chrome webview version with good ES6 support, even if they happen to be running on a factory versioned device.

See the mailing thread for the full discussion:
https://lists.apache.org/thread/zcgof080hdzzo2j96mjz0qpj0gotmn57

### Checklist

### DIFF
--- a/framework/cdv-gradle-config-defaults.json
+++ b/framework/cdv-gradle-config-defaults.json
@@ -1,5 +1,5 @@
 {
-    "MIN_SDK_VERSION": 22,
+    "MIN_SDK_VERSION": 24,
     "SDK_VERSION": 33,
     "COMPILE_SDK_VERSION": null,
     "GRADLE_VERSION": "7.4.2",


brodybits:
> API 22 & API 24

nit: I think you meant "API 22 & API 23", and think it would be ideal if we could get this fixed in the commit message as well.

erisu (approved):
LGTM

jcesarmobile (approved):

PR #1574, opened 2023-03-21
feat: add plugin hooks for WebViewClient.onRenderProcessGone

peitschie:
### Platforms affected
Android

### Motivation and Context
This change allows more advanced handling of the webview render process termination. This can be useful for capturing additional information, or even conceivably allow the crash to be prevented should a plugin choose to.

### Description
Expose the [WebViewClient.onRenderProcessGone](https://developer.android.com/reference/android/webkit/WebViewClient#onRenderProcessGone(android.webkit.WebView,%20android.webkit.RenderProcessGoneDetail)) method down to plugins, and allow a plugin to interrupt the crash should it choose so.

### Testing
I created the hello world cordova application and added a javascript call to load `chrome://crash` (recommended by Android docs as the best way to trigger a renderer crash).

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/CordovaPlugin.java
+++ b/framework/src/org/apache/cordova/CordovaPlugin.java
@@ -31,6 +31,8 @@ import android.content.res.Configuration;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
+import android.webkit.RenderProcessGoneDetail;
+import android.webkit.WebView;
 
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -441,5 +443,18 @@ public class CordovaPlugin {
      */
     public CordovaPluginPathHandler getPathHandler() {
         return null;
+    }
+
+    /**
+     * Called when the WebView's render process has exited.
+     *
+     * See <a href="https://developer.android.com/reference/android/webkit/WebViewClient#onRenderProcessGone(android.webkit.WebView,%20android.webkit.RenderProcessGoneDetail)">WebViewClient#onRenderProcessGone</a>
+     *
+     * @return  true if the host application handled the situation that process has exited,
+     *          otherwise, application will crash if render process crashed, or be killed
+     *          if render process was killed by the system.
+     */
+    public boolean onRenderProcessGone(final WebView view, RenderProcessGoneDetail detail) {
+        return false;
     }
 }
--- a/framework/src/org/apache/cordova/CordovaPlugin.java
+++ b/framework/src/org/apache/cordova/CordovaPlugin.java
@@ -446,9 +446,11 @@ public class CordovaPlugin {
     }
 
     /**
-     * Called when the WebView's render process has exited.
+     * Called when the WebView's render process has exited. Can be used to collect information regarding the crash for crashlytics or optionally attempt to gracefully handle/recover the crashed webview by recreating it.
      *
      * See <a href="https://developer.android.com/reference/android/webkit/WebViewClient#onRenderProcessGone(android.webkit.WebView,%20android.webkit.RenderProcessGoneDetail)">WebViewClient#onRenderProcessGone</a>
+     *
+     * Note: A plugin must not attempt to recover a webview that it does not own/manage.
      *
      * @return  true if the host application handled the situation that process has exited,
      *          otherwise, application will crash if render process crashed, or be killed
--- a/framework/src/org/apache/cordova/PluginManager.java
+++ b/framework/src/org/apache/cordova/PluginManager.java
@@ -32,6 +32,8 @@ import android.net.Uri;
 import android.os.Bundle;
 import android.os.Debug;
 import android.os.Build;
+import android.webkit.RenderProcessGoneDetail;
+import android.webkit.WebView;
 
 /**
  * PluginManager is exposed to JavaScript in the Cordova WebView.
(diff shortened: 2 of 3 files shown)


lgh06:
Nice work. I face webview hang , found no solution. I will try your cordova-android and write a plugin

breautek:
> or even conceivably allow the crash to be prevented should a plugin choose to.

Given that the Android Docs states

> The given WebView can't be used, and should be removed from the view hierarchy, all references to it should be cleaned up, e.g any references in the Activity or other classes saved using [View.findViewById(int)](https://developer.android.com/reference/android/view/View#findViewById(int)) and similar calls, etc.

I think it's unsafe to attempt to "prevent" a crash or continue using the webview. I think at minimum the webview needs to be recreated. I do agree that it is definitely useful for debugging or crashlytics however.

The `RenderProcessGoneDetail` API was added in API 26. cordova-android@12 supports as far back as API 24, but we will compile with API 33. Because this is a delegate function, I believe this is fine, it just won't be invoked by the underlying system on API 24 or 25, I presume. Was this tested on API 24 devices/emulators?

breautek (commented):
Tested both on API 24 and API 33 emulators, both of which I believe works as expected.

On API 24, the hook is simply not invoked and the process will simply crash.

On API 33, the hook is invoked before the crash, allowing the system to gracefully handle the crash and to potential recover from it.

framework/src/org/apache/cordova/PluginManager.java
```diff
@@ -617,4 +618,29 @@ public ArrayList<CordovaPluginPathHandler> getPluginPathHandlers() {
...
+     *          otherwise, application will crash if render process crashed, or be killed 
+     *          if render process was killed by the system.
+     */
+    public boolean onRenderProcessGone(RenderProcessGoneDetail detail) {
+        boolean result = false;
+        synchronized (this.entryMap) {
+            for (PluginEntry entry : this.entryMap.values()) {
+                CordovaPlugin plugin = pluginMap.get(entry.service);
+                if (plugin != null) {
+                    if (plugin.onRenderProcessGone(app, detail)) {
```
breautek, line 637: Here we are assuming the affected webview is indeed the cordova webview.

> Multiple WebView instances may be associated with a single render process; onRenderProcessGone will be called for each WebView that was affected. The application's implementation of this callback should only attempt to clean up the specific WebView given as a parameter, and should not assume that other WebView instances are affected.

I don't believe this is a safe assumption. While as far as Cordova is concerned, there is only 1 webview instance, however plugins may add additional instances (e.g. the in app browser plugin). Therefore I believe it is possible for a webview to crash in which is not the main cordova webview, but will still go through this hook.

Therefore, I think a better approach here is to pass down the `WebView` instance down and let plugins decide if it's the webview it cares about.

The `CordovaAndroid` plugin (the android core) may use this to check if the crashed webview is indeed the main webview, while the InAppBrowser plugin may use this reference to determine if it's one of the webviews it has created.

peitschie: That's a good point @breautek!

I've made the change so that the original webview is passed down instead.
There may be some plugins that want to listen to every crash (e.g., a crash report integration plugin of some sort), but would theoretically allow the core plugins to be extended to recreate the webview after a crash.

peitschie:
> I think it's unsafe to attempt to "prevent" a crash or continue using the webview. I think at minimum the webview needs to be recreated. I do agree that it is definitely useful for debugging or crashlytics however.

Yep! Agree that more work would be needed if some kind of crash recovery was needed... returning `true` here would not be a good thing to do 😅 

I've tweaked this a bit, so should be good for another review when you have time @breautek

breautek:
Alright, had to do some tweaks in cordova-android to test it, but at the core I think this is good.

I implemented the on hook in `CoreAndroid` and was able to do something like:

```java
@Override
    public boolean onRenderProcessGone(final WebView view, RenderProcessGoneDetail detail) {
        if (webView.getEngine().getView() == view) {
            // Is the crash view the cordova webview?
            LOG.e(TAG, "CORDOVA WEBVIEW CRASHED!");
            return true;
        }
        return false;
    }
```

Naturally returning true prevents the app from crashing and actually yields the webview still visible, but in an unsafe manner. Returning false will crash the application, assuming nothing else actually handles the webview. To be clear, I'm not suggesting to implement something like this into this PR, I'm just using the PR to demonstrate what is possible.

> There may be some plugins that want to listen to every crash (e.g., a crash report integration plugin of some sort),

I do agree with this sentiment. I see that PluginManager will still call through all plugins hooks even if a plugin has already return true, so I think that covers that.

framework/src/org/apache/cordova/CordovaPlugin.java
```diff
@@ -442,4 +444,17 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions,
...
     }
+
+    /**
+     * Called when the WebView's render process has exited.
+     *
+     * See <a href="https://developer.android.com/reference/android/webkit/WebViewClient#onRenderProcessGone(android.webkit.WebView,%20android.webkit.RenderProcessGoneDetail)">WebViewClient#onRenderProcessGone</a>
+     *
+     * @return  true if the host application handled the situation that process has exited,
+     *          otherwise, application will crash if render process crashed, or be killed
+     *          if render process was killed by the system.
```
breautek, line 455: Just a nitpick here,

I think we should be explicit that plugins must not attempt to gracefully recover from a webview crash on a webview that it does not own/manage. So in other words the plugins must not return `true` for a webview that it doesn't manage.

Otherwise we open ourselves in the future of plugins conflicting and if they do that disregarding the note, then that's on them.

```suggestion
     * Called when the WebView's render process has exited. Can be used to collect information regarding the crash for crashlytics or optionally attempt to gracefully handle/recover the crashed webview by recreating it.
     *
     * See <a href="https://developer.android.com/reference/android/webkit/WebViewClient#onRenderProcessGone(android.webkit.WebView,%20android.webkit.RenderProcessGoneDetail)">WebViewClient#onRenderProcessGone</a>
     *
     * Note: A plugin must not attempt to recover a webview that it does not own/manage.
     *
     * @return  true if the host application handled the situation that process has exited,
     *          otherwise, application will crash if render process crashed, or be killed
     *          if render process was killed by the system.
```

breautek: Not sure if github UI messed with the formatting, so if the formatting got broken, then I apologize.

breautek (approved):

peitschie:
Thanks for the review and merge @breautek 🍻 

Agree with the wording change there. We definitely want plugins to leave the return value alone unless they really know what they're doing!

erisu (approved):
LGTM

breautek:
Awesome work, thanks @peitschie

This feature will be available in the upcoming cordova-android@12 release.

PR #1583, opened 2023-04-03
GoogleMaps temporal fix

efidatansantana:
Deleting " assert this.cordova == null", the maps shows, this is a temporal solution and not recomended form prod release.

### Checklist


breautek:
> Deleting " assert this.cordova == null", the maps shows, this is a temporal solution and not recommended form prod release.

If you believe that the code change is not recommended for production use, then we certainly shouldn't accept the PR. Am I misunderstanding something?

efidatansantana:
yes you have right, this is not recommended to production use, but is a easy solution for prevent the error momentarily, because I did not find another solution either

Issue #1409, opened 2022-03-29
Apple M1 Support


fabltd:
Hi 

I cannot seem to get cordova-andriod to work on an M1 Mac.

The requirement is for JAVA-11 there is no M1 native verison.

There first M1 native version of JAVA is 17/18

What is required to get Cordova-Andriod to reconise a later JAVA version?

Thank you

timbru31:
As of cordova-android@10 we've removed the Java version check.
What's the version of cordova-android you are using?

fabltd:
Hi 

I updated to V11 and Java 18 but I cannot build I get this error:

* Where:
Settings file 'xxxxxxxxxx/platforms/android/settings.gradle'

* What went wrong:
Could not compile settings file 'xxxxxxx/android/settings.gradle'.
> startup failed:
  General error during conversion: Unsupported class file major version 62
  
  java.lang.IllegalArgumentException: Unsupported class file major version 62

Then lots of files that will not compile followed by:

Deprecated Gradle features were used in this build, making it incompatible with Gradle 8.0.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

See https://docs.gradle.org/7.1.1/userguide/command_line_interface.html#sec:command_line_warnings

breautek:
> I updated to V11 and Java 18 but I cannot build I get this error:

Afaik, Google only supports Java 8 to 11 on API 30 or lower, and requires Java 11 on API 31 or later.

That error you posted was on a gradle file, and Gradle 7.x only supports up to Java 17. [Gradle](https://docs.gradle.org/current/userguide/compatibility.html) explicitly states that Java 18 is not supported yet.

It's not Cordova that sets these requirements, and like Tim said, as of cordova-android@10 we removed any java version checks that was applied in previous cordova-android versions.

If you can't get/run Java 11 on M1 macs, then that is a problem Cordova cannot solve.

Perhaps this [SO Answer](https://stackoverflow.com/a/68651120/4685664) will help you get a Java 11 implementation via OpenJDK running on M1 mac, if you're fine with OpenJDK licensing.

fabltd:
Thanks

I got further by installing OpenJDK 11

Now have a different error:

android/CordovaLib/cordova.gradle' line: 69

* What went wrong:
A problem occurred evaluating script.
> No installed build tools found. Install the Android build tools version 30.0.3 or higher.

I do however have the build tools installed.

Any suggestions?

[...]

PR #1596, opened 2023-04-14
feat!: Upgrade to Gradle and AGP 8

jcesarmobile:


erisu:
@jcesarmobile try updating this line to use JDK 17. Include title change.

https://github.com/apache/cordova-android/blob/master/.github/workflows/ci.yml#L41-L45

jcesarmobile:
updated gradle to 8.0.2 as gradle website recommends using that over 8.0

breautek:
The [README](https://github.com/apache/cordova-android/blob/master/README.md?plain=1#L35) should be updated to reflect the new JDK requirement.

Part of me would rather wait for AGP 8 support until we intend to support API 34, which is still in beta phase and won't be released until August at the earliest. I see no concrete date in their roadmap for the final release. The rationale for waiting is to defer the JDK 17 upgrade when it's not needed for a future cordova-android@13 release for API 34. I think Cordova should release a version supporting API 33 no later than at the end of July to give the month to users to refactor their own apps.

On the other hand, we do have an issue with M1 macs where it can be rather [difficult](https://github.com/apache/cordova-android/issues/1409#issuecomment-1510008696) to setup due to the current JDK 11 requirement which the only prebuilt JDK available and is provided through Android Studio Electric Eel. Now that Flamingo is released, I presume it ships with JDK 17 which I've seen reports that it doesn't work properly with the Android Tooling (however I haven't confirmed this myself). So having build tools that supports JDK 17 will now give M1 users more options as there are other prebuilt JDK environments available and would be more aligned to our existing documentation.

breautek:
Also related: https://docs.gradle.org/7.6/userguide/upgrading_version_7.html#configuring_custom_build_layout

We use `-b` flag currently to specify where our build config is. Not sure if this is even necessary however.

breautek:
> Do you recommand sir @breautek to downgrade or to wait for the new release ? (my company is in harry to deploy their first app on the store , so maybe we gonna traget sdk 32 only)

To add to jcesar's comments, cordova-android@12 is not going to support AGP 8 or JDK 17. JDK 17 will probably still work for the most part, but it will probably be best to use the recommended JDK for AGP 7 which is JDK 11. JDK 17 may not work completely as expected when using it with AGP 7.

The namespace attribute came in AGP 7.4 or 7.5 (I forget which version exactly) but cordova-android@12 will configure the app & cordova framework to use the `namespace` attribute.

cordova-android@12 is currently being voted on for release (See the [Vote Thread](https://lists.apache.org/thread/f0rgv8dc66lxsq0hnfkdny928ko4yc4d)). So if you gave nightly a test, consider giving us feedback in the [DISCUSS](https://lists.apache.org/thread/f0rgv8dc66lxsq0hnfkdny928ko4yc4d) thread and maybe a +1 or a -1 in the [VOTE](https://lists.apache.org/thread/f0rgv8dc66lxsq0hnfkdny928ko4yc4d) thread.

If you need more information on how Apache voting works, see [Apache Voting Process](https://www.apache.org/foundation/voting.html).

On AGP 8 / Gradle 8 support, as of writing I think that is planned for another major release, probably cordova-android@13.

PR #1605, opened 2023-04-21
fix!: Make CoreAndroid plugin instantiate on load

breautek:
I don't anticipate breaking changes from this change, however it is a difference in behaviour since CoreAndroid won't be lazily loaded, therefore I've marked this commit has a breaking change.

### Platforms affected

Android / CoreAndroid plugin

### Motivation and Context

CoreAndroid was lazily loaded via  `getPlugin` call. This can create a race condition in some situations, particularly 
with a new feature in the works: https://github.com/apache/cordova-android/pull/1574 where onRenderProcessGone may be invoked before anything gets a chance to call `getPlugin` for `CoreAndroid`.

In general, because this is **Core** plugin, I feel like it should always be loaded in, not lazily loaded in.

### Description

Inside `PluginManager`, I added a new public API: `addService(String serviceName, String className, bool onload)`.
The `addService(String serviceName, String className)` implementation has been moved to the new public API.

`CordovaWebviewImpl` will now add the CoreAndroid service with `onload` set to true.

### Testing

Ran `npm test` on linux.
Also manual test while testing the onRenderProcessGone PR.

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/CordovaWebViewImpl.java
+++ b/framework/src/org/apache/cordova/CordovaWebViewImpl.java
@@ -115,7 +115,7 @@ public class CordovaWebViewImpl implements CordovaWebView {
         // This isn't enforced by the compiler, so assert here.
         assert engine.getView() instanceof CordovaWebViewEngine.EngineView;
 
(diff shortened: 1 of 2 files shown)


dpogue (approved):
LGTM

Issue #1604, opened 2023-04-20
--gradleArg=--stacktrace does not appear to work


treecaptcha:
# Bug Report

## Problem
`--gradleArg=--stacktrace` does not appear to work

### What is expected to happen?
Gradle prints stacktrace when encountering error when `--gradleArg=--stacktrace` is passed

### What does actually happen?
Gradle does not print a stack trace and it recommends the `--stacktrace` argument

## Information

https://cordova.apache.org/docs/en/latest/guide/platforms/android/index.html#configuring-gradle

### Command or Code

`cordova build android --gradleArg=--stacktrace`
```
treecaptcha@TreeCaptcha:~/gitstuff/MyApp$ cordova build android --gradleArg=--stacktrace
Checking Java JDK and Android SDK versions
ANDROID_SDK_ROOT=undefined (recommended setting)
ANDROID_HOME=undefined (DEPRECATED)
Using Android SDK: /usr/lib/android-sdk

FAILURE: Build failed with an exception.

* What went wrong:
Could not create service of type ScriptPluginFactory using BuildScopeServices.createScriptPluginFactory().
> Could not create service of type PluginResolutionStrategyInternal using BuildScopeServices.createPluginResolutionStrategy().

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 0s
Command failed with exit code 1: /usr/share/gradle/bin/gradle -p /home/treecaptcha/gitstuff/MyApp/platforms/android wrapper -b /home/treecaptcha/gitstuff/MyApp/platforms/android/wrapper.gradle
```

### Environment, Platform, Device

OS: Ubuntu 20.04.5 LTS x86_64 
Host: XPS 8930 1.1.19 
Kernel: 5.15.0-67-generic 
Uptime: 25 days, 47 mins 
Packages: 3823 (dpkg), 14 (snap) 
Shell: bash 5.0.17 
Resolution: 3440x1440 
DE: Plasma 
WM: KWin 
WM Theme: Breeze 
Theme: Breeze [Plasma], Breeze [GTK2/3] 
Icons: breeze [Plasma], breeze [GTK2/3] 
Terminal: gnome-terminal 
CPU: Intel i5-9400 (6) @ 4.100GHz 
GPU: NVIDIA GeForce RTX 2070 
GPU: Intel UHD Graphics 630 
[...]

PR #1606, opened 2023-04-21
fix: Gradle Args parsing

breautek:
### Platforms affected

Android / Gradle Argument pass through

### Motivation and Context

gradle expects a specific usage which Cordova doesn't conform to.

```
$ gradle --help

USAGE: gradle [option...] [task...]
```

Currently we append the gradle arguments at the end of the gradle argument list, but option arguments are suppose to appear before the task(s).

Closes https://github.com/apache/cordova-android/issues/1604

Additionally, while testing I also discovered that arguments that requires multiple values (e.g. `--warning-mode all`) doesn't get passed through properly, yielding an argument array that looks like:

```javascript
[ "--warning-mode all" ]
````

when the expect array should look like:

```javascript
[ "--warning-mode", "all" ]
```

### Description

- Re-ordered the gradle argument array creation so that options appears before the task.
- Corrected tests who assumed the task will appear at index 0. Tasks will always appear at the end of the array now and have been updated accordingly with that specific expectation.
- Added a dependency to [string-argv](https://www.npmjs.com/package/string-argv). `nopt` does not fit here because it expects that you know what your CLI arguments are. However since these arguments are gradle arguments, it's better if we can parse them and pass them through. This package handles quotations and such to ensure a proper argv array is created.

### Testing

Ran `npm test` and manually tested in hello world app using commands such as:

`cordova build android --minSdkVersion 30 -- --gradleArg="--stacktrace"`

`cordova build android --minSdkVersion 30 -- --gradleArg="--stacktrace --warning-mode all"`

### Checklist

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -21,6 +21,7 @@ const path = require('path');
 const fs = require('fs');
 const nopt = require('nopt');
 const untildify = require('untildify');
+const ARGVParser = require('string-argv');
 
 const Adb = require('./Adb');
 
@@ -36,7 +37,11 @@ function parseOpts (options, resolvedTarget, projectRoot) {
         minSdkVersion: String,
         maxSdkVersion: String,
         targetSdkVersion: String,
+
+        // This needs to be an array since nopts will parse its entries as further options for this process
+        // It will be an array of 1 string: [ "string args" ]
         gradleArg: [String, Array],
+
         keystore: path,
         alias: String,
         storePassword: String,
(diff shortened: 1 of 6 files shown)


dpogue (commented):
This looks okay to me.

It would maybe be good to have a test case that covered the splitting of `gradleArgs` into multiple `extraArgs` tokens.

erisu (approved):
Just a nitpik but, looks good to me.

breautek:
> This looks okay to me.
> 
> It would maybe be good to have a test case that covered the splitting of `gradleArgs` into multiple `extraArgs` tokens.

I added a new `build.spec.js` unit test file since gradle argument parsing happens earlier than what `ProjectBuilder.spec.js` tests. I only added testing that was relevant in the PR but could probably be expanded further to match more closely to what `run.spec.js` does.

Added series of different test cases regarding using the `--gradleArg` flag and testing the parsed argv array, which would include the usage of `nopt` + `string-argv` third-party dependencies.

dpogue (approved):

PR #1385, opened 2021-12-21
(android)feat: Added camera select to file input tags

kortgat:
### Motivation and Context

Allowing the user to use the camera or file browser with a html file input tag

[https://github.com/apache/cordova-android/issues/816](https://github.com/apache/cordova-android/issues/816)

### Description

Reworked the action tray to use a chooser intent in cordova\engine\SystemWebChromeClient.java to add option for the camera

### Testing

Tested using a emulator and actual device to take photos and upload to my server as well as upload pdf files from file system.
I am also currently using this in production in my application.

### Checklist

- [ ] I've run the tests to see all new and existing tests pass
- [ ] I added automated test coverage as appropriate for this change
- [ ] Commit is prefixed with `(platform)` if this change only applies to one platform (e.g. `(android)`)
- [ ] If this Pull Request resolves an issue, I linked to the issue in the text above (and used the correct [keyword to close issues using keywords](https://help.github.com/articles/closing-issues-using-keywords/))
- [ ] I've updated the documentation if necessary

PR #1609, opened 2023-04-26
feat: add camera intent with file input capture

### LINKED ISSUE #816: Android clicking input type=file brings up system file browser, not camera options
# Bug Report

## Problem

When clicking on an html type="file" input a system file picker appears instead of camera options. 

ss

If I perform the same action in the system browser I get the expected dialog:

ss2

### Version information

```
"cordova-android": "^8.0.0",
"cordova-plugin-camera": "^4.1.0"
```

cordova-cli 9.0.0

Full [test case repo here](https://bitbucket.org/alex-crooks/android-file-test/src/master/)

## Checklist

KenCorbettJr:
### Platforms affected

Android

### Motivation and Context
Allowing the user to use the camera or file browser with an HTML file input tag.  This is possible on the mobile web and in Cordova apps on other platforms, but on android, the only option is to choose from images in the gallery.

Closes #816

### Description
Reworked the action tray to use a chooser intent in cordova\engine\SystemWebChromeClient.java to add an option for the camera.

### Testing
I developed the changes in a plugin (https://www.npmjs.com/package/cordova-plugin-android-image-file-input) that copies in the updated file.  The plugin works beautifully but is so hacky.  I shouldn't need to patch CordovaLib.  But the issues has been open since 2019, and another pull request (https://github.com/apache/cordova-android/pull/1385) has been open since 2021.

### Checklist


dpogue (commented):
It's a bit hard to review this due to whitespace and reformatting changes to other parts of the file. Is it possible to include only the changes related to file picking without the other reformatting?

KenCorbettJr:
Sure!  I went ahead and removed those from this pull request.

KenCorbettJr:
I didn't think it would make a difference, but just to be sure I went ahead and re-ran all my tests and the file input works beautifully.

templates/project/res/xml/opener_paths.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- https://developer.android.com/reference/android/support/v4/content/FileProvider.html#SpecifyFiles -->
+<paths xmlns:android="http://schemas.android.com/apk/res/android">
```
dpogue, line 3: Should this file be managed as part of the cordova-file-plugin (which provides API to exposes those paths) or is this also exposed by cordova-android directly?

framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
```diff
@@ -211,54 +216,137 @@ public View getVideoLoadingProgressView() {
...
+
+    public void openFileChooser(ValueCallback<Uri> uploadMsg, String acceptType) {
+        this.openFileChooser(uploadMsg, acceptType, null);
+    }
+
+    public void openFileChooser(final ValueCallback<Uri> uploadMsg, String acceptType, String capture) {
+        Intent intent = new Intent(Intent.ACTION_GET_CONTENT);
+        intent.addCategory(Intent.CATEGORY_OPENABLE);
+        intent.setType("*/*");
+        parentEngine.cordova.startActivityForResult(new CordovaPlugin() {
```
dpogue, line 235: This anonymous inner-class CordovaPlugin feels sketchy to me, but I see it's the same pattern that already exists so... I guess it's okay?

If you have ideas for a better way to handle this that has less potential for weird side-effects, I'd love to hear it.

framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
```diff
@@ -211,54 +216,137 @@ public View getVideoLoadingProgressView() {
         return mVideoProgressView;
     }
 
+    // <input type=file> support:
+    // openFileChooser() is for pre KitKat and in KitKat mr1 (it's known broken in
+    // KitKat).
+    // For Lollipop, we use onShowFileChooser().
```
dpogue, line 222: TODO: double check Cordova supported versions, I'm 99% sure we've dropped KitKat support already, and I'm pretty sure Lollipop is also out (or will be out in the upcoming major version bump)

breautek: Kitkat (Android 4.4 / API 19) was indeed dropped since cordova-android@10.
Lollipop (Android 5.0 / API 21) is being dropped in the next release in cordova-android@12, which will hvae a Min SDK of 24 (Android 7 / Nougat)

But I assume [onShowFileChooser](https://developer.android.com/reference/android/webkit/WebChromeClient#onShowFileChooser(android.webkit.WebView,%20android.webkit.ValueCallback%3Candroid.net.Uri[]%3E,%20android.webkit.WebChromeClient.FileChooserParams)) is to be used for API 21+, so I think it's still relevant. It's not deprecated.

erisu: Since Cordova-Android 9.0 the `minSdkVersion` is actually set to SDK 22 (Android 5.1) Lollipop MR1, not API 21.

KenCorbettJr: Nice.  Then we can just remove all this code.  I will delete it.

templates/project/res/xml/opener_paths.xml
dpogue: hmm, so the file needs to exist, and probably needs at least one path defined...

I guess one option would be for cordova-android to provide a default (mostly empty) implementation of this file, and then we update the cordova-plugin-file to use `edit-config` to add the rest of the paths at install time. That would only really work with major bumps to both cordova-android and cordova-plugin-file, but we're looking at a major for this repo anyways.

@erisu @breautek Thoughts? Are there downsides or security risks to including all these paths here by default? I'm not too familiar with Android's FileProvider stuff 😕

breautek: > I guess one option would be for cordova-android to provide a default (mostly empty) implementation of this file, and then we update the cordova-plugin-file to use edit-config to add the rest of the paths at install time.

As much as having cordova-plugin-file actually handle this kind of stuff, and it using `<edit-config>` to change the config where necessary might be the "correct" approach... `edit-config` directive is incredibly broken... in the sense that it's overly sensitive to detecting conflicts. So if two things are editing the same node, there's a high chance they won't work or be compatible together, even if they aren't touching or editing the same attribute.

As for security risks... I'm not sure, I'm not entirely proficient in this area of Android either. So I'll need to do some reading. I do believe setting up a `FileProvider` does open the door to sharing files to third-party applications, which might be undesirable for some applications, especially if it's allowing `dataDirectory` files to be shared (which is normally a private storage container). But I'm not stating this with any kind of certainty...

I think to **actually** share/expose a file to another app, the application must use [Context.grantUriPermission](https://developer.android.com/reference/android/content/Context.html#grantUriPermission(java.lang.String,%20android.net.Uri,%20int)) on the `content://` url. So just by having a FileProvider available might not be enough to share files, I think this needs to be confirmed.

KenCorbettJr: Who might know the security implications of having a FileProvider like this?

framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
```diff
@@ -212,53 +217,110 @@ public View getVideoLoadingProgressView() {
...
-            intent.putExtra(Intent.EXTRA_MIME_TYPES, acceptTypes);
+            fileIntent.setType("*/*"); // Accept all, filter mime types by Intent.EXTRA_MIME_TYPES.
+            fileIntent.putExtra(Intent.EXTRA_MIME_TYPES, acceptTypes);
+        }
+
+        // Image from camera intent
+        Uri tempUri = null;
+        Intent captureIntent = null;
+        if (fileChooserParams.isCaptureEnabled()) {
+            captureIntent = new Intent(MediaStore.ACTION_IMAGE_CAPTURE);
```
erisu, line 242: Please confirm:

I believe you need to add this to the [AndroidManifest queries](https://developer.android.com/guide/topics/manifest/queries-element). 

```xml
<queries>
    <intent>
        <action android:name="android.media.action.IMAGE_CAPTURE" />
    </intent>
</queries>
 ```
 
Ref: https://developer.android.com/training/package-visibility

Because the Intent [MediaStore.ACTION_IMAGE_CAPTURE](https://developer.android.com/reference/android/provider/MediaStore#ACTION_IMAGE_CAPTURE) is for accessing the system's camera application for capturing an image and returning it, the app might not have visibility to the package.

> When an app targets Android 11 (API level 30) or higher and queries for information about the other apps that are installed on a device, the system filters this information by default. This filtering behavior means that your app can’t detect all the apps installed on a device, which helps minimize the potentially sensitive information that your app can access but doesn't need to fulfill its use cases.

Here is also an SO referencing similar feedback: https://stackoverflow.com/questions/63246442/android-11-r-return-empty-list-when-querying-intent-for-action-image-capture

**Review by erisu — CHANGES_REQUESTED** [review]

File: templates/project/res/xml/opener_paths.xml (hunk 1 above)
Comment (breautek) [reply templates/project/res/xml/opener_paths.xml:3]: On the security front, I think we are in the clear.

Having the file provider and declaring which files/folders are shareable/visible isn't enough to share files. It is part of the process to setup fire sharing, but the app needs to do additional steps to actually share files.

Most of the time this is done via setting up a specialized [Activity](https://developer.android.com/training/secure-file-sharing/share-file), where a third-party app would create an Intent, which may target your app in which your app will start an activity to select a file and provide a result. The `Context.grantUriPermission` can achieve the same result, but permisison will be granted permanently, or until the app explicitly revokes the permission, whereas the Intent way is a temporary permission grant.

Cordova apps only have a single activity, generally speaking, which is the main activity for the application. So I believe the only way this can be an issue is if someone decides to implement an activity that responds to intents and uses this file provider being introduced in this PR, which is an unlikely scenario.

We could write a note for future app developers/plugin authors in `templates/project/res/xml/opener_paths.xml` that describing that the file provider may only be used for the application itself and should **not** be used for responding to Intents because it will expose the app's private data folders.

File: templates/project/AndroidManifest.xml [hunk 5]
```diff
@@ -46,5 +46,8 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        <provider android:name="androidx.core.content.FileProvider" android:authorities="${applicationId}.provider" android:exported="false" android:grantUriPermissions="true">
```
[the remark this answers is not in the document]
KenCorbettJr: This change has been applied.

templates/project/AndroidManifest.xml
```diff
@@ -46,5 +46,8 @@
...
+            <meta-data android:name="android.support.FILE_PROVIDER_PATHS" android:resource="@xml/opener_paths" />
```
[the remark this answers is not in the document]
KenCorbettJr: I like it, changed.

templates/project/res/xml/opener_paths.xml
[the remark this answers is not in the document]
KenCorbettJr: I have adopted your suggestion.  Updated.

templates/project/res/xml/opener_paths.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- https://developer.android.com/reference/android/support/v4/content/FileProvider.html#SpecifyFiles -->
```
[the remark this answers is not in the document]
KenCorbettJr: Fixed.  Thank you.

templates/project/res/xml/opener_paths.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8"?>
```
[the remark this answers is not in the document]
KenCorbettJr: Added.

templates/project/res/xml/opener_paths.xml
KenCorbettJr: Great idea.  I added a comment to the `cdv_core_file_provider_paths.xml` file that reads:

    Note: This File provider may only be used by the Cordova core itself
    and should not be used for responding to Intents because it will
    expose the app's private data folders.

Does that sound OK?

framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
KenCorbettJr: I added this to the AndroidManifest as you suggested.

framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
dpogue: Let's leave it for now

templates/project/res/xml/cdv_core_file_provider_paths.xml
```diff
@@ -0,0 +1,39 @@
...
+    <!-- cordova.file.dataDirectory -->
+    <files-path name="files" path="." />
+    <!-- cordova.file.cacheDirectory -->
+    <cache-path name="cache" path="." />
+    <!-- cordova.file.externalDataDirectory -->
+    <external-files-path name="external-files" path="." />
+    <!-- cordova.file.externalCacheDirectory -->
+    <external-cache-path name="external-cache" path="." />
+    <!-- cordova.file.externalRootDirectory -->
+    <external-path name="external" path="." />
```
dpogue, line 38: Are we using all of these, or do we only need access to the cache-path?

Ideally, we only set up providers for the access we actually need, to reduce the potential for security concerns, but it's not entirely clear to me from the code how exactly these are being used. Maybe we do need all of them?

KenCorbettJr: I tested, and we only need `cache-path`.  Reading through https://developer.android.com/reference/androidx/core/content/FileProvider this makes sense.  We are only calling `context.getCacheDir()` so that should be the only one that we really need.

dpogue (approved):
Thanks! This looks good to me now. I'll see about trying to test it on a device sometime this week :)

KenCorbettJr:
What additional testing needs to be done to get this merged in?

erisu (approved):
LGTM

erisu:
```html
<!-- // Display: File browser
// restrictions: no file type
// file limit: 1 -->
<h2>1. File</h2>
<input type="file" />

<!-- // Display: File browser
// restrictions: must be an image file type
// file limit: 1 -->
<h2>2. File - accept image</h2>
<input type="file" accept="image/*" />

<!-- // Display: Chooser (File or Camera)
// restrictions: no file type
// file limit: 1 -->
<h2>3. File - cature</h2>
<input type="file" capture />

<!-- // Display: Chooser (File or Camera)
// restrictions: must be an image file type
// file limit: 1 -->
<h2>4. File - accept image and capture</h2>
<input type="file" accept="image/*" capture />

<!-- // Display: File browser
// restrictions: no file type
// file limit: many -->
<h2>5. File - multiple</h2>
<input type="file" multiple />

<!-- // Display: File browser
// restrictions: must be an image file type
// file limit: many -->
<h2>6. File - accept image and multiple</h2>
<input type="file" accept="image/*" multiple />

<!-- // Display: Chooser (File or Camera)
// restrictions: no file type
// file limit: many -->
<h2>7. File - capture & multiple</h2>
<input type="file" capture multiple />

<!-- // Display: Chooser (File or Camera)
// restrictions: must be an image file type
// file limit: many -->
<h2>8. File - accept image, capture & multiple</h2>
<input type="file" accept="image/*" capture multiple  />

<!-- // BAD CASES
// If accept type is restrictive to a non-image format and capture flag was provided.
// Display: Chooser (File or Camera)
// restrictions: must be a PDF
// file limit: 1 -->
<h2>9. File - accept pdf and capture</h2>
<input type="file" accept=".pdf" capture />
```

* File browser appears
  * I can see all file types
* File browser appears and I see only images
* Chooser intent appears
  * I can select Camera or File
  * I see all file type
* Chooser intent appears
  * I can select Camera or File
  * I see only image files
* Multiple flag does not work.
  * I see all file type
* Multiple flag does not work.
  * I see image only type
* Chooser intent appears.
  * I can select Camera or File 
  * I see all file type
  * Multiple flag doesn't seem to work
* Chooser intent appears.
  * I can select Camera or File
  * I see  only image files
  * Multiple flag doesn't seem to work
* Chooser intent appears
  * I can select Camera or File
  
The list above is in the order of the test sampe.
These were run on Android 13 device.
I created a blank Cordova default app with no additional plugins.

End results:
* The chooser intent displays as expected.
* The camera option does not seem to return the image. The file selector continues to say "No file chosen"
* The multiple file selector flag does not seem to work. Only can select one file.

This PR will be pushed back for a later release, until the above issues are resolved.

ArturKp:
Hi. We've had a related issue in our backlog for quite some time and last time when the topic came up again in our team, I was quite happy to find this PR which was made a few days earlier. Now it's been without an update for quite some time.

Are you @KenCorbettJr still motivated and planning to finish it? Any updates?

KenCorbettJr:
@ArturKp I would still love to get this done.  I was busy for a bit, but I have some more time now and I think we are close to getting this one over the line.

I created a test app that pointed at my fork of cordova android to test all the permutations (git repo: https://github.com/KenCorbettJr/cordova-android-file-tester) and at least on my Samsung S22 everything seems to work great.  Every time there was a `capture` attribute on the file input I was shown the option to take a picture using my camera.  The `multiple` and `accept` attributes seemed to work great too.

@erisu is there any way you can take another look with my test app?  I am pretty sure I set it up correctly.

erisu:
I will try and re-test this PR tomorrow.

christiaan:
> 
> End results:
> 
> * The chooser intent displays as expected.
> * The camera option does not seem to return the image. The file selector continues to say "No file chosen"
> * The multiple file selector flag does not seem to work. Only can select one file.
> 
> This PR will be pushed back for a later release, until the above issues are resolved.

@erisu Did you try long-pressing the files to select multiple? Because as soon as I clicked one it selected it, but long-pressing a file allowed me to select multiple.

I've tested https://github.com/KenCorbettJr/cordova-android-file-tester on Android 13 (Samsung Galaxy A22), 10 (Moto g8+) and 8 (Samsung Galaxy A3 2017) devices and was not able to reproduce any of the issues.

erisu:
@christiaan I do not recall if I long-pressed when testing.

But, as of right now, I am stuck on how to proceed with this PR.

I stumbled across this StackOverflow, which seems to have the same code, almost word for word. I am guessing the slight differences are due to previous feedback/suggestions...

https://stackoverflow.com/questions/57586259/cordova-android-clicking-input-type-file-brings-up-system-file-browser-not-came

What I am stuck on is this Apache rule: https://www.apache.org/legal/resolved.html#stackoverflow

> Can I Copy Code From Stack Overflow And Contribute It To An ASF Project?
> 
> No, not without contacting the original author and getting permission from them to use the code in an Apache project under the Apache License 2.0.

Even though on StackOverflow's Terms of Services, under "Content Permissions, Restrictions, and Creative Commons Licensing" for "Subscriber Content", it says:

> You should be aware that all Public Content you contribute is available for public copy and redistribution, and all such Public Content must have appropriate attribution.

It still goes against Apache's statement.

I might have to reach out to Apache Legal and confirm if their statement still stands, knowing what StackOverflow's Terms of Services says..

I just noticed the SO post and started dwelling on this...

christiaan:
@NoodleOfDeath you are the author of the Stack Overflow answer which appears to be one on one copied to this PR.
Would you mind if it is incorporated in cordova android?

NoodleOfDeath:
By all means!

christiaan:
@erisu Is this enough permission? :smile:

breautek (approved):
- resource files are prefixed with `cdv` :+1: 
- authority has `cdv` prefix in the namespace :+1: 
- PR using `createTempFile` with the app's internal cache directory, which avoids permission issues :+1: 
- Uses MediaStore and Intents to delegate to the camera, avoiding requiring special permissions :+1: 

I haven't tested the code but codewise it looks good.

And yes, I believe NoodleOfDeath's written approval here is sufficient. Thank you.

templates/project/AndroidManifest.xml
```diff
@@ -46,5 +46,14 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        <provider android:name="androidx.core.content.FileProvider" android:authorities="${applicationId}.cdv.core.file.provider" android:exported="false" android:grantUriPermissions="true">
```
NoodleOfDeath, line 49: lgtm

NoodleOfDeath (approved):

christiaan:
@breautek @erisu not to sound impatient but, what is holding this back from being merged?

breautek:
> @breautek @erisu not to sound impatient but, what is holding this back from being merged?

It's in the 13.x milestone, I'm not really sure why it was placed there. I don't think there is anything actually breaking,  @erisu do you remember why this was placed in in the 13.x milestone?

erisu:
There was no specific reason; it simply didn't have time to be included in the 12.x milestone, which was designated for the major release. Therefore, I chose another milestone randomly.

As it's a new feature, it can be included into the next minor release, 12.1.x.

erisu (approved):

Scooby27:
Hi all, is there any timeline this will be released? Thanks!

lovelyelfpop:
When will this be merged?

jacobg:
What about requesting for camera permission?

breautek:
> What about requesting for camera permission?

If you mean [android.Manifest.permission.CAMERA](https://developer.android.com/reference/android/Manifest.permission#CAMERA) this permission is only required for accessing the camera APIs, not for using the intents.

jacobg:
> > What about requesting for camera permission?
> 
> If you mean [android.Manifest.permission.CAMERA](https://developer.android.com/reference/android/Manifest.permission#CAMERA) this permission is only required for accessing the camera APIs, not for using the intents.

Using the onShowFileChooser implementation in this pull request, and choosing Camera, the Camera screen does not appear. But if I request CAMERA permission first and user grants it, then the Camera will show.

jacobg:
It seems that while it works on a Samsung Galaxy A50 running Android 11, it neither works on any Android simulator SDK version (30-34), nor does it work on Pixel 8 running Android 14. "Doesn't work" means that after taking a photo and ok'ing it, the input change event does not fire.

jacobg:
The `intent == null` if-condition should be removed. It should just be:
```java
if (captureUri != null) { // camera
   LOG.v(LOG_TAG, "Adding camera capture: " + captureUri);
   uris.add(captureUri);
}
```

KenCorbettJr:
@jacobg how do they need to be revised?

jacobg:
Hi @KenCorbettJr  Thanks for your pull request. Do you see a "review" that was opened and displayed above your last message.

I'll paste it again here:
--


The intent == null clause should be removed, and getData() condition moved to the top with non-null intent clause. The camera result may have an intent but without data.

So like this:
```java
if (intent != null && intent.getData() != null) { // single file
    LOG.v(LOG_TAG, "Adding file (single): " + intent.getData());
    uris.add(intent.getData());
} else if (captureUri != null) { // camera
    LOG.v(LOG_TAG, "Adding camera capture: " + captureUri);
    uris.add(captureUri);
} else if (intent != null && intent.getClipData() != null) { // multiple files
    ClipData clipData = intent.getClipData();
    int count = clipData.getItemCount();
    for (int i = 0; i < count; i++) {
        Uri uri = clipData.getItemAt(i).getUri();
        LOG.v(LOG_TAG, "Adding file (multiple): " + uri);
        if (uri != null) {
            uris.add(uri);
        }
    }
}
```

And camera would not open for me without first requesting CAMERA permission.

KenCorbettJr:
I applied your code review feedback.

Should I also add a request for camera permission?  Or somehow check if the app already has that permission?

jacobg:
Thanks!

I was able to get it working in my app but requesting permission in javascript code first before opening chooser:
```js
    getFileEl () {
      return document.getElementById(`file-${this.$.uid}`)
    },
    async requestCameraPermission () {
      const diagnostic = window.cordova?.plugins?.diagnostic
      if (!diagnostic) return true
      const releasePermissionLock = await this.rootStore.acquireLockToRequestPermissions()
      return new Promise((resolve, reject) => {
        diagnostic.requestCameraAuthorization(
          status => resolve(status === diagnostic.permissionStatus.GRANTED),
          error => reject(error),
          false
        )
      })
        .finally(async () => {
          releasePermissionLock()
        })
    },
    async addDocument () {
      if (this.saving) return

      // The permissions and file picker will pause the webview, so we need to prevent that.
      this.idleStore.setPreventPause(true)

      // First we need to ask for camera permission, even though we don't know
      // whether they'll choose camera or not.
      // TODO: It seems Cordova needs to do this for us:
      // https://github.com/apache/cordova-android/pull/1609#issuecomment-1905156891
      const granted = await this.requestCameraPermission()
      if (!granted) {
        console.warn('User did not grant camera permission before adding document')
      }

      const fileEl = this.getFileEl()
      if (!fileEl) {
        console.warn('File input element not mounted after requesting camera permission')
        return
      }

      // Open file picker.
      fileEl.click()
    },
```

It would be nice if the native code could intercept the user choosing Camera, and only then ask for permissions, rather than ask for permissions before showing the chooser even though user may choose from gallery. But I'm not sure why @breautek thinks intent doesn't need to grant permission? Maybe there's a different approach?

breautek:
> But I'm not sure why @breautek thinks intent doesn't need to grant permission? Maybe there's a different approach?

I know there was behaviour which was handled in the camera plugin where if the `CAMERA` permission is declared in the manifest, then the `CAMERA` permission needs to be granted at the app-level. (Which somewhat defeats the purpose of using Intents...) Perhaps you have the `CAMERA` permission declared in your android manifest somehow, via another plugin or something. The camera plugin handles this by checking if the camera permission is present in the manifest and only request the camera permission if it is.

Google has announced that starting in August 2024, that they are going to start cracking down on apps declaring or requesting permissions that the app doesn't actually use or need, so we need to pay close attention when it comes to adding new permission requirements.

@jacobg If you could, open up the app that you're testing with inside Android Studio and view your `AndroidManifest.xml` file. The Android Studio editor for the manifest file will have a tab for `Merged Manifest` view, which should represent your manifest as natively compiled. If you can see that the `CAMERA` permission is declared, it should tell what library/module it is coming from. Likely it will be the "app" module since a plugin that you have installed is probably adding it. If you could temporarily remove the CAMERA declaration in the manifest and re-test, I think you should find that it will work as expected without the camera permission. You would also of course have to temporarily disable any code that attempts to request the camera permission as well.

jacobg:
Ah, so that explains it. My app does declare CAMERA in the manifest because it has a separate use-case where it needs the camera. If the user has not granted CAMERA permission yet for that other use-case, then the file input won't open until granting CAMERA permission there.

So I guess there's two options here:
1) App request permission in javascript if it needs it.
2) Cordova can check if CAMERA is in the manifest, and request permission.

martjag12:
Can this be included in next minor releases? Thank you!

HazzMan2409:
Am I missing something or do we still not have parity with the web here because we cannot get video from an input element?

KenCorbettJr:
@HazzMan2409  Yes, you are right.  This pull request only added camera not video.

PR #1678, opened 2023-10-21
feat!: API 34 Support

breautek:
### Platforms affected

Android

### Motivation and Context

Adds in support for the next android platform

Closes https://github.com/apache/cordova-android/issues/1661
Closes https://github.com/apache/cordova-android/pull/1596

### Description

This PR supercedes and includes https://github.com/apache/cordova-android/pull/1596

This PR does a few different things to provide API 34 support which will be noted below.

#### Breaking Changes

These are notable changes that is breaking that may affect plugins or workflows. It doesn't highlight *all* breaking changes in which were hidden inside implementation details.

- JDK 17 is now required
- Cordova-Android will now use Gradle 8.7 by default
- Android Studio Jellyfish. Android Studio is not required for building via the Cordova CLI.

#### Default Configuration Changes

Target SDK: 34
Gradle Version: 8.7
AGP Version: 8.3.0
Kotlin JVM Target: 1.8 (If kotlin is enabled)
Min Build Tools: 34

#### Java/Kotlin Changes

~Source/Target compatibility we changed from Java 8 to 17 on both the framework and app template projects.~

Source & Target compatibility have been reverted back to 8, but 2 new preferences are added to allow developers to experiment: `AndroidJavaSourceCompatibility` and `AndroidJavaTargetCompatibility` which adjusts the respective gradle compile options. By default, these values are 8, which is the same values used before.

Additionally, `AndroidKotlinJVMTarget` preference was added to select the Kotlin JVM target. By default, the Kotlin JVM Target will be same as `AndroidJavaTargetCompatibility`. Updating AndroidJavaTargetCompatibility will influence the Kotlin JVM target unless if `AndroidKotlinJVMTarget` is also explicitly defined.

#### BuildConfig

Previously this I believe was implicitly enabled, but now needs to be explictly enabled otherwise the following may occur

```
* What went wrong:
A problem occurred configuring project ':app'.
> com.android.builder.errors.EvalIssueException: defaultConfig contains custom BuildConfig fields, but the feature is disabled.
  To enable the feature, add the following to your module-level build.gradle:
  `android.buildFeatures.buildConfig true`
```

I however did not encounter this on a simple project, so it might be required depending on what the plugins does.

#### AGP 8.3

AGP 8.2 is the minimum version required for API 34 builds. Attempting build against API 34 on earlier versions will produce warnings and/or errors.

AGP 8.3 is however the latest version that can be used with standard Android Studio configurations.

AGP 8.4 was tested with and will mostly work, but will not work inside Android Studio without further configuration. Because we don't want to impose Android Studio configurations onto the user, AGP 8.3 is chosen for cordova-android@13 release.

#### Gradle `-b` flag

This flag has been deprecated and scheduled to be removed in gradle's next major release.
The gradle projects must follow a standard structure now. Cordova's generated project already follows the standard structure thus for `cordova build` commands the `-b` append can simply just be removed, we don't need to explicitly declare the `build.gradle` file.

It was also used against the `wrapper.gradle` file which changes are noted below.

#### gradle.wrapper

The "intentionally empty" gradle.wrapper file has been removed.

We no longer manually manage gradle versions. Instead, we let gradle do it for us by utilising their `--gradle-version` argument. A new function was introduced to replace the old `runGradleWrapper` to:

1. setup the gradle wrapper if not already present
2. install/update the gradle wrapper to the specified version.

By default, this will be `8.7` as defined in the defaults json config file, but may be overwritten by `GradleVersion` preference, or `CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL` environment variable.

If `CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL` is defined, the preference will be ignored and `--distribution-url` will be used instead, maintaining the ability for corporations to install their own gradle wrapper from their own source.

The Gradle Wrapper was previously checked into the repo and was removed. It has since returned as using the system gradle to generate the wrapper caused issues. Namely it would require end-users to have an appropriate gradle version that satisfies AGP, which kind of defeats the purpose of using wrappers.

Because `-b` flag will be going away, there is now a `tools` project that is templated and copies. This project serves as the old `wrapper.gradle`, in that it's an empty project that can be used by the system gradle of a varying version range to produce the gradle wrapper at the version that we need to build the android project.

<details>
<summary>~obsolete gradle-wrapper notes~</summary>

**This section is no longer in the PR**

While personally I do not agree with how Gradle handles this, it is their [recommendation](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:adding_wrapper) to check in the Gradle wrapper, including the jar file into the source.

> To make the Wrapper files available to other developers and execution environments, you need to check them into version control. Wrapper files, including the JAR file, are small. Adding the JAR file to version control is expected. Some organizations do not allow projects to submit binary files to version control, and there is no workaround available.

> As usual, you should commit the changes to the Wrapper files to version control.

This done for both our test project and the app template. On platform add, the wrapper will be copied from the template. Individual projects via `GradleVersion` may change the wrapper version.

Should the end user choose a gradle version below what was required by AGP, gradle will update the gradle version, but builds will fail. Attempting to update the version back to an acceptable version will cause the gradle wrapper task to fail as the wrapper task also loads in AGP. To avoid this, Cordova does a version check before executing the wrapper task and will not update the wrapper if it's below the minimum required.

Note that while by preference both Gradle Version and AGP versions are configurable, the version check is not. This means that even if you drop Gradle and AGP down a version, the current >= 8.6 check will remain. I think this is OK since these preferences are documented for internal development use.

Lastly, because of the introduction of using `--gradle-version` and letting gradle manage our wrappers, the clean gradle wrapper script we had for our test project became obsolete and was removed.
</details>

#### Android Studio

With AGP 8.3, [Android Studio Jellyfish](https://developer.android.com/build/releases/gradle-plugin) will now be required for opening the native project.

#### CI

JDK was updated from 11 to 17, which is a requirement to build for API 34

#### Testing

Ran NPM test
Ran in sample cordova hello world app, with variety of different configurations
Ran in one of my projects

#### Known Issues

- The core splashscreen plugin does not compile when targeting JDK 17.

### Checklist


framework/build.gradle
```diff
@@ -50,8 +50,8 @@ android {
     buildToolsVersion cordovaConfig.BUILD_TOOLS_VERSION
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_17
```
EiyuuZack, line 53: Hey guys, quick question I am struggling to find answers for:
Considering the `minSdk` currently in effect (API Level 24), how can you be sure there won't be any issues targeting Java 17?

As of today, generating a pure Android project using AS Hedgehog RC and AGP 8.2.0-rc03 still results in a Gradle file containing:
```
    compileOptions {
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
    }
    kotlinOptions {
        jvmTarget = '1.8'
    }
```
It seems Google is not ready to pull the trigger on Java 17 yet, so I am wondering if Cordova should?

breautek: JDK 17 compatibility was [added in AGP 8.1](https://developer.android.com/build/releases/gradle-plugin#8-1-0)

> how can you be sure there won't be any issues targeting Java 17?

It's my understanding that it will enable JDK 17 features on supported devices. Different devices will have different java runtimes,so plugin authors may still need to do an API check. But targeting JDK 17 doesn't necessarily limit the ability to run on older devices.

Resources:
https://developer.android.com/build/jdks#compileSdk
https://developer.android.com/about/versions/14/features#core-libraries

Though you do bring a good point, it's hard to say with any level of certainty that this won't break numerous of plugins that are likely built with JDK 8 as a target. If there are any breaking changes in the language, the compilation target may break builds. It hasn't in my own testing with the plugins I have used, it might be smart to introduce a preference to set these fields.

EiyuuZack: > JDK 17 compatibility was [added in AGP 8.1](https://developer.android.com/build/releases/gradle-plugin#8-1-0)

That is a requirement to run AGP itself and does not seem to have anything to do with how the project is setup with regards to JVM targeting.

> Resources:
https://developer.android.com/build/jdks#compileSdk
https://developer.android.com/about/versions/14/features#core-libraries

From my understanding, the documentation refers to applications built with JDK 17 as a target + running on Android 14.

For 13 and below it would require desugaring of JDK 17 APIs via the `com.android.tools:desugar_jdk_libs:2.0.3` dependency as per https://developer.android.com/studio/write/java8-support#library-desugaring.

It is hard to tell because that documentation seems stopped in time (unlike the dependency itself which was updated [earlier this month](https://mvnrepository.com/artifact/com.android.tools/desugar_jdk_libs/2.0.4)), with no mention to AGP 8 or JDK 17 for that matter. The [changelog](https://github.com/google/desugar_jdk_libs/blob/master/CHANGELOG.md) of the dependency does not mention JDK 17 either.

The way I see it if you want to perform this change you would need to add the dependency and set the `isCoreLibraryDesugaringEnabled` property to `true`, assuming that it is indeed capable to desugar JDK 17 APIs.

What are your thoughts on this?

breautek: I've revert the change back to 8, but powered the source and target compatibility options via 2 new preferences so that it can be experimented with. The default will remain using Java 8.

Overall I do agree that we should try to maintain the Android Studio default as much as possible.

breautek: Added `AndroidKotlinJVMTarget`, which will only be effective if kotlin is enabled (by default its disabled but there is an existing preference to enable it). The default value is for the JVM target `1.8`.

Unlike the java source / target compatibility, `AndroidKotlinJVMTarget` only applies to the app target, not the cordova framework. Cordova framework currently doesn't setup/enable kotlin nor does it use kotlin, so it doesn't read or set the JVM target. Trying to do so will cause build errors.

lib/builders/ProjectBuilder.js
```diff
@@ -116,16 +114,17 @@ class ProjectBuilder {
...
+     * @param {string} gradleVersion The gradle version to install. Ignored if CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL environment variable is defined
+     * @returns {Promise<void>}
+     */
+    async installGradleWrapper (gradleVersion) {
+        if (process.env.CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL) {
+            events.emit('verbose', `Overriding Gradle Version via CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL (${process.env.CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL})`);
+            await execa('gradle', ['-p', this.root, 'wrapper', '--gradle-distribution-url', process.env.CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL], { stdio: 'inherit' });
         } else {
-            return execa(gradle_cmd, ['-p', this.root, 'wrapper', '-b', wrapperGradle], { stdio: 'inherit' });
+            await execa('gradle', ['-p', this.root, 'wrapper', '--gradle-version', gradleVersion], { stdio: 'inherit' });
```
bryan-hunter, line 127: was upgrading some stuff myself and used this as reference.

I was getting this error:

```
FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':app:wrapper' (type 'Wrapper').
  - In plugin 'org.gradle.buildinit.plugins.BuildInitPlugin$Inject' type 'org.gradle.api.tasks.wrapper.Wrapper' property 'validateDistributionUrl' doesn't have a configured value.
    
    Reason: This property isn't marked as optional and no value has been configured.
    
    Possible solutions:
      1. Assign a value to 'validateDistributionUrl'.
      2. Mark property 'validateDistributionUrl' as optional.
    
    For more information, please refer to https://docs.gradle.org/8.7/userguide/validation_problems.html#value_not_set in the Gradle documentation.
    ```
    
    I added `'--validate-url'` to both of these gradle wrapper commands and the error went away.  There may be a better way, but figured I'd help if it comes up for anyone else

File: framework/cdv-gradle-config-defaults.json [hunk 3]
```diff
@@ -1,16 +1,19 @@
...
     "ANDROIDX_WEBKIT_VERSION": "1.6.0",
     "ANDROIDX_CORE_SPLASHSCREEN_VERSION": "1.0.0",
     "GRADLE_PLUGIN_GOOGLE_SERVICES_VERSION": "4.3.15",
     "IS_GRADLE_PLUGIN_GOOGLE_SERVICES_ENABLED": false,
     "IS_GRADLE_PLUGIN_KOTLIN_ENABLED": false,
-    "PACKAGE_NAMESPACE": "io.cordova.helloCordova"
+    "PACKAGE_NAMESPACE": "io.cordova.helloCordova",
+    "JAVA_SOURCE_COMPATIBILITY": 8,
+    "JAVA_TARGET_COMPATIBILITY": 8,
+    "KOTLIN_JVM_TARGET": "1.8"
```
erisu, line 18: Is there a conflict with this and `KOTLIN_VERSION`?

lib/prepare.js
```diff
@@ -110,7 +110,10 @@ function getUserGradleConfig (configXml) {
         { xmlKey: 'AndroidXWebKitVersion', gradleKey: 'ANDROIDX_WEBKIT_VERSION', type: String },
         { xmlKey: 'GradlePluginGoogleServicesVersion', gradleKey: 'GRADLE_PLUGIN_GOOGLE_SERVICES_VERSION', type: String },
         { xmlKey: 'GradlePluginGoogleServicesEnabled', gradleKey: 'IS_GRADLE_PLUGIN_GOOGLE_SERVICES_ENABLED', type: Boolean },
-        { xmlKey: 'GradlePluginKotlinEnabled', gradleKey: 'IS_GRADLE_PLUGIN_KOTLIN_ENABLED', type: Boolean }
+        { xmlKey: 'GradlePluginKotlinEnabled', gradleKey: 'IS_GRADLE_PLUGIN_KOTLIN_ENABLED', type: Boolean },
+        { xmlKey: 'AndroidJavaSourceCompatibility', gradleKey: 'JAVA_SOURCE_COMPATIBILITY', type: Number },
+        { xmlKey: 'AndroidJavaTargetCompatibility', gradleKey: 'JAVA_TARGET_COMPATIBILITY', type: Number },
+        { xmlKey: 'AndroidKotlinJVMTarget', gradleKey: 'KOTLIN_JVM_TARGET', type: String }
```
erisu, line 116: Where is this used? Some reason I can't find `AndroidKotlinJVMTarget` used anywhere...

breautek: The `AndroidKotlinJVMTarget` itself isn't used in the NodeJS side of preparation.

lib/builders/ProjectBuilder.js
breautek: I assume `--validate-url` only makes sense if `--gradle-distribution-url` is used.

Which would only be the cause if there is a system environment variable set for `CORDOVA_ANDROID_GRADLE_DISTRIBUTION_URL`.

breautek: Scratch that..

After updating gradle to the latest version I reproduced this problem and it is also erroring on the wrapper --gradle-version command.

erisu (approved):
LGTM

erisu (changes requested):
Having an issue with `npm t` not able to download and use gradle wrapper.

breautek:
> Having an issue with `npm t` not able to download and use gradle wrapper.

I think I know what you're running into. It's dumb but gradle cannot install the gradle wrapper if the project does a gradle version check (which the android gradle plugin does). In otherwords if AGP requires Gradle >= 8.7.0 it will prevent a lesser gradle version from using the gradle wrapper task to install a 8.7.0 wrapper.

Using gradle's `--gradle-version` is hella lot easier than what we did before, which involved deleting the wrapper because even if you switched to the new version it won't use it if it already had a wrapper installed. The wrapper task manages all of that for us and is what Gradle [recommends](https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper).

They also recommend to check in the wrapper, which is also dumb imo (breaks the general rule of no "built" artefacts in the repo, cause it causes headaches during PR and merge times).

So we either:

1. check in gradle wrapper into the repo, and exclusively use it always (e.g. remove all references of the `gradle` executable, which is only ever used to manage the wrapper).

2. Require users to require the appropriate system gradle version.

Option 1 is following Gradle's recommendation but I think can run into issues that can be problematic. For example, what happens if the end user uses the preference to install a gradle wrapper less than the minimum required by AGP? Will it run and switch for it to fail later once it starts using the unsupported wrapper version? Once the wrapper is out of date, the only way to fix it is to use the appropriate system gradle at the proper version. Option 1 seems like a better approach and follows Gradle recommendations has some concerns.

Option 2 seems like a much safer approach, just more annoying for end users and defeats the purpose of using the wrapper really.

dpogue:
IIRC we're not allowed to commit the gradle jar file, this was an issue that came up when we were originally moving from ant to gradle and there were ASF policies that prevented redistribution of the jar.

breautek:
> IIRC we're not allowed to commit the gradle jar file, this was an issue that came up when we were originally moving from ant to gradle and there were ASF policies that prevented redistribution of the jar.

Do you recall where Apache states so?

https://infra.apache.org/licensing-howto.html#permissive-deps makes it sound like it's permissible for bundling dependencies, for as long as the license permits redistributions. And there is clarification on how to handle binaries [here](https://infra.apache.org/licensing-howto.html#binary)

dpogue:
It's not regarding licensing per se, it's under the source packages policy:

> ### Source packages
>
> Every ASF release MUST contain one or more source packages, which MUST be sufficient for a user to build and test the release provided they have access to the appropriate platform and tools. A source release SHOULD not contain compiled code.

That last line in particular.

With that said, I looked and see several other ASF projects with gradle wrappers committed to their repositories, so I don't think I have any actual objection to this.

breautek:
> It's not regarding licensing per se, it's under the [source packages policy](https://www.apache.org/legal/release-policy.html#source-packages):
> 
> > ### Source packages
> > Every ASF release MUST contain one or more source packages, which MUST be sufficient for a user to build and test the release provided they have access to the appropriate platform and tools. A source release SHOULD not contain compiled code.
> 
> That last line in particular.
> 
> With that said, I looked and see several other ASF projects with gradle wrappers committed to their repositories, so I don't think I have any actual objection to this.

I believe another section allows it given certain requirements:

https://www.apache.org/legal/release-policy.html#distribute-other-artifacts

>ASF releases typically contain additional material together with the source package. This material may include documentation concerning the release but must contain LICENSE and NOTICE files. As mentioned above, these artifacts must be signed by a committer with a detached signature if they are to be placed in the project's distribution directory.
>
>Again, these artifacts may be distributed only if they contain LICENSE and NOTICE files. For example, the Java artifact format is based on a compressed directory structure and those projects wishing to distribute jars must place LICENSE and NOTICE files in the META-INF directory within the jar.

I unpacked the wrapper jar file and found that they have their own license file, but does lack a NOTICE file.

At the end of the day, the contents is still going to be signed, comes from a trusted source, assuming it gets updated via gradle tooling and the jar is a build tool dependency, it isn't part of the apache source code. But I understand how that could tie into the "provided they have access to the appropriate platform and tools." In legal terms they also used `SHOULD` instead of `MUST` or `SHALL`. "Should" is more of a suggestion rather than mandatory whereas "must" or "shall" is mandatory terms.

Therefore including compiled code should be a thoughtful process (like this one). It shouldn't be made a habit, it should only be done when it's truly necessary. I think the gradle wrapper fits all of those ticks.

jcesarmobile:
Latest Android Studio updates gradle wrapper to 8.6, should we use the same version Android Studio updates to instead of latest?

breautek:
> Latest Android Studio updates gradle wrapper to 8.6, should we use the same version Android Studio updates to instead of latest?

Unless if there is a specific issue with 8.7 I see no reason to use it by default as the current stable release. Android Studio might default to 8.6 perhaps as the stable gradle at the time of their release.

[8.7](https://docs.gradle.org/current/release-notes.html) includes a ton of bug fixes and improves including security patches that affects 8.6.

breautek:
FYI, was talking with Erisu last night and we came up with a plan to remove the JAR and thus avoid the security concerns about the integrity (e.g. xz backdoor style attacks) of the JAR file by introducing an empty gradle project, which effectively will serve as the old `wrapper.gradle` that we had.

Running system gradle against the actual project will result in loading up the AGP plugin and thus requiring the end-user to have a system gradle version that satisfies AGP gradle requirements. Previously with the (now deprecated/pending removal of `-b` flag) wrapper.gradle file, it was used to install the wrapper without loading the AGP plugin, virtually allowing the user to be able to use any gradle version to install the wrapper.

With the removal of the `-b` flag gradle is enforcing a consistent project structure. So the plan is to introduce a secondary gradle project that is just empty and serves the purpose of installing gradle wrapper, which should be usable by the cordova build. If time permits, I'm planning on making these changes later today.

lib/builders/ProjectBuilder.js
breautek: I've reverted the `--validate-url`. I'm not confident it actually fixed anything.

[Gradle Docs](https://docs.gradle.org/8.2/userguide/gradle_wrapper.html) states that `--validate-url` enables validation on the distribution url and that it's enabled by default. So it doesn't much sense that adding `--validate-url` would have changed anything, and explicitly adding it will require a minimum system gradle of 8.2 or later since that's when the flag was added.

I've also haven't reproduced the issue since the first time. Even after trying to force the issue by messing with `validateDistributionUrl` in the gradle properties file.

breautek:
Some last few changes which I think improves the PR significantly.

Like in the original post I really don't like submitting generated content. Utilising an empty gradle project we are able to now generate the wrapper on the fly rather than checking it in and redistributing it.

Using an empty project allows us to avoid AGP's version requirements when running the system gradle. Allowing us to use virtually any gradle version to generate the 8.6+ gradle wrapper.

The gradle wrapper and it's jar has been removed.

erisu (changes requested):
A missing small revert change.

Also the `.ratignore` changes could be reverted, but if reverted, need to add at least `gradle-wrapper.properties`. This this file is detected if anyone runs locally a rat test after running `npm test`.

Also needs to rebase with the main branch and fix the conflict in `ci.yml`.

Other then the above, everything LGTM.

I ran an `npm t` and confirmed that Gradle issues were resolved.

erisu:
Additional note regarding PR description:

> Android Studio
> 
> The recommended Android Studio version to use with AGP 8.4 is [Android Studio Jellyfish](https://developer.android.com/build/releases/gradle-plugin).

I consider it a '**hard requirement**' for users to upgrade to **Android Studio Jellyfish** at a minimum for successful builds and seamless functionality from Android Studio, as the term '**recommended**' suggests that the version is not mandatory.

Older versions of Android Studio are not compatible with AGP 8.4 and will fail. 

For instance, attempting to build on the following older versions of Android Studio will result in the following error messages:

**Android Studio Hedgehog**

```
The project is using an incompatible version (AGP 8.4.0) of the Android Gradle plugin. The latest supported version is AGP 8.2.0-rc03.
See Android Studio & AGP compatibility options.
```

**Android Studio Iguana**

```
The project is using an incompatible version (AGP 8.4.0) of the Android Gradle plugin. The latest supported version is AGP 8.3.0-alpha14.
See Android Studio & AGP compatibility options.
```

breautek:
> Also the .ratignore changes could be reverted, but if reverted, need to add at least gradle-wrapper.properties. This this file is detected if anyone runs locally a rat test after running npm test.

Sounds like we can keep the `.ratignore` as is then, cause if it detects those files after a test run, then you'll probably need the gradlew scripts to be ignored as well.

breautek:
> I consider it a 'hard requirement' for users to upgrade to Android Studio Jellyfish at a minimum for successful builds and seamless functionality from Android Studio, as the term 'recommended' suggests that the version is not mandatory.
>
>Older versions of Android Studio are not compatible with AGP 8.4 and will fail.

I've updated the wording, also added a smaller Breaking Changes section that highlights in short bullet-form of the notable breaking changes.

breautek:
I've reverted AGP down to 8.3

Jellyfish appears to bundle Gradle 8.4 by default and will fail on installing the wrapper due to AGP 8.4's requirement on Gradle 8.6. In order to use AGP 8.4 on Jellyfish, the user would have to do additional configurations to specify an alternate gradle.

erisu (approved):
Code LGTM

I have tested the following:

- `cordova platform add`
- `cordova build android`
- Android Studio Jellyfish
  - Gradle Sync
  - Build and Run on Emulator
  
I confirmed that Gradle 8.7 and AGP 8.3 is used.

ramona-cristea:
Hello,

Does Cordova support Android 14? I saw you were working on it (api34-support), but didn't see any update on the oficial documentation here https://cordova.apache.org/docs/en/12.x/guide/platforms/android/index.html ?
Is there a version 13.0.0 of Cordova which targets Android 14? Thank you!

breautek:
> Is there a version 13.0.0 of Cordova which targets Android 14? Thank you!

Yes. See [here](https://cordova.apache.org/announcements/2024/05/23/cordova-android-13.0.0.html) for more details.

PR #1696, opened 2024-02-25
feat: Add ResolveServiceWorkerRequests preference

jcesarmobile:
Ionic webview plugin has had this preference for a few years, it makes possible for service worker requests to go through the asset loader.

### DIFF
--- a/framework/src/org/apache/cordova/engine/SystemWebViewClient.java
+++ b/framework/src/org/apache/cordova/engine/SystemWebViewClient.java
@@ -118,7 +118,7 @@ public class SystemWebViewClient extends WebViewClient {
         });
 
         this.assetLoader = assetLoaderBuilder.build();
-        boolean setAsServiceWorkerClient = parentEngine.preferences.getBoolean("ResolveServiceWorkerRequests", false);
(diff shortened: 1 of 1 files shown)


erisu (approved):
LGTM

breautek (approved):

framework/src/org/apache/cordova/engine/SystemWebViewClient.java
```diff
@@ -116,6 +118,18 @@ public SystemWebViewClient(SystemWebViewEngine parentEngine) {
         });
 
         this.assetLoader = assetLoaderBuilder.build();
+        boolean setAsServiceWorkerClient = parentEngine.preferences.getBoolean("ResolveServiceWorkerRequests", false);
```
breautek, line 121: I feel like if this is going to be introduced in the 13.x release then we could default to `true` so that service workers just "works" out of the box.

But I don't have a strong opinion to block this and I'm okay if this remains using `false` as a default.

jcesarmobile: I've set it to be the same default value as `cordova-plugin-ionic-webview` in case people is migrating from there, but if we want it to be `true` and it's not going to be released as a minor, I can change it.

erisu (approved):
LGTM

PR #1718, opened 2024-06-20
Fix for app restart when BT keyboard is connected in some devices

5r1m:
### Platforms affected
From API 13

### Motivation and Context
To avoid app restart on connecting/disconnecting BT keyboard in some devices. Issue is faced by many and is also described here: https://stackoverflow.com/questions/25735227/bluetooth-keyboard-will-cause-activity-destroy-and-recreate/27238892#27238892

There is no easy way to extend this using config's.

### Description
In some devices (especially pixel) connecting with BT keyboard is resulting in application being restarted. The existing "keyboard" seems to be not sufficient as in this case it was triggering "navigation" change.

### Testing
Tested on problematic devices by connecting and disconnecting keyboards

### Checklist

### DIFF
--- a/templates/project/AndroidManifest.xml
+++ b/templates/project/AndroidManifest.xml
@@ -39,7 +39,7 @@
                 android:launchMode="singleTop"
                 android:theme="@style/Theme.App.SplashScreen"
                 android:windowSoftInputMode="adjustResize"
-                android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|smallestScreenSize|screenLayout|uiMode"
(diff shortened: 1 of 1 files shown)


breautek (approved):
lgtm

breautek:
If there are no objections I'll merge sometime next week. Or earlier if I have other green votes.

I do not believe this is a breaking change.

https://developer.android.com/guide/topics/manifest/activity-element#config states:

> TA change to the navigation type (trackball or D-pad). Normally, this does not happen.

For the "navigation". Cordova apps primarily operate in the webview, the webview should be able to handle input changes dynamically.

dpogue (approved):
LGTM

PR #1746, opened 2024-11-13
fix: remove pollOnce from NativeToJsMessageQueue.popAndEncodeAsJs

### LINKED ISSUE #1735: module cordova/plugin/android/polling not found
We are seeing that some users are experiencing this error `module cordova/plugin/android/polling not found`

Which seems to spawn from this queue polling function, that points to a module that dosent seem to exist?

https://github.com/apache/cordova-android/blob/3503bfa31b10cc7d6948356d4c101c7b76109688/framework/src/org/apache/cordova/NativeToJsMessageQueue.java#L209

erisu:
### Motivation and Context

Remove the require block for `cordova/plugin/android/polling`, a module that doesn't exist.

Resolves #1735 

### Description

The `cordova/plugin/android/polling` module appears to have been removed since 2012.

The last known version of Cordova-JS where this module existed was **[cordova-js@2.1.0 - polling.js](https://github.com/apache/cordova-js/blob/2.1.0/lib/android/plugin/android/polling.js)** in the `polling.js` file, with the version tag created on Sep 13, 2012.

The module was removed in `cordova-js@2.2.0rc1`, a tag/release created on Oct 16, 2012.

The **[exact commit](https://github.com/apache/cordova-js/commit/01a2683adffd6198f6d28344b1d2c87ba98023b7)** that removed the module from Cordova-JS does not have an associated PR, GitHub issue, or CB ticket to explain the reason for this change.

Additionally, the platform-specific Cordova-JS modules, which reside in the `cordova-js-src` directory of this repository ([originally labeled as `platform_modules`](https://github.com/apache/cordova-android/commit/828edb3a439ace4ecd169ee8e87a93418fc623b9)), were not created until 2015—2 years and 4 months later. This directory has also never contained the polling module as far as I can see.

Since this part of the code has not been functional since 2012, I believe it is safe to remove it.

### Testing

n/a

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/NativeToJsMessageQueue.java
+++ b/framework/src/org/apache/cordova/NativeToJsMessageQueue.java
@@ -205,9 +205,6 @@ public class NativeToJsMessageQueue {
                     sb.append("}finally{");
                 }
             }
-            if (!willSendAllMessages) {
-                sb.append("window.setTimeout(function(){cordova.require('cordova/plugin/android/polling').pollOnce();},0);");
-            }
             for (int i = willSendAllMessages ? 1 : 0; i < numMessagesToSend; ++i) {
                 sb.append('}');
             }


dpogue:
I believe the polling implementation was brought into the main exec file, and still exists here: https://github.com/apache/cordova-android/blob/5a2c50d1ed67c1cdad0d12f1b6fbbd04ab55dff6/cordova-js-src/exec.js#L120-L132

I don't know whether that means we should be calling it from Java or not... clearly this has been broken for a long while

erisu:
> I believe the polling implementation was brought into the main exec file, and still exists here:
> 
> https://github.com/apache/cordova-android/blob/5a2c50d1ed67c1cdad0d12f1b6fbbd04ab55dff6/cordova-js-src/exec.js#L120-L132
> 
> I don't know whether that means we should be calling it from Java or not... clearly this has been broken for a long while

Might be able to expose that method by added it to `androidExec`: 

```javascript
androidExec.pollOnce = pollOnce;
```

Then instead of deleting the line, update it to:

```java
            if (!willSendAllMessages) {
                sb.append("window.setTimeout(function(){cordova.require('cordova/exec').pollOnce();},0);");
            }
```

But I don't know what the behavior it will have after adding a pollOnce call to something that never ran for 12 years...

PR #1761, opened 2025-01-21
fixed maxBuildToolVersion calculation

ssiyo:
### Platforms affected
Android

### Motivation and Context

The `doFindLatestInstalledBuildTools` function in `cordova.gradle` has a bug: it can exclude valid Build Tools versions.
I installed a version newer than MIN_BUILD_TOOLS_VERSION but within the same major version, in my case the minimum version is 34.0.0, my installed version is 35.0.1. The script considers 35.0.1 not within 34.0.0-35.0.0, which is a range that incorrectly calculates its upper bound.

### Description

changed line:70 `def maxVersion = new Version((minBuildToolsVersion.getMajor() + 1) + ".0.0")`
to `def maxVersion = new Version("999.0.0")`

### Testing

Ran `cordova build android`

### Checklist

### DIFF
--- a/framework/cordova.gradle
+++ b/framework/cordova.gradle
@@ -58,7 +58,7 @@ String doFindLatestInstalledBuildTools(String minBuildToolsVersionString) {
     }
 
     def minBuildToolsVersion = new Version(minBuildToolsVersionString)
-    def maxVersion = new Version((minBuildToolsVersion.getMajor() + 1) + ".0.0")
+    def maxVersion = new Version("999.0.0")
 
     def highestBuildToolsVersion = buildToolsDirContents
         .collect { new Version(it) }


breautek:
The intent is to allow >= the min build tools version, but not allow the next major version.

```
def maxVersion = new Version((minBuildToolsVersion.getMajor() + 1) + ".0.0")
```

This is exactly what this grabs. so if the min version is `34.0.1`, then the max version value is `35.0.0`, the equation is `version >= 34.0.1 && version < 35.0.0` (exclusive check, in otherwords build tools version `35.0.0` will also fail and this is intentional).

> The script considers 35.0.1 not within 34.0.0-35.0.0, which is a range that incorrectly calculates its upper bound.

35.0.1 is in fact not within  34.0.0-35.0.0.

Updating the max version to be hard-coded arbitrary value to allow anything above the min version will result unexpected breakage. The build tools package often includes breaking changes.

Instead you should just install/use the latest available patch of version `34` of build tools.

PR #1772, opened 2025-01-28
refactor: replace fs-extra with node:fs

erisu:
### Platforms affected

n/a

### Motivation and Context

Replace `fs-extra` dependency with node internals `node:fs`

### Description

Remove `fs-extra` and updated code to require and use `node:fs` APIs.

### Testing

`npm t`

### Checklist

### DIFF
--- a/lib/builders/ProjectBuilder.js
+++ b/lib/builders/ProjectBuilder.js
@@ -342,7 +342,7 @@ class ProjectBuilder {
         const args = this.getArgs('clean', opts);
         return execa(wrapper, args, { stdio: 'inherit', cwd: path.resolve(this.root) })
             .then(() => {
-                fs.rmSync(path.join(this.root, 'out'));
+                fs.rmSync(path.join(this.root, 'out'), { recursive: true, force: true });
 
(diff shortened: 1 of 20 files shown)


lib/builders/ProjectBuilder.js
```diff
@@ -342,15 +342,15 @@ class ProjectBuilder {
         const args = this.getArgs('clean', opts);
         return execa(wrapper, args, { stdio: 'inherit', cwd: path.resolve(this.root) })
             .then(() => {
-                fs.removeSync(path.join(this.root, 'out'));
+                fs.rmSync(path.join(this.root, 'out'));
```
dpogue, line 345: Does this need `{ recursive: true, force: true }`?

erisu: It was not exactly clear if "out" was a directory or file. 

I thought this was an output log file.

I could not find in code where this directory or file is being created. But looking at the test spec, it refers to it as a directory. 

If correct, adding `{ recursive: true, force: true }` is required, which I just pushed.

dpogue (approved):
Created a project and ran prepare & build with this locally and it all seems to be working :+1:

PR #1778, opened 2025-02-10
feat: Account for Node security patch

kant2002:
As of https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2#command-injection-via-args-parameter-of-child_processspawn-without-shell-option-enabled-on-windows-cve-2024-27980---high Cordova produce unrecognized error on Windows.

Fixes: https://github.com/apache/cordova-cli/issues/456

### DIFF
--- a/lib/check_reqs.js
+++ b/lib/check_reqs.js
@@ -110,6 +110,8 @@ module.exports.get_gradle_wrapper = function () {
     let program_dir;
     // OK, This hack only works on Windows, not on Mac OS or Linux.  We will be deleting this eventually!
     if (module.exports.isWindows()) {
+        // "spawn" option enabled for CVE-2024-27980 (Windows) Mitigation
+        // See https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2 for more details
         const result = execa.sync(path.join(__dirname, 'getASPath.bat'), { spawn: true });
         // console.log('result.stdout =' + result.stdout.toString());
         // console.log('result.stderr =' + result.stderr.toString());
--- a/lib/check_reqs.js
+++ b/lib/check_reqs.js
@@ -110,7 +110,7 @@ module.exports.get_gradle_wrapper = function () {
     let program_dir;
     // OK, This hack only works on Windows, not on Mac OS or Linux.  We will be deleting this eventually!
     if (module.exports.isWindows()) {
-        const result = execa.sync(path.join(__dirname, 'getASPath.bat'));
+        const result = execa.sync(path.join(__dirname, 'getASPath.bat'), { spawn: true });
         // console.log('result.stdout =' + result.stdout.toString());
         // console.log('result.stderr =' + result.stderr.toString());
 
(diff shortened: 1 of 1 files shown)


erisu (commented):
Can you confirm if this change is valid?

The article you're linking is about Node.js `spawn` while the code appears to be using the npm package `execa` and the README for version 5.1.1 of this package does not to mention `spawn` as a valid option.

erisu:
The article is talking about the `shell` option and the `execa` package does have this flag.

https://www.npmjs.com/package/execa/v/5.1.1#shell

Bit curious, since the dependency suggests setting it is

> unsafe, potentially allowing command injection.

breautek:
> The article is talking about the `shell` option and the `execa` package does have this flag.
> 
> https://www.npmjs.com/package/execa/v/5.1.1#shell
> 
> Bit curious, since the dependency suggests setting it is
> 
> > unsafe, potentially allowing command injection.
> 
> But since we are defining the bat file and not taking in user arg, it should be OK.

I assume execa is using `spawn` behind-the-scenes, and it's generally recommended to not enable `shell` because it can be a security concern:

This is from the NodeJS docs

>If the shell option is enabled, do not pass unsanitized user input to this function. Any input containing shell metacharacters may be used to trigger arbitrary command execution.

The security concern arises if we are passing in user supplied values, which we are not:

```
path.join(__dirname, 'getASPath.bat')
```

`path` is a NodeJS API, and `__dirname` is a module-level NodeJS variable, concating 
 hard-coded string `getASPath.bat`. So I think using `shell` option here is safe.

I do think we should comment why we are using `shell` though so that someone else doesn't for optimization or to "harden".

lib/check_reqs.js
```diff
@@ -110,7 +110,7 @@ module.exports.get_gradle_wrapper = function () {
     let program_dir;
     // OK, This hack only works on Windows, not on Mac OS or Linux.  We will be deleting this eventually!
     if (module.exports.isWindows()) {
-        const result = execa.sync(path.join(__dirname, 'getASPath.bat'));
+        const result = execa.sync(path.join(__dirname, 'getASPath.bat'), { spawn: true });
```
breautek, line 113:
```suggestion
        // "spawn" option enabled for CVE-2024-27980 (Windows) Mitigation
        // See https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2 for more details
        const result = execa.sync(path.join(__dirname, 'getASPath.bat'), { spawn: true });
```

breautek: Also, I can't find any `spawn` option documented in execa, is this suppose to be `shell` ?

kant2002: I indeed manage to made mistake in the one-liner. Thank you. It's `shell`

lib/check_reqs.js
```diff
@@ -110,7 +110,9 @@ module.exports.get_gradle_wrapper = function () {
     let program_dir;
     // OK, This hack only works on Windows, not on Mac OS or Linux.  We will be deleting this eventually!
     if (module.exports.isWindows()) {
-        const result = execa.sync(path.join(__dirname, 'getASPath.bat'));
+        // "spawn" option enabled for CVE-2024-27980 (Windows) Mitigation
```
breautek, line 113:
```suggestion
        // "shell" option enabled for CVE-2024-27980 (Windows) Mitigation
```

I wrote in "spawn" because that was in the original PR but since we determine the "shell" option is correct, that should be reflected in the comment too.

erisu (approved):
Changes looks good to me.

PR #1779, opened 2025-02-14
feat: add AndroidEdgeToEdge preference & theme flag

erisu:
### Motivation and Context

- Disable the Edge-to-Edge setting that SDK 35 enabled by default.
- Allow users to opt-in to the Edge-to-Edge feature with a flag.

In the next Cordova-Android release, we will temporarily disable the Edge-to-Edge feature to maintain the same behavior that existed previously. The new feature appears to have issues with safe-area-*.

When setting `viewport-fit` to `auto` or `contain`, the content will render behind the safe-area which is believed to be an unintended behavior. iOS for example will automatically render the content below the statusbar. 

### Description

- Added theme flag `android:windowOptOutEdgeToEdgeEnforcement` and default to `true`
- Added preference flag `AndroidEdgeToEdge` to allow changing of the theme flag
  - When `AndroidEdgeToEdge` is set to `true`, `android:windowOptOutEdgeToEdgeEnforcement` will be `false`
  - When `AndroidEdgeToEdge` is set to `false`, an **empty string**, **not defined**, or **invalid entry**, `windowOptOutEdgeToEdgeEnforcement` will be `true`

### Testing

- build app and test flag

### Checklist

### DIFF
--- a/lib/prepare.js
+++ b/lib/prepare.js
@@ -381,6 +381,26 @@ function updateProjectTheme (platformConfig, locations) {
     const themes = xmlHelpers.parseElementtreeSync(locations.themes);
     const splashScreenTheme = themes.find('style[@name="Theme.App.SplashScreen"]');
 
+    // Update edge-to-edge settings in app theme.
+    let hasE2E = false; // default case
+
+    const preferenceE2E = platformConfig.getPreference('AndroidEdgeToEdge', this.platform);
+    if (!preferenceE2E) {
+        events.emit('verbose', 'The preference name "AndroidEdgeToEdge" was not set. Defaulting to "false".');
+    } else {
+        const hasInvalidPreferenceE2E = preferenceE2E !== 'true' && preferenceE2E !== 'false';
+        if (hasInvalidPreferenceE2E) {
+            events.emit('verbose', 'Preference name "AndroidEdgeToEdge" has an invalid value. Valid values are "true" or "false". Defaulting to "false"');
+        }
+        hasE2E = hasInvalidPreferenceE2E ? false : preferenceE2E === 'true';
+    }
+
+    const optOutE2EKey = 'android:windowOptOutEdgeToEdgeEnforcement';
+    const optOutE2EItem = splashScreenTheme.find(`item[@name="${optOutE2EKey}"]`);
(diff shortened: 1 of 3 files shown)


dpogue:
Note that because Edge-To-Edge is disabled by default, `viewport-fit=cover` also won't do anything.

You can opt-in to working `viewport-fit=cover` with `safe-area-inset-top` values by setting the `AndroidWindowOptOutEdgeToEdgeEnforcement` preference to false in config.xml.
(Further note: `safe-area-inset-bottom` is never correct on Android, so be aware of that if you are wanting to opt-in)

dpogue:
Just a note that the `windowOptOutEdgeToEdgeEnforcement` option is removed in Android 16 (as per https://android-developers.googleblog.com/2025/02/second-beta-android16.html)

> ### Edge to edge opt-out going away
> [Android 15 enforced edge-to-edge](https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge) for apps targeting Android 15 (SDK 35), but your app could opt-out by setting [`R.attr#windowOptOutEdgeToEdgeEnforcement`](https://developer.android.com/reference/android/R.attr#windowOptOutEdgeToEdgeEnforcement) to true. Once your app targets Android 16 (Baklava), `R.attr#windowOptOutEdgeToEdgeEnforcement` is deprecated and disabled and your app cannot opt-out of going edge-to-edge. To be compatible with Android 16 Beta 2, ensure your app supports edge-to-edge and remove any use of `R.attr#windowOptOutEdgeToEdgeEnforcement`. To support edge-to-edge, see the [Compose](https://developer.android.com/develop/ui/compose/layouts/insets) and [Views](https://developer.android.com/develop/ui/views/layout/edge-to-edge) guidance. Please let us know about concerns in our tracker on the [feedback page](https://developer.android.com/about/versions/16/feedback).

erisu:
Last call for reviews before merging in.

dpogue:
I do sortof question if it's worth merging this knowing that it's going to be deprecated immediately in the next release.

Unfortunately there aren't really any alternatives at the moment, so it's probably better to have something than nothing :\

dpogue (approved):

PR #1781, opened 2025-02-19
fix: copy gradle wrapper from tools to platform root dir

### LINKED ISSUE #1760: Impossible to open Project in Android Studio
Hi,

We are using Cordova since a long time. Build is well working through the Cordova CLI.

However, i'm wondering why gradle is now moved to "tools" folder into the android project.

Due to that, we cannot import the project into Android Studio.  I have the following error.

```
> Failed to apply plugin 'com.android.internal.version-check'.
   > Minimum supported Gradle version is 8.4. Current version is 8.2. If using the gradle wrapper, try editing the distributionUrl in /Users/<user_name>/workspace/<client_name>/<project_name>/platforms/android/gradle/wrapper/gradle-wrapper.properties to gradle-8.4-all.zip`

The gradle folder is placed into a TOOLS folder...
```

Do I miss something?

Thanks in advance

Ionic:

   Ionic CLI                     : 7.2.0
   Ionic Framework               : @ionic/angular 6.3.2
   @angular-devkit/build-angular : 13.3.9
   @angular-devkit/schematics    : 13.3.9
   @angular/cli                  : 13.3.9
   @ionic/angular-toolkit        : 6.1.0

Cordova:

   Cordova CLI       : 12.0.0 (cordova-lib@12.0.2)
   Cordova Platforms : android 13.0.0
   Cordova Plugins   : cordova-plugin-ionic-keyboard 2.2.0, cordova-plugin-ionic-webview 5.0.0, (and 15 other plugins)

Utility:

   cordova-res : 0.15.4
   native-run  : not installed globally

System:

   Android SDK Tools : 26.1.1
   ios-deploy        : 1.12.2
   ios-sim           : 8.0.2
   NodeJS            : v16.20.2 
   npm               : 8.19.4
   OS                : macOS Unknown
   Xcode             : Xcode 15.4 Build version 15F31d

Kr.

TiBz0u

erisu:
### Motivation and Context

Copy the gradle wrapper files from `platforms/android/tools` to `platforms/android`

Fixes #1760

### Description

Updated install step to just copy all the gradle files up one directory.

### Testing

- Created a Cordova Project
- Added Platform
- Built Android Platform

### Checklist

### DIFF
--- a/lib/builders/ProjectBuilder.js
+++ b/lib/builders/ProjectBuilder.js
@@ -18,6 +18,7 @@
 */
 
 const fs = require('node:fs');
+const fsp = require('node:fs/promises');
 const path = require('node:path');
 const execa = require('execa');
 const glob = require('fast-glob');
@@ -289,6 +290,10 @@ class ProjectBuilder {
             .then(function () {
                 events.emit('verbose', `Using Gradle: ${config.GRADLE_VERSION}`);
                 return self.installGradleWrapper(config.GRADLE_VERSION);
+            }).then(async function () {
+                await fsp.cp(path.join(self.root, 'tools', 'gradle'), path.join(self.root, 'gradle'), { recursive: true, force: true });
+                await fsp.cp(path.join(self.root, 'tools', 'gradlew'), path.join(self.root, 'gradlew'), { recursive: true, force: true });
+                await fsp.cp(path.join(self.root, 'tools', 'gradlew.bat'), path.join(self.root, 'gradlew.bat'), { recursive: true, force: true });
             }).then(function () {
                 return self.prepBuildFiles();
             }).then(() => {


breautek:
> removed tools/settings.gradle. Appears not needed.

It will be needed if the installed gradle at the system level is less than what AGP requires. That was the whole point of adding it, to serve as a dummy project with no dependencies so that we can get our gradle wrapper.

breautek:
To solve https://github.com/apache/cordova-android/issues/1760

I think what we should do is copy the generated `gradle`, `gradlew`, and `gradlew.bat` files up a directory on each gradle wrapper install into `platforms/android/`. If the wrapper files already exists, then it should be replaced. We could also experiment with creating a symlink to `tools/gradlew`, `tools/gradle`, and `tools/gradlew.bat` (but idk if Android Studio will follow symlinks or if that will play nicely on windows)

We should also retain the tools/ gradle wrapper copy for future install commands. Like I mentioned, the `tools/` project is a dummy project with no dependencies, so it allows for the user to have a wide range of system gradle installed. On each build, the install command is issued, but if the wrapper is already installed at the desired version, it's nearly a no-operation, so it will help subsequent builds as well.

Android Studio will use the wrapper installed at `platforms/android/`. The CLI could continue to use the wrapper in `platforms/android/tools/` or be updated to use the copied version at `platforms/android`, that's up to preference at this point. But the CLI should make sure that `platforms/android/` gradle install is identical to the one inside `platform/android/tools/`.

I hope this makes sense. I know having a copy is not ideal, but it helps give us the best of both worlds:

1. Virtually any Gradle version can be used to install the gradle wrapper of the version needed by Cordova.
2. Installed gradle wrapper will be used by Android Studio if the project is opened in the IDE.

If the symlink idea works, then that could be use to reduce the maintenance burden as well.

breautek:
This is kind of out of scope for https://github.com/apache/cordova-android/issues/1760 but I also notice that we don't install the wrapper on create. I think it may only occur on build.

We should also consider doing a gradle install on project creation too (on `cordova platform add android`) so that the wrapper is immediately available. We should still continue to install on build as well in case they change the gradle version preference...

As it stands right now, the wrapper is only installed on `cordova build android` which means if someone added the android platform and then opened it up in Android Studio, it won't have the expected wrapper version, until they run a `cordova build android`

Just something I noticed while testing the current behaviour.

erisu:
> we should do is copy the generated gradle, gradlew, and gradlew.bat files up a directory on each gradle wrapper install into platforms/android/. 

This PR now copy up a directory as suggested.

> If the wrapper files already exists, then it should be replaced. 

This should be complete. It always runs a copy with recursive+force on build.

> experiment with creating a symlink 

Symlink will not be used because we need to ensure that the files remain when the platforms directory are zipped. It seems that there is a use case to prepare the platforms in one environment and built in another.

> We should also retain the tools/ gradle wrapper copy for future install commands. 

Done

> Android Studio will use the wrapper installed at platforms/android/. The CLI could continue to use the wrapper in platforms/android/tools/

CLI uses tools version.

> CLI should make sure that platforms/android/ gradle install is identical to the one inside platform/android/tools/

CLI should always call the install gradle wrapper method and copy step during build. This would handle the case if users changes the version by preference.

This PR does not, and will not, cover the cases if the users upgrades from android studio. CLI build will undo those changes and replace grade back with tools version.

This PR will not cover creating of the gradle wrapper on `platform add`.

breautek (approved):

PR #1785, opened 2025-03-17
feat: bump gradle to 8.13

erisu:
### Platforms affected

n/a

### Motivation and Context

- Use latest Gradle
- [Resolve FileLockCommunicator compatible issue with macOS 15.1 firewall](https://github.com/gradle/gradle/issues/31245)

### Description

Update default version and test to 8.13

### Testing

- `npm t`
- Build test

### Checklist

### DIFF
--- a/framework/cdv-gradle-config-defaults.json
+++ b/framework/cdv-gradle-config-defaults.json
@@ -2,7 +2,7 @@
     "MIN_SDK_VERSION": 24,
     "SDK_VERSION": 35,
     "COMPILE_SDK_VERSION": null,
-    "GRADLE_VERSION": "8.9",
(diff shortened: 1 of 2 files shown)


dpogue (commented):
This looks fine to me, but I'm not up to speed on Gradle versions and AGP requirements, so appreciate if someone more familiar with the Android side of things can also take a look

breautek (approved):
Gradle versions are pretty safe to update as long as you're not crossing major versions.

It's the AGP that can introduce breaking changes on any upgrade, as they don't practice semver, but another versioning scheme.

AGP 8.9 which I believe is the current latest requires minimum gradle 8.9 so this changes looks good to me.

PR #1790, opened 2025-03-21
fix: replace fs-extra.ensureFileSync with fs.writeFileSync

erisu:
### Motivation and Context

Fix undefined `ensureFileSync` method.

### Description

The `fs-extra` npm package was removed which caused `ensureFileSync` to be not a function.

This PR replaces `ensureFileSync` with `writeFileSync`. The signing properties file is always removed before being recreated, and the platform directory should always exist, so it should be safe to call `writeFileSync`.

### Testing

- npm t
- created a testing debug keystore file
- configured `build.json` to use the testing debug keystore. 
- cordova build android
- cordova run android

### Checklist

### DIFF
--- a/lib/builders/ProjectBuilder.js
+++ b/lib/builders/ProjectBuilder.js
@@ -301,7 +301,7 @@ class ProjectBuilder {
 
                 if (fs.existsSync(signingPropertiesPath)) fs.rmSync(signingPropertiesPath);
                 if (opts.packageInfo) {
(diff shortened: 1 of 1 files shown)


dpogue (approved):
This looks fine to me.

With regards to the filesystem race condition, since we always overwrite the file if `opts.packageInfo` is set, is it fine to move the deletion to an `else` block and do it with `{ force: true }` so we don't need to check existence?

breautek (approved):
lgtm

PR #1795, opened 2025-03-29
fix: configure gradle java.home

erisu:

Fix an issue mentioned in a [discussion thread](https://github.com/apache/cordova/discussions/530#discussioncomment-12605698).

1. Ensure that Cordova-Android properly creates `config.properties`.
2. Ensure that `java.home` is set correctly:
   1. Check and use the `CORDOVA_JAVA_HOME` environment variable.
   2. If `CORDOVA_JAVA_HOME` is missing, fall back to `JAVA_HOME`.
   3. If neither is set, remove the `java.home` property (if previously set), allowing Android Studio to display a warning and configure its internal Java automatically.
3. Copy the `tools/.gradle` directory up one level.

**Note:** Step 2.3 should not occur under normal conditions, but was included as a safeguard. When running `cordova build`, if `JAVA_HOME` is missing, its expected to not continue as its apart of the requirement checks.

This change also ensures that Cordova-Android uses the correct Java version specified by the user via environment variables. According to the Cordova documentation, these variables must be configured for Cordova-Android setup. Defaulting to Android Studio's bundled Java could lead to unexpected results and not using versions that the App developer configured.

### Additional Feedback

It appears that the change came from AGP 8.2. It created a [new way to specify the JDK path](https://developer.android.com/build/releases/past-releases/agp-8-2-0-release-notes#jdk-macro) was added and became default. This new macro is called `GRADLE_LOCAL_JAVA_HOME`

From [Android Developer - Java versions in Android builds](https://developer.android.com/build/jdks):

> The macros enable dynamic project JDK path selection:
> - `JAVA_HOME`: uses the environment variable with the same name
> - `GRADLE_LOCAL_JAVA_HOME`: uses the java.home property in the .gradle/config.properties file which defaults to the JetBrains Runtime.

### Testing

- Create project
- Add cordova-android platform with changes
- Build for Android

1. Use Case 1.a

Confirmed that `java.home` was set with configured environment variables.

2. Use Case 1.b

Re-run build with `CORDOVA_JAVA_HOME` set to another version of Java and confirmed the `java.home` was updated.

3. Use Case 2

Created a `before_build` hookscript  which would run before  `cordova build`.
Confirm that the additional properties, defined in the hookscript, were injected in the `config.properties` file and that `java.home` was set correctly by the build script.

### Checklist

### DIFF
--- a/lib/builders/ProjectBuilder.js
+++ b/lib/builders/ProjectBuilder.js
@@ -286,14 +286,49 @@ class ProjectBuilder {
     prepEnv (opts) {
         const self = this;
         const config = this._getCordovaConfig();
+        const configPropertiesPath = path.join(self.root, 'tools', '.gradle', 'config.properties');
+
         return check_reqs.check_gradle()
             .then(function () {
                 events.emit('verbose', `Using Gradle: ${config.GRADLE_VERSION}`);
                 return self.installGradleWrapper(config.GRADLE_VERSION);
             }).then(async function () {
+                try {
+                    // Try to create "config.properties" file if missing
+                    const fd = fs.openSync(configPropertiesPath, 'wx+', 0o600);
+                    fs.writeFileSync(fd, '', 'utf8');
+                    fs.closeSync(fd);
+                } catch {
+                    // File already existed, nothing to do.
+                }
+
+                // Loads "config.properties"for editing
+                const configProperties = createEditor(configPropertiesPath);
+                /*
+                 * File is replaced on each build. "before_build" hook scripts can inject
+                 * custom settings before this step runs. This step will still overwrite
+                 * the "java.home" setting. Ensure environment variables are properly
+                 * configured if want to use custom "java.home".
+                 *
+                 * Sets "java.home" using the "CORDOVA_JAVA_HOME" environment variable.
+                 * If unavailable, fallback to "JAVA_HOME".
+                 * If neither is set, "java.home" is removed (if previously set),
+                 * allowing Android Studio to display a warning and auto-configure
+                 * to use its internal (bundled) Java.
+                 */
+                const javaHome = process.env.CORDOVA_JAVA_HOME || process.env.JAVA_HOME || false;
+                if (javaHome) {
+                    configProperties.set('java.home', javaHome);
+                } else {
+                    configProperties.unset('java.home');
+                }
+                // Saves changes
+                configProperties.save();
+            }).then(async function () {
                 await fsp.cp(path.join(self.root, 'tools', 'gradle'), path.join(self.root, 'gradle'), { recursive: true, force: true });
                 await fsp.cp(path.join(self.root, 'tools', 'gradlew'), path.join(self.root, 'gradlew'), { recursive: true, force: true });
                 await fsp.cp(path.join(self.root, 'tools', 'gradlew.bat'), path.join(self.root, 'gradlew.bat'), { recursive: true, force: true });
+                await fsp.cp(path.join(self.root, 'tools', '.gradle'), path.join(self.root, '.gradle'), { recursive: true, force: true });
             }).then(function () {
                 return self.prepBuildFiles();
             }).then(() => {


dpogue (approved):
LGTM

GitToTheHub:
Hi,

### `JAVA_HOME` not applied in `config.properties`
I did set `JAVA_HOME`:

`/Library/Java/JavaVirtualMachines/jdk-17.jdk/Contents/Home`

(macOS)

`.gradle/config.properties` has the content:

```properties
#Sun Mar 30 17:59:48 CEST 2025
java.home=/Applications/Android Studio.app/Contents/jbr/Contents/Home
```

Points to something in the Android Studio app

Is this normal?

### Build without JAVA_HOME or CORDOVA_JAVA_HOME

You wrote:

> When running cordova build, if JAVA_HOME is missing, its expected to not continue as its apart of the requirement checks.

When I didn't set anything I can build successfully. The `java` command is still available, maybe it's working because of this.

erisu:
Can you provide the exact step you used to get that result?

I don't get Android Studio's `jbr` with this PR.

```
% echo $JAVA_HOME
/Users/cordova/.sdkman/candidates/java/current

% cordova create cordovaTest com.erisu.cordovaTest cordovaTest
% cd cordovaTest
% cordova platform add github:erisu/cordova-android\#feat/set-gradle-java-home
% cordova build android

% cat platforms/android/.gradle/config.properties
java.home=/Users/cordova/.sdkman/candidates/java/current
```

If you are not building with Cordova and only opening with first with Android Studio, you wont have the right value and there is nothing we can do about that.

GitToTheHub:
Hi,

after adding the platform via

`cordova platform add github:erisu/cordova-android\#feat/set-gradle-java-home`

I opend it with Android Studio and get the result I described. I didn't build anything in Android Studio, just opened it.

GitToTheHub:
> > ### Build without JAVA_HOME or CORDOVA_JAVA_HOME
> > You wrote:
> > > When running cordova build, if JAVA_HOME is missing, its expected to not continue as its apart of the requirement checks.
> > 
> > 
> > When I didn't set anything I can build successfully. The `java` command is still available, maybe it's working because of this.
> 
> Here is example of what I mean by builds wont work when `JAVA_HOME` is missing.
> 
> ```
> % echo $JAVA_HOME
> /Users/cordova/.sdkman/candidates/java/current
> 
> % unset JAVA_HOME
> % echo $JAVA_HOME
> 
> % cordova build android
> Checking Java JDK and Android SDK versions
> ANDROID_HOME=/Users/erisu/Library/Android/sdk (recommended setting)
> ANDROID_SDK_ROOT=/Users/erisu/Library/Android/sdk (DEPRECATED)
> Failed to find 'JAVA_HOME' environment variable. Try setting it manually.
> ```
> 
> Build failed the requirement check and stopped.

I did it like you wrote, but the build succeded:

```bash
% echo $JAVA_HOME
/Library/Java/JavaVirtualMachines/jdk-17.jdk/Contents/Home
% unset JAVA_HOME
% echo $JAVA_HOME

% cordova build android
ANDROID_HOME=/Users/manuelbeck/Library/Android/sdk (recommended setting)
ANDROID_SDK_ROOT=undefined (DEPRECATED)
Using Android SDK: /Users/manuelbeck/Library/Android/sdk
Reading build config file: /Users/manuelbeck/Projekte/app_entwicklung/abfuhr_app/build.json
Reading the keystore from: /Users/manuelbeck/Projekte/app_entwicklung/abfuhr_app/Signierung/Android/netactive.keystore

BUILD SUCCESSFUL in 261ms
```

erisu:
> I opend it with Android Studio and get the result I described. I didn't build anything in Android Studio, just opened it.

If you opened with Android Studio and you did not run `cordova build` first, then yes you will get Android Studio's JBR and is expected behavior.

GitToTheHub:
> If you opened with Android Studio and you did not run `cordova build` first, then yes you will get Android Studio's JBR and is expected behavior.

So I have to always first do `cordova build android` before opening it the first time in Android Studio? If yes, this should be documented.

erisu:
This PR only focuses on resolving the issue with cordova build. Running cordova build should properly detect JAVA_HOME and update the gradle config.

However, this PR does not address cases if users open Android Studio first. In that scenario, it's expected that you'll see a warning on the first load, and then Android Studio configures the java.home with JBR.

Ideally, running cordova build first shouldn’t be a requirement—users should be able to just open Android Studio. But unfortunately, fixing the issue where JBR is configured on first launch is not something that might be resolved until next major release.

Currently, Cordova doesn’t create the Gradle wrapper until the project is built using the Cordova command. As far as I remember, it has always worked this way.

To support opening a project in Android Studio immediately after adding the platform, Cordova would need to move the Gradle wrapper build process earlier in the workflow—perhaps at the end of platform add, or even during prepare, since that runs after adding a platform. If on prepare, it would also trigger on every plugin add, unfortunately.

Hook scripts already allow users to run custom scripts during key phases, like after_platform_add and before_compile. If we could also execute the Gradle wrapper build at these points, that would be ideal—but again, this would require major changes.

GitToTheHub:
Hi,

thanks for the explanation. I unterstand that this is not in scope of this pr. I only mentioned it because I noticed it.
But it sounds like you already have some ideas :) Maybe a build could be triggered after platform add? So the user sees already if everything is ok with it.

PR #1819, opened 2025-07-04
Allows additional settings to be included on settings.gradle

beroso:
### Motivation and Context
I'm integrating a flutter module inside a ionic cordova app. So accordingly to the [flutter guide](https://docs.flutter.dev/add-to-app/android/project-setup#depend-on-the-modules-source-code) I need to put the following config on settings.gradle.

```gradle
setBinding(new Binding([gradle: this]))
def filePath = settingsDir.parentFile.toString() + "/flutter_module/.android/include_flutter.groovy"
apply from: filePath 
```

But the file is regenerated on each build, and my changes are lost.

### Description

Sometimes we need some extra settings on settings.gradle file, but the file is regenerated on each build.
The proposed solution applies a `settings-additional.gradle` file (if it exists) to the `settings.gradle`.

### Checklist

### DIFF
--- a/lib/builders/ProjectBuilder.js
+++ b/lib/builders/ProjectBuilder.js
@@ -205,7 +205,8 @@ class ProjectBuilder {
             '// GENERATED FILE - DO NOT EDIT\n' +
             'apply from: "cdv-gradle-name.gradle"\n' +
             'include ":"\n' +
-            settingsGradlePaths.join(''));
+            settingsGradlePaths.join('') +
+            '\nif (file("settings-additional.gradle").exists()) { apply from: "settings-additional.gradle" }\n');
 
         // Touch empty cdv-gradle-name.gradle file if missing.
         if (!fs.existsSync(path.join(this.root, 'cdv-gradle-name.gradle'))) {


lib/builders/ProjectBuilder.js
```diff
@@ -205,7 +205,8 @@ class ProjectBuilder {
             '// GENERATED FILE - DO NOT EDIT\n' +
             'apply from: "cdv-gradle-name.gradle"\n' +
             'include ":"\n' +
-            settingsGradlePaths.join(''));
+            settingsGradlePaths.join('') +
+            '\nif (file("settings-additional.gradle").exists()) { apply from: "settings-additional.gradle" }\n');
```
breautek, line 209: we have a concept of `build-extras.gradle` which is virtually the same thing but for the build.gradle file. So I'm thinking to keep naming consist we should actually look for `settings-extra.gradle`

breautek:
I think this is a good clean enhancement and I think there isn't a risk of a breaking change. Because we have the same concept for `build-extras.gradle`, I don't think there is any issue doing the same for settings. Considering that Apache just made a release not that long ago, it might be some time before a feature release is made however.

A workaround that should work in the meantime if you don't want to use a custom version of cordova-android is the [after prepare hook](https://cordova.apache.org/docs/en/12.x-2025.01/guide/appdev/hooks/index.html). It should allow you to manipulate the `settings.gradle` after Cordova has generated the file.

PR #1821, opened 2025-07-08
Android: Fix gradle deprecation warnings about property assignment

GitToTheHub:
When building a plain Android Cordova app without plugins, there will be Gradle depraction warnings: `Deprecated Gradle features were used in this build, making it incompatible with Gradle 9.0.`. Compiling it with `--warning-mode all` shows, that some properties are assigned without an `=`, which will not be supported any longer with Gradle 10.0.

### Platforms affected
Android

### Checklist

### DIFF
--- a/framework/build.gradle
+++ b/framework/build.gradle
@@ -44,7 +44,7 @@ allprojects {
 apply plugin: 'com.android.library'
 
 android {
-    namespace 'org.apache.cordova'
+    namespace = 'org.apache.cordova'
 
     compileSdkVersion cordovaConfig.COMPILE_SDK_VERSION
     buildToolsVersion cordovaConfig.BUILD_TOOLS_VERSION
--- a/templates/project/app/build.gradle
+++ b/templates/project/app/build.gradle
@@ -183,10 +183,10 @@ task cdvPrintProps {
 }
 
 android {
-    namespace cordovaConfig.PACKAGE_NAMESPACE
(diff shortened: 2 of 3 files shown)


breautek:
> I'm guessing `namespace` needs the `=` because it's a property and `compileSdkVersion` doesn't because it's... not a property? And the only way to know that is to dig into gradle documentation? 😕
> 
> The changes seem fine to me, I just worry that now it _looks_ inconsistent and it's not going to be obvious in future changes when a `=` is needed or not (but that's Gradle's fault, not your fault)

I believe `compileSdkVersion` is also deprecated in favour for `compileSdk` which is a property that you do "assign", vs `compileSdkVersion` which is actually a method, that you supply a single argument into.

The Kotlin DSL is much more strict on this stuff, but is also much more streamlined/consistent. I think it will be pretty easy to migrate too, kotlin gradle I believe can still reference groovy gradle for extensions (like build-extras.gradle)... so might be worth considering doing in one of our major releases at some point.

breautek (approved):

GitToTheHub:
I was just correcting, what Gradle complained about. I'm also wondering, why some things should be properties and some not. When I look for the `namespace` property on Google's documentation it's even [there](https://developer.android.com/build/configure-app-module#set-namespace) not documented with an `=` for Groovy.

erisu (approved):

PR #1823, opened 2025-07-09
fix: Deprecation warning in PluginManager for using Class.newInstance

GitToTheHub:
In `instantiatePlugin`, `Class.getInstance` was called, but is [deprecated since Java 9](https://docs.oracle.com/javase/9/docs/api/java/lang/Class.html#newInstance--). Instead `Class.getDeclaredConstructor().newInstance()` should be called.

### Platforms affected
Android

### Motivation and Context
Resolving deprecation warnings for cordova-android

### Checklist


dpogue (approved):
LGTM

PR #1831, opened 2025-07-30
fix(android): Back button override on API 36+

breautek:
### Platforms affected

Android

### Motivation and Context

Android API 36 will crash on hardware back button presses due to breaking changes noted [here](https://developer.android.com/about/versions/16/behavior-changes-16#predictive-back)

### Description

Basically intercepting key events using `onDispatchKeyEvent` no longer works for back button. The return value is ignored for back key events and the system will use the system default action, in addition to whatever code is being triggered by `onDispatchKeyEvent`.

On API 36+, the predictive back feature may take you out of your app, and if it does, the activity may be destroyed. We have key event listeners that will send events to the webview, but in this circumstance the webview will be unusable leading to a crash.

To prevent system behaviour, the OnBackInvokedCallback needs to be registered with the dispatcher. This PR introduces a no-op callback created when the back button is overridden. A reference is kept so that if the app wishes to disable the override later, we can unregister the callback.

There is a comment explaining why the callback is no-op, but to summarize here, attaching the event is all that is needed to prevent default system behaviour. In my testing, the key events are still dispatched just like in earlier API levels, which handles event propagation to the webview... However this may be a bug inside Android considering it's documented that:

> ... Additionally, onBackPressed is not called and [KeyEvent.KEYCODE_BACK](https://developer.android.com/reference/android/view/KeyEvent#KEYCODE_BACK) is not dispatched anymore.

We may need to watch this closely.

### Testing

Manual testing in-app on API 36 Google Play simulator, as well as npm test passes.

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/CoreAndroid.java
+++ b/framework/src/org/apache/cordova/CoreAndroid.java
@@ -28,9 +28,12 @@ import org.json.JSONObject;
 import android.content.BroadcastReceiver;
 import android.content.Context;
 import android.content.Intent;
+import android.os.Build;
 import android.content.IntentFilter;
 import android.telephony.TelephonyManager;
 import android.view.KeyEvent;
+import android.window.OnBackInvokedCallback;
+import android.window.OnBackInvokedDispatcher;
 
 import java.util.HashMap;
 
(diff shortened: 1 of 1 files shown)


erisu (approved):
LGTM

PR #1838, opened 2025-08-21
fix(android): use remove(0) instead of removeFirst() for compatibilit…

pilz97:
…y with Android < 15

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/NativeToJsMessageQueue.java
+++ b/framework/src/org/apache/cordova/NativeToJsMessageQueue.java
@@ -154,7 +154,7 @@ public class NativeToJsMessageQueue {
 
             StringBuilder sb = new StringBuilder(totalPayloadLen);
             for (int i = 0; i < numMessagesToSend; ++i) {
-                JsMessage message = queue.removeFirst();
+                JsMessage message = queue.remove(0);
                 packMessage(message, sb);
             }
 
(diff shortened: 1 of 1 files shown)


breautek:
Can you explain the rationale and what problem this is fixing?

Both APIs, [removeFirst](https://developer.android.com/reference/java/util/LinkedList#removeFirst()) and [remove(int)](https://developer.android.com/reference/java/util/LinkedList#remove(int)) are supported by the Android SDK since API 1 with no deprecation notices.

breautek:
I'm keeping this closed for now, because I don't really think removing `removeFirst` is a real solution, even though it may solve this particular instance of the problem.

> java.lang.NoSuchMethodError: No static method removeFirst(Ljava/util/List;)Ljava/lang/Object;

`java.util.List.removeFirst` as a static method was introduced in API 21. It's not referenced by Cordova directly, likely within the implementation of `LinkedList.removeFirst` instance method.

The fact that a core API 1 API is using an JDK 21 API suggest that you may have your java source/target compatibility set to 21+... or it may just be enabled due to targeting API 35, which implies a desugaring issue. In my research I couldn't really find a concrete answer.

Your android project should not be using anything higher than JDK 11, which is the default configuration the Android Studio uses when creating new android projects.

There are multiple ways to configure the source/target JDK settings:

- `AndroidJavaSourceCompatibility` / `AndroidJavaTargetCompatibility` - These preferences are undocumented, but would be recommended to remove these preferences if they are present in your `config.xml`.
- Via [gradle arguments](https://cordova.apache.org/docs/en/dev/guide/platforms/android/index.html#configuring-gradle) using `-pJAVA_SOURCE_COMPATIBILITY` / `-pJAVA_TARGET_COMPATIBILITY`
- Via hooks to manually manipulate the build gradle configs.

Not to confuse Java Source & Target Compatibility settings with Android Compile & Target SDK which is something completely different.

If you can, can you confirm that your project is configured to use JDK source/target compatibility 11?

The fact that the codebase is trying to JDK 21 API is troublesome and if it's happening inside `removeFirst`, there may be other APIs also using JDK 21 and thus would run into similar situation, which is why there might be a broader issue here.

PR #1852, opened 2025-10-21
fix: updateStatusBar(int bgColor) to work on non edge-to-edge android…

OS-kepatotorica:
… versions

### Platforms affected

Android

### Motivation and Context

This is a fix for what I believe to be a bug introduced in this PR: https://github.com/apache/cordova-android/pull/1817

With the recent changes to support edge to edge, an android view was added above and below the webview that contains web content. updateStatusBar(int bgColor) successfully sets the color for those views, but will not set the color on a device that doesn't support edge-to-edge, so any device running android 14 and below.

Here is the effect on an android 10 devices with the new changes:
image

And here it is with what currently exists:
image

Notice that since we also automatically set the color of the content, that having a light background color also makes the icons fully disappear as well, since they are trying to maintain contrast with the background, but the background is not being set.

### Description

Use the window.setStatusBarColor() method to correctly set the color of the native status bar where it can exist.

### Testing

I tested setting the status bar color on several versions of android, and changed the value of `AndroidEdgeToEdge` to ensure compatibility with the new preference.

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/SystemBarPlugin.java
+++ b/framework/src/org/apache/cordova/SystemBarPlugin.java
@@ -231,6 +231,10 @@ public class SystemBarPlugin extends CordovaPlugin {
     private void updateStatusBar(int bgColor) {
         Window window = cordova.getActivity().getWindow();
 
+        if (Build.VERSION.SDK_INT <= 35) {
+            window.setStatusBarColor(bgColor);
+        }
+
         View statusBar = getStatusBarView(webView);
         if (statusBar != null) {
             statusBar.setBackgroundColor(bgColor);


erisu:
Can you explain what changes your making that leads to this break?

Using the Cordova-Android's nightly build works as expected. 

In the application theme, the SystemBar is set to transparent. This means the window's native systembar is no longer being used. It was replaced with the custom StatusBar View which is being set with: `statusBar.setBackgroundColor`.

Here are screenshots of current nightly with the following `config.xml`

```xml
<platform name="android">
    <preference name="AndroidEdgeToEdge" value="false" />
    <preference name="BackgroundColor" value="#8E44AD" />
    <preference name="StatusBarBackgroundColor" value="#FFFF00" />
</platform>
```

| SDK 24 | SDK 25 | SDK 27 | SDK 28 | SDK 30 | SDK 34 | SDK 35 | SDK 36 |
| -- | -- | -- | -- | -- | -- | -- | -- |
| SDK24 | SDK25 | SDK27 | SDK28 | SDK30 | SDK34 | SDK35 | SDK36 |

Ignore the black navigational bar on SDK 24 and 25 as this is expected results for these older SDKs.

---

If you are using custom themes, ensure that your status bar is transparent by setting the following:

```xml
<item name="android:statusBarColor">@android:color/transparent</item>
```

If you have the StatusBar plugin installed, I recommend removing it during testing. It may conflict with Cordova's core logic. The plugin has not yet been updated for the upcoming major release of Cordova-Android, and I believe all of its native Android code will be removed.

erisu:
I guess we can add back the deprecated API that you are using in this PR.

When I tested, I saw no difference with or without it in a default Cordova project on SDK 34 and 28. I guess there is no harm to include it.

In another test, I removed the `<item name="android:statusBarColor">@android:color/transparent</item>` to see how it also behaves, with and without the API. It appears to behave the same and looks correct. But, there is a visual visual glitch but this is not in Cordova's scope since Cordova uses transparent by default.

For advance users who decide to make custom themes, they would have to manage it on their end to prevent the glitch. For example, if they set the status bar color to yellow in `config.xml` they will have to update the custom theme to match. Basically they need to manage two locations.

Users who uses default Cordova, they only need to manage via preference flag.

The glitch in the video you will see it going from purple (splash screen) -> black status bar -> yellow status bar.

| Test Case | Result |
| -- | -- |
| Cordova Default (Transparent StatusBar Theme Setting Without Deprecated API) | [TransparentStatusBarThemeWithoutAPI.webm](https://github.com/user-attachments/assets/35cff7e4-ac2f-43c2-805f-d9db3305873d) |
| Cordova Default (Transparent StatusBar Theme Setting With Deprecated API) | [TransparentStatusBarThemeWithAPI.webm](https://github.com/user-attachments/assets/01e8fd5e-1b53-48fe-a6ef-c44ed719b3a2) |
| Removed Transparent StatusBar Theme Setting Without Deprecated API | [NoStatusBarThemeWithOutAPI.webm](https://github.com/user-attachments/assets/1542078a-b255-451b-aeb4-ffb149f018de) | 
| Removed Transparent StatusBar Theme Setting With Deprecated API | [NoStatusBarThemeWithAPI.webm](https://github.com/user-attachments/assets/8a519945-7181-4430-8e53-1eedbd211c49) |

PR #1895, opened 2026-02-24
fix: handle permissions for getUserMedia

### LINKED ISSUE #1888: WebView permission request breaks with no native permission granted
# Bug Report

Requesting permission to use media resources from the WebView (with JS code) should throw an error if the native permission for that resource isn't granted.

For example, if I call:

```js
navigator.mediaDevices.getUserMedia({video: true})
	.then(() => console.log('success!'))
	.catch(() => console.log('error!'));
```

I expected `error!` to be logged if the app doesn't have native permission to use the camera.

The expected behavior happens the first time. But every request after that results in an infinite Promise that never resolves:

Image

Even after native permission is granted, these promises are stuck like that. The only way to fix it is restarting the WebView somehow (for example, by restarting the app)

## Information
I did some research to try and fix it, and it led me to this code:

```java
    @Override
    public void onPermissionRequest(final PermissionRequest request) {
        LOG.d(LOG_TAG, "onPermissionRequest: " + Arrays.toString(request.getResources()));
        request.grant(request.getResources());
    }
```

Here's the link to the commit that added it: https://github.com/apache/cordova-android/commit/eb70f05168dcf3a07d8716c772db3a3c3c3a59f8

And here's the PR that the commit above closes: https://github.com/apache/cordova-android/pull/178

That code will **always** grant the permission to the WebView, without checking if the app even has that permission. If the native permission is not granted, an error is thrown, which would be fine by itself, but the code also breaks, and future requests for permissions coming from the WebView won't even trigger the `onPermissionRequest` callback, which I believe simply stops working after the first error.

This probably would work fine for versions before Android 6, when permissions were granted when the app is being installed (at least I assume so).

There are ways to get around that. For example, by always checking the permission with native  code before using JS code to request permission. But that's not ideal, and leads to more complex code that can break more easily.

Also, I tried overriding the `onPermissionRequest` callback with my own code. This works, but is very unreliable, as Cordova sometimes overrides my override. Still, this helped me be more confident that the issue is indeed related to this code.

### Command or Code
Simply run this in the WebView console **before** native permission is granted:

```js
navigator.mediaDevices.getUserMedia({video: true})
	.then(() => console.log('success!'))
	.catch(() => console.log('error!'));
```

I tested and successfully reproduced this issue on:
- Pixel 7 Pro (Android 13) virtual device, running on Android Studio on Ubuntu
- Galaxy A01 (Android 6) physical device

### Version information
I'm using the [Meteor](https://www.meteor.com/) framework. More specifically, METEOR@2.16.1-beta.0

## Checklist

jcesarmobile:
closes https://github.com/apache/cordova-android/issues/1888

The code is based on Capacitor's implementation.

It uses newer `ActivityResultCallback`/`ActivityResultLauncher` instead of the old `requestPermissions`.

This could later on be extended to request geolocation permissions as at the moment it requires the Geolocation plugin to be installed.

### DIFF
--- a/framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
+++ b/framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
@@ -23,6 +23,9 @@ import java.io.File;
 import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
+
+import android.Manifest;
 import android.app.Activity;
 import android.content.ClipData;
 import android.content.Context;
@@ -45,6 +48,10 @@ import android.webkit.PermissionRequest;
 import android.widget.LinearLayout;
 import android.widget.ProgressBar;
 import android.widget.RelativeLayout;
+
+import androidx.activity.result.ActivityResultCallback;
+import androidx.activity.result.ActivityResultLauncher;
+import androidx.activity.result.contract.ActivityResultContracts;
 import androidx.core.content.FileProvider;
 
 import org.apache.cordova.CordovaDialogsHelper;
@@ -59,6 +66,12 @@ import org.apache.cordova.LOG;
  */
 public class SystemWebChromeClient extends WebChromeClient {
 
+    private interface PermissionListener {
+        void onPermissionSelect(Boolean isGranted);
+    }
+
+    private ActivityResultLauncher permissionLauncher;
+    private PermissionListener permissionListener;
     private static final int FILECHOOSER_RESULTCODE = 5173;
     private static final String LOG_TAG = "SystemWebChromeClient";
     private long MAX_QUOTA = 100 * 1024 * 1024;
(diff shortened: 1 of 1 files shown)


framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
```diff
@@ -325,7 +347,27 @@ private Uri createUriForFile(Context context, File tempFile) throws IOException
...
+        if (!permissionList.isEmpty()) {
+            String[] permissions = permissionList.toArray(new String[0]);
+            permissionListener = (isGranted) -> {
+                if (isGranted) {
+                    request.grant(request.getResources());
+                } else {
+                    request.deny();
+                }
+            };
+            permissionLauncher.launch(permissions);
```
erisu: The following warning appears when `permissionLauncher` is declared as `ActivityResultLauncher` instead of `ActivityResultLauncher<String[]>`.

> Unchecked call to 'launch(I)' as a member of raw type 'androidx.activity.result.ActivityResultLauncher'

erisu (changes requested):
Overall, the code looks good to me.

I left a couple of change requests. Can you confirm if they're OK?

I also tested the following use case, with and without my suggested changes:

1. Create a default Cordova application
2. Add the following to `config.xml`:

```xml
<platform name="android">
    <config-file target="AndroidManifest.xml" parent="/*" xmlns:android="http://schemas.android.com/apk/res/android">
        <uses-permission android:name="android.permission.CAMERA" />
        <uses-feature
            android:name="android.hardware.camera"
            android:required="false" />
    </config-file>
</platform>
```

3. Add the following to `index.html`:

```html
<video id="video" autoplay playsinline></video>
<script>
navigator.mediaDevices.getUserMedia({ video: true }).then(stream => {
    const video = document.getElementById('video');
    video.srcObject = stream;
}).catch(error => {
    console.error('Camera Access Error:', error);
});
</script>
```

- If `uses-permission`/`uses-feature` are not in the AndroidManifest file, nothing happens when launching the app.
- If `uses-permission`/`uses-feature` are defined in the AndroidManifest file, a runtime permission dialog appears when the app launches.

jcesarmobile:
Addressed the comments.

Where you say "If uses-permission/uses-feature are not in the AndroidManifest file, nothing happens when launching the app.", it should be going to the catch now and you should see the `console.error('Camera Access Error:', error);`, which is what the PR fixes, at the moment we are always "grating" the WebView permission, after this change we would grant or reject based on if the permission is present and granted (the method call will succeed), or if not present or present but not granted, it will go to the catch.

erisu (approved):
LGTM

PR #1902, opened 2026-03-12
fix: simplify edge-to-edge preference logic in CordovaActivity and SystemBarPlugin

seamlink-aalves:
Simplify edge-to-edge preference logic

### Platforms affected
Android

### Motivation and Context
The motivation and context was described here: https://github.com/apache/cordova-discuss/issues/114

When having cordova android in Edge To Edge the app does not extend to the full height of the device in version bellow Android 15. This change allows the app to behave the same way it was possible to do with the statusbar plugin preference `StatusBarOverlaysWebView` given that this stopped having any effect in cordova-android 15. 

### Description
This PR removes the limitation for Android version <15 when preference `AndroidEdgeToEdge` is true. 

### Testing
Tested in Android 12, 13 and 16

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -121,8 +121,7 @@ public class CordovaActivity extends AppCompatActivity {
         // need to activate preferences before super.onCreate to avoid "requestFeature() must be called before adding content" exception
         loadConfig();
 
-        canEdgeToEdge = preferences.getBoolean("AndroidEdgeToEdge", false)
-                && Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM;
+        canEdgeToEdge = preferences.getBoolean("AndroidEdgeToEdge", false);
 
         String logLevel = preferences.getString("loglevel", "ERROR");
         LOG.setLogLevel(logLevel);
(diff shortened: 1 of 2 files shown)


erisu:
This appears to have introduce an issue on Android 7 where content is rendered behind the navigation bar, which should not happen, and results in parts of the page not being visible. 

While I understand the idea is to try and re-implement the behavior of `StatusBarOverlaysWebView`, which only affect the statusbar, I think it would have been easier to add back the preference and only update only the top margin.

Android’s edge-to-edge behavior on the other hand would applies to both the status bar and the navigation bar, while E2E was not the original behavior of the previous preference.

GitToTheHub:
> This appears to have introduce an issue on Android 7 where content is rendered behind the navigation bar, which should not happen, and results in parts of the page not being visible.

If this only affects Android 7, maybe we can set the minimum Android to 8?

GitToTheHub:
> While I understand the idea is to try and re-implement the behavior of StatusBarOverlaysWebView, which only affect the statusbar, I think it would have been easier to add back the preference and only update only the top margin.

I think you are right. It's confusing to enable AndroidEdgeToEdge for Android versions pre 15. It should have no effect on Android < 15. Would you make a PR for it?

GitToTheHub:
It would be also possible to hide the statusbar by the new Status Bar JS API `window.statusbar.visible = false;` when fullscreen is required and the Android EdgeToEdge logic is only applied on Android 15 and newer.

GitToTheHub:
I opened PR  #1962, which will revert this PR.

PR #1917, opened 2026-04-21
feat: add setContentStyle method to statusBar for dynamic content style changes

seamlink-aalves:
### Platforms affected
Android

### Motivation and Context
In order to replace statusbar plugin with the native implementation of the statusbar in cordova 15 a function to set the color of the content was missing. This adds support for setting the color of the content.

### Description
A new function `setContentStyle` is added with this PR accepting either `default` or `lightcontent`. It will dinamically set the color of the content in statusbar.

### Checklist

### DIFF
--- a/cordova-js-src/plugin/android/statusbar.js
+++ b/cordova-js-src/plugin/android/statusbar.js
@@ -89,4 +89,27 @@ Object.defineProperty(statusBar, 'setBackgroundColor', {
     }
 });
 
+Object.defineProperty(statusBar, 'setContentStyle', {
+    configurable: false,
+    enumerable: false,
+    writable: false,
+    value: function (value) {
+        if (!['default', 'lightcontent'].includes(value)) {
+            throw new TypeError('Invalid status bar content style: ' + value);
+        }
+        const isDefault = value === 'default';
+        if (window.StatusBar) {
+            // try to let the StatusBar plugin handle it
+            if (isDefault) {
+                window.StatusBar.styleDefault();
+            } else {
+                window.StatusBar.styleLightContent();
+            }
+        } else {
+            statusBarVisible = value;
+            exec(null, null, 'SystemBarPlugin', 'setStatusBarContentStyle', [!!isDefault]);
+        }
+    }
+});
+
 module.exports = statusBar;
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -121,8 +121,7 @@ public class CordovaActivity extends AppCompatActivity {
         // need to activate preferences before super.onCreate to avoid "requestFeature() must be called before adding content" exception
         loadConfig();
 
-        canEdgeToEdge = preferences.getBoolean("AndroidEdgeToEdge", false)
-                && Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM;
+        canEdgeToEdge = preferences.getBoolean("AndroidEdgeToEdge", false);
 
         String logLevel = preferences.getString("loglevel", "ERROR");
         LOG.setLogLevel(logLevel);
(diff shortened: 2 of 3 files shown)


erisu:
Is there a specific reason we need to support this manual logic? The code already handles changing the font color of the statusbar area using luminance-based thresholds to ensure sufficient contrast (e.g., when setting dark background it should automatically select light text). I'm wondering if keeping manual control could reintroduce cases where text becomes hard to read.

seamlink-aalves:
As mentioned in https://github.com/apache/cordova-discuss/issues/114 when having the App in edge-to-edge the background is the webview. 

This requires that the color of the content to be controlled dynamically and changed at any time, for example, when I'm changing from dark to light mode and the background changes color.

erisu:
Could you provide a reproducible repo with minimal changes to a standard Cordova app that demonstrates the color conflict and resulting readability issues requiring manual overrides?

From what I recall, in a default app:

- Switching between light and dark mode automatically updates the text color on resume
- If the background color is hardcoded, the text color remains unchanged
- If the background color is changed via the JS API, the text color should update based on luminance-based thresholds

seamlink-aalves:
In edge-to-edge there is no statusbar background since the webview extends to the full height of the device.

In this case the background color is never set, either by configuration or by JS API. This is why I'm proposing this PR to set the content color.

erisu:
Ah, I understand the issue now.

---

**Issue Use Case Summary:**

In this example, we have two vertically stacked boxes. Each box fills the area of the screen, making the app scrollable.

The first box has a black background, and the second box has a white background.

Initially, we expect the status bar appearance (text) to be white so it is visible against the black background. As the webview scrolls and the white box enters the status bar area, we expect the status bar text to switch to black to maintain readability but in the present condition, the text remains white and is unreadable.

---

I might have an idea for how we could update the appearance dynamically, but I don't think adding it to the Cordova-Android core is the right decision. In fact, I believe we should consider opening a ticket with Google to see if they can properly implement this within Chromium and the AndroidX WebView library.

Although the release announcement states that the status bar plugin will be deprecated, I believe it can still be installed and that the `styleDefault()` and `styleLightContent()` APIs can still be used. I tested this, and it worked.

While the plugin includes initialization code, I didn't notice any issues with it. We could continue using the plugin for handling the status bar appearance as no changes would be required. If there are changes in core, it would be simple adding a JS API to map to the plugin, if the plugin exists. 

And then open a ticket with Google to see if we can get a proper fix.

seamlink-aalves:
In order to move to cordova-android 15 we had to keep the dependency for the statusbar plugin and use the `styleDefault()` and `styleLightContent()` APIs.

Our intention with this PR was to have this functionality in the core platform to be able to remove the plugin dependency.

If this was accepted we would also do a PR for the `cordova-ios` platform which has the same problem/limitation.

dpalou:
+1 to this feature, both in Android and iOS. Recent iOS versions seem to automatically calculate the text color based on the color behind it, but older versions (e.g. iOS 16) don't do that. If we want to have a good experience in all versions it would be nice to be able to control this ourselves. In Android, since the WebView can be updated in older Android versions I guess the impact would be lower, but we don't know when will the Google team implement this so we'll have to keep using the status bar plugin until then.

In our case the problem is not with scroll, it's because our header color (we use Ionic) can change dynamically so we want to update the statusbar text color whenever the header color changes.

GitToTheHub:
I would merge this PR, since the users need this and we recommend to use the core functionality instead of the plugin.

Since the plugin works with a String for the preference `StatusBarStyle`, I would give this value directly to Java and handle it there and not handle it in JavaScript by converting it to a boolean. It's the question if the preference `StatusBarStyle` should be also added back.

dpogue:
The status bar implementation at the platform level was meant to be a very simple one that obeyed theme-color and viewport-fit, with more complicated stuff still being possible with the status bar plugin. That's how it was done on iOS, and that's how I'd hope it works here. We do not want a bunch of custom status bar APIs in the platform that we will have to maintain for years to come because we want to encourage people to not use the built-in web platform features for this (and yes, I know those features are currently a bit lacking on Android so we have to compromise a bit).

My recommendation for people who want to precisely control the text colour and background colour and overlay mechanics rather than accepting common default behaviour is to keep using the status bar plugin.

GitToTheHub:
Ok since two people are against it, it makes sense to close this. The recommendation is to use the status bar plugin for advanced use cases.

PR #1924, opened 2026-05-14
fix: webview loses scrollability and creates empty bottom space

### LINKED ISSUE #1923: cordova-android@15 WebView loses scrollability and creates empty bottom space when soft keyboard appears
# Bug Report

## Problem
When an input field receives focus on a scrollable screen, the Android soft keyboard opens and pushes the entire WebView upward. This leaves an empty, unrendered space at the bottom of the Activity and completely locks the page scroll.

* The WebView should smoothly resize or adjust its viewport when the soft keyboard appears, keeping the bottom content accessible.
* The native web scrolling should remain fully active while the keyboard is visible so users can see other elements or input fields.
* Elements that rely on viewport heights or scroll listeners should render and position themselves correctly.

* The native Activity shifts the WebView up, creating a blank space at the bottom equal to the keyboard's height. 
* As shown in the attached screenshot **(the green line represents the end of the document `<body>`)**, the webview rendering area finishes prematurely.
* The document completely loses scrollability while the keyboard remains open, breaking UI layouts that depend on scrolling adjustments.

Image

Image

https://github.com/user-attachments/assets/5b277982-85f3-4776-8f76-ed50d6c7f267

### Android versions
physical device: 
 - Android 16
 - Android 11
 - Android 10

emulator: 
 - Android 8

### Version information
Cordova - 12.0.1 and 13.0.0 (the same problem in both versions)
Cordova-Android 15.0.0 (the problem occurs only in v15.00)
ionic-cli 6.12.3
ionic 4.6.2

### Plugins in use
"cordova-clipboard": "1.3.0",
"cordova-open-native-settings": "1.5.2",
"cordova-plugin-add-swift-support": "2.0.2",
"cordova-plugin-advanced-http": "3.2.2",
"cordova-plugin-app-version": "0.1.9",
"cordova-plugin-datepicker": "0.9.3",
"cordova-plugin-device": "2.0.3",
"cordova-plugin-file": "8.0.1",
"cordova-plugin-fingerprint-aio": "6.0.1",
"cordova-plugin-firebase-analytics": "8.0.0",
"cordova-plugin-idfa": "3.0.0",
"cordova-plugin-inappbrowser": "6.0.0",
"cordova-plugin-ionic-webview": "5.0.1",
"cordova-plugin-network-information": "2.0.2",
"cordova-plugin-networkinterface": "2.0.0",
"cordova-plugin-screen-orientation": "3.0.4",
"cordova-plugin-secure-storage-echo": "5.0.1",
"cordova-plugin-x-socialsharing": "6.0.4",
"cordova-support-android-plugin": "2.0.4",
"cordova-support-google-services": "1.4.1",

## Checklist

gabriel2mm:
### Platforms affected
 - Android

### Motivation and Context
This change is required to fix UI layout and scrolling bug present in `cordova-android`@15.0.0. When an input field receives focus on a scrollable screen, the Android soft keyboard pushes the entire WebView upward. This leaves an empty, unrendered space at the bottom of the Activity and completely locks the page scroll, which breaks applications relying on viewport heights or scroll adjustments.

Closes https://github.com/apache/cordova-android/issues/1923

### Description
This PR adjusts how the native root layout handles window insets when the soft keyboard (IME) is displayed.

Previously, the layout did not account for the IME insets natively, causing Android to shift the entire Activity upward and cut off the `<body>` boundaries. By adding `WindowInsetsCompat.Type.ime()` to the `ViewCompat.setOnApplyWindowInsetsListener` configuration, the root layout now listens to the keyboard dimensions alongside system bars and display cutouts. This correctly applies the bottom margin dynamically to the `webViewParams`.

after adding: 

Captura de Tela 2026-05-14 às 16 16 07

This behavior is very similar to how cordova-android@14 behaves.

### Testing
 - npm t
 - Build testing

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -18,6 +18,7 @@
 */
 package org.apache.cordova;
 
+import java.io.ObjectInputFilter.Config;
 import java.util.ArrayList;
 import java.util.Locale;
 
@@ -220,7 +221,7 @@ public class CordovaActivity extends AppCompatActivity {
         // Handle Window Insets
         ViewCompat.setOnApplyWindowInsetsListener(rootLayout, (v, insets) -> {
             Insets bars = insets.getInsets(
-                    WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout()
+                    WindowInsetsCompat.Type.ime() | WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout()
             );
 
             boolean isStatusBarVisible = statusBarView.getVisibility() != View.GONE;
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -220,18 +220,25 @@ public class CordovaActivity extends AppCompatActivity {
         // Handle Window Insets
         ViewCompat.setOnApplyWindowInsetsListener(rootLayout, (v, insets) -> {
             Insets bars = insets.getInsets(
-                    WindowInsetsCompat.Type.ime() | WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout()
+                    WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout()
             );
+            Insets ime = insets.getInsets(WindowInsetsCompat.Type.ime());
 
             boolean isStatusBarVisible = statusBarView.getVisibility() != View.GONE;
             int top = isStatusBarVisible && !canEdgeToEdge && !isFullScreen ? bars.top : 0;
-            int bottom = !canEdgeToEdge && !isFullScreen ? bars.bottom : 0;
             int left = !canEdgeToEdge && !isFullScreen ? bars.left : 0;
             int right = !canEdgeToEdge && !isFullScreen ? bars.right : 0;
 
-            FrameLayout.LayoutParams webViewParams = (FrameLayout.LayoutParams) webView.getLayoutParams();
-            webViewParams.setMargins(left, top, right, bottom);
-            webView.setLayoutParams(webViewParams);
+            int bottom = 0;
+            if (!isFullScreen) {
+                bottom = !canEdgeToEdge ? Math.max(bars.bottom, ime.bottom) : ime.bottom;
+            }
+
+            if (webViewParams.leftMargin != left || webViewParams.topMargin != top
+                    || webViewParams.rightMargin != right || webViewParams.bottomMargin != bottom) {
+                webViewParams.setMargins(left, top, right, bottom);
+                webView.setLayoutParams(webViewParams);
+            }
 
             FrameLayout.LayoutParams statusBarParams = new FrameLayout.LayoutParams(
                     ViewGroup.LayoutParams.MATCH_PARENT,
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -222,7 +222,6 @@ public class CordovaActivity extends AppCompatActivity {
             Insets bars = insets.getInsets(
                     WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout()
             );
-            Insets ime = insets.getInsets(WindowInsetsCompat.Type.ime());
 
             boolean isStatusBarVisible = statusBarView.getVisibility() != View.GONE;
             int top = isStatusBarVisible && !canEdgeToEdge && !isFullScreen ? bars.top : 0;
@@ -230,10 +229,16 @@ public class CordovaActivity extends AppCompatActivity {
             int right = !canEdgeToEdge && !isFullScreen ? bars.right : 0;
 
             int bottom = 0;
+            Insets imeInsets = insets.getInsets(WindowInsetsCompat.Type.ime());
+            // When in fullscreen mode, we ignore bottom system insets (like the navigation bar)
+            // to allow the WebView to span the entire screen and avoid being pushed up.
             if (!isFullScreen) {
-                bottom = !canEdgeToEdge ? Math.max(bars.bottom, ime.bottom) : ime.bottom;
+                bottom = canEdgeToEdge ? imeInsets.bottom : Math.max(bars.bottom, imeInsets.bottom);
             }
 
+            FrameLayout.LayoutParams webViewParams = (FrameLayout.LayoutParams) webView.getLayoutParams();
+            // Only update layout margins if the values have actually changed.
+            // This prevents redundant layout passes and potential infinite layout loops
             if (webViewParams.leftMargin != left || webViewParams.topMargin != top
                     || webViewParams.rightMargin != right || webViewParams.bottomMargin != bottom) {
                 webViewParams.setMargins(left, top, right, bottom);
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -222,18 +222,20 @@ public class CordovaActivity extends AppCompatActivity {
             Insets bars = insets.getInsets(
                     WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout()
             );
-            Insets ime = insets.getInsets(WindowInsetsCompat.Type.ime());
 
             boolean isStatusBarVisible = statusBarView.getVisibility() != View.GONE;
             int top = isStatusBarVisible && !canEdgeToEdge && !isFullScreen ? bars.top : 0;
             int left = !canEdgeToEdge && !isFullScreen ? bars.left : 0;
             int right = !canEdgeToEdge && !isFullScreen ? bars.right : 0;
 
-            int bottom = 0;
-            if (!isFullScreen) {
-                bottom = !canEdgeToEdge ? Math.max(bars.bottom, ime.bottom) : ime.bottom;
-            }
+            Insets imeInsets = insets.getInsets(WindowInsetsCompat.Type.ime());
+            // When in fullscreen mode, we ignore bottom system insets (like the navigation bar)
+            // to allow the WebView to span the entire screen and avoid being pushed up.
+            int bottom = isFullScreen ? 0 : canEdgeToEdge ? imeInsets.bottom : Math.max(bars.bottom, imeInsets.bottom);
 
+            FrameLayout.LayoutParams webViewParams = (FrameLayout.LayoutParams) webView.getLayoutParams();
+            // Only update layout margins if the values have actually changed.
+            // This prevents redundant layout passes and potential infinite layout loops
             if (webViewParams.leftMargin != left || webViewParams.topMargin != top
                     || webViewParams.rightMargin != right || webViewParams.bottomMargin != bottom) {
                 webViewParams.setMargins(left, top, right, bottom);
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -18,7 +18,6 @@
 */
 package org.apache.cordova;
 
-import java.io.ObjectInputFilter.Config;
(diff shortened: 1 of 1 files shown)


gabriel2mm (commented):
Added support to EdgeToEdge enabled

framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -222,16 +222,24 @@ protected void createViews() {
             Insets bars = insets.getInsets(
                     WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout()
             );
+            Insets ime = insets.getInsets(WindowInsetsCompat.Type.ime());
```
GitToTheHub: Put this variable after `int bottom = 0;` and name it `imeInsets`

framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -222,16 +222,24 @@ protected void createViews() {
...
 
             boolean isStatusBarVisible = statusBarView.getVisibility() != View.GONE;
             int top = isStatusBarVisible && !canEdgeToEdge && !isFullScreen ? bars.top : 0;
-            int bottom = !canEdgeToEdge && !isFullScreen ? bars.bottom : 0;
             int left = !canEdgeToEdge && !isFullScreen ? bars.left : 0;
             int right = !canEdgeToEdge && !isFullScreen ? bars.right : 0;
 
+            int bottom = 0;
+            if (!isFullScreen) {
```
GitToTheHub: What does this code? Add documentation why `if (!isFullScreen)` is set here

framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -222,16 +222,24 @@ protected void createViews() {
...
+                bottom = !canEdgeToEdge ? Math.max(bars.bottom, ime.bottom) : ime.bottom;
```
GitToTheHub: You can revert the condition and don't need the `!` at the beginning and it's easier to read:

```java
bottom = canEdgeToEdge ? ime.bottom : Math.max(bars.bottom, ime.bottom);
```

framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -222,16 +222,24 @@ protected void createViews() {
...
+            }
+
             FrameLayout.LayoutParams webViewParams = (FrameLayout.LayoutParams) webView.getLayoutParams();
-            webViewParams.setMargins(left, top, right, bottom);
-            webView.setLayoutParams(webViewParams);
+            if (webViewParams.leftMargin != left || webViewParams.topMargin != top
```
GitToTheHub, line 239: Why are these conditions here? Please document it.

GitToTheHub (changes requested):
I added some notes to your changes.

framework/src/org/apache/cordova/CordovaActivity.java
GitToTheHub: You can simply the condition:

```java
int bottom = isFullScreen ? 0 : canEdgeToEdge ? ime.bottom : Math.max(bars.bottom, ime.bottom);
```

GitToTheHub (changes requested):
Some condition rewrite for the bottom variable

framework/src/org/apache/cordova/CordovaActivity.java
erisu: My recommendation is to avoid nested ternary operators if possible. They can be hard to read. The outer if condition may be sufficient.

I found this reference that talked about avoiding nested ternary operators: https://docs.openrewrite.org/recipes/staticanalysis/ternaryoperatorsshouldnotbenested

gabriel2mm: I believe nested ternaries can really hurt code readability. I suggest keeping it as a separate if block, just like @erisu mentioned.

framework/src/org/apache/cordova/CordovaActivity.java
gabriel2mm: Updated as suggested.

framework/src/org/apache/cordova/CordovaActivity.java
gabriel2mm: Updated as suggested.

framework/src/org/apache/cordova/CordovaActivity.java
gabriel2mm: I've added the inline documentation. Overall, the fullscreen option shouldn't consider the system insets when applying margins, so the WebView doesn't get incorrectly pushed up.

framework/src/org/apache/cordova/CordovaActivity.java
gabriel2mm: I've added the inline documentation.

I believe we should avoid updating the WebView if the margin parameters haven't changed. Re-applying the exact same parameters triggers redundant layout passes, which can cause unwanted behaviors like infinite layout loops.

framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -225,13 +225,25 @@ protected void createViews() {
...
+                bottom = canEdgeToEdge ? imeInsets.bottom : Math.max(bars.bottom, imeInsets.bottom);
+            }
+
             FrameLayout.LayoutParams webViewParams = (FrameLayout.LayoutParams) webView.getLayoutParams();
-            webViewParams.setMargins(left, top, right, bottom);
-            webView.setLayoutParams(webViewParams);
+            // Only update layout margins if the values have actually changed.
+            // This prevents redundant layout passes and potential infinite layout loops
+            if (webViewParams.leftMargin != left || webViewParams.topMargin != top
+                    || webViewParams.rightMargin != right || webViewParams.bottomMargin != bottom) {
```
GitToTheHub, line 240: This code will only be executed when the `OnApplyWindowInsetsListener` is fired. Is this condition really necessary to guard:

```java
webViewParams.setMargins(left, top, right, bottom);
webView.setLayoutParams(webViewParams);
```

or would this re-trigger the `OnApplyWindowInsetsListener`?

GitToTheHub (changes requested):

framework/src/org/apache/cordova/CordovaActivity.java
GitToTheHub: Ok I never had know, that this should be avoided. For me it depends always on the code itself if it makes sense to use a nested ternary operator or not. It should always improve readability. For e.g. this nested ternary operator as shown as an example on https://docs.openrewrite.org/recipes/staticanalysis/ternaryoperatorsshouldnotbenested makes the code really not easy to read:

```java
"a".equals(a) ? "a" : "b".equals(a) ? "b" : "nope";
```

but in this context, it makes it easier to read:

```java
Insets imeInsets = insets.getInsets(WindowInsetsCompat.Type.ime());
// When in fullscreen mode, we ignore bottom system insets (like the navigation bar)
// to allow the WebView to span the entire screen and avoid being pushed up.
int bottom = isFullScreen ? 0 : canEdgeToEdge ? imeInsets.bottom : Math.max(bars.bottom, imeInsets.bottom);
```

instead of 

```java
int bottom = 0;
Insets imeInsets = insets.getInsets(WindowInsetsCompat.Type.ime());
// When in fullscreen mode, we ignore bottom system insets (like the navigation bar)
// to allow the WebView to span the entire screen and avoid being pushed up.
if (!isFullScreen) {
    bottom = canEdgeToEdge ? imeInsets.bottom : Math.max(bars.bottom, imeInsets.bottom);
}
```

Also the bottom variable is a one-liner and could be grouped together with the other assignments.

framework/src/org/apache/cordova/CordovaActivity.java
gabriel2mm: I believe this validation is highly necessary to prevent redundant updates. During debugging, I observed that opening the soft keyboard alone triggers this event multiple times with the exact same parameters. Reapplying setLayoutParams when the margins haven't changed forces an unnecessary re-layout on the WebView, which negatively impacts performance and could cause visual jitter.

I also believe that the event can be triggered multiple times during system animations.


Example debug code:

https://github.com/user-attachments/assets/95b9f310-c910-4821-bd1f-334ba3ed304c

framework/src/org/apache/cordova/CordovaActivity.java
gabriel2mm: I believe this might be a matter of personal preferences or coding style. However, keeping the code on a single line is perfectly fine too. I will follow your suggestion and push a commit to address this point

gabriel2mm: Updated as suggested.

framework/src/org/apache/cordova/CordovaActivity.java
```diff
@@ -225,13 +225,22 @@ protected void createViews() {
...
             boolean isStatusBarVisible = statusBarView.getVisibility() != View.GONE;
             int top = isStatusBarVisible && !canEdgeToEdge && !isFullScreen ? bars.top : 0;
-            int bottom = !canEdgeToEdge && !isFullScreen ? bars.bottom : 0;
             int left = !canEdgeToEdge && !isFullScreen ? bars.left : 0;
             int right = !canEdgeToEdge && !isFullScreen ? bars.right : 0;
 
+            Insets imeInsets = insets.getInsets(WindowInsetsCompat.Type.ime());
+            // When in fullscreen mode, we ignore bottom system insets (like the navigation bar)
+            // to allow the WebView to span the entire screen and avoid being pushed up.
+            int bottom = isFullScreen ?  0 : canEdgeToEdge ? imeInsets.bottom : Math.max(bars.bottom, imeInsets.bottom);
```
GitToTheHub: Remove one doubled whitespace in `?  0`

GitToTheHub (changes requested):

framework/src/org/apache/cordova/CordovaActivity.java
gabriel2mm: Updated as suggested.

framework/src/org/apache/cordova/CordovaActivity.java
GitToTheHub: Ok. In your video the method is only called two times when the keyboard appears or dismisses, which is not so much, but the values were the same for left, top, right, bottom. So I think this is ok to keep this.

GitToTheHub:
@gabriel2mm Thanks for the changes.

@erisu Are you ok with using nested ternary operators here for calculating the bottom variable?:

```java
int bottom = isFullScreen ? 0 : canEdgeToEdge ? imeInsets.bottom : Math.max(bars.bottom, imeInsets.bottom);
```

Or do you prefer it with an outer condition?:

```java
int bottom = 0;
if (!isFullScreen) {
    bottom = canEdgeToEdge ? imeInsets.bottom : Math.max(bars.bottom, imeInsets.bottom);
}
```

GitToTheHub (approved):

erisu:
I believe this might have introduced a small visual glitch during keyboard animations where the app background briefly becomes visible. This is caused by the WebView being resized continuously during the IME animation and is especially noticeable when the app background contrasts with the page content.

E.g. set the web page's background to something light and set the app's background, with config preference, to something dark.

I just noticed this while I was testing something else. I’ll open an issue ticket so it can be further investigated.

GitToTheHub:
Ok good you noticed that 🙂I hope we can find a good solution for that.

gabriel2mm:
> I believe this might have introduced a small visual glitch during keyboard animations where the app background briefly becomes visible. This is caused by the WebView being resized continuously during the IME animation and is especially noticeable when the app background contrasts with the page content.
> 
> E.g. set the web page's background to something light and set the app's background, with config preference, to something dark.
> 
> I just noticed this while I was testing something else. I’ll open an issue ticket so it can be further investigated.

Interesting observation, and thanks for reporting it.

For the record, from my perspective and based on my testing, this PR did not introduce this behavior. I believe the WebView fix simply made it more noticeable. However, I previously reported similar behaviors on cordova-android@14 which, in my opinion, are related to Android 16's edge-to-edge behavior and may require a new approach for handling the bottom system bar.

Link to comment: https://github.com/apache/cordova-android/issues/1946#issuecomment-4554452055

PR #1925, opened 2026-05-15
chore: cleanup PermissionHelper

GitToTheHub:
### Platforms affected

- Android

### Motivation and Context

- AndroidStudio complained about methods not found linked in JavaDoc for the class CordovaInterface: requestPermission(), requestPermissions() and hasPermission(). The parameter types were missing and were added.
- Removing unused imports
- Added missing `@Deprecated` Java annotations

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/PermissionHelper.java
+++ b/framework/src/org/apache/cordova/PermissionHelper.java
@@ -18,12 +18,6 @@
 */
 package org.apache.cordova;
 
-import java.util.Arrays;
-
-import org.json.JSONException;
-
-import android.content.pm.PackageManager;
-
 /**
  * This class is permission helper class when compiling against older versions of cordova-android pre 5.0.0
  * and provides reflective methods for permission requesting and checking so that plugins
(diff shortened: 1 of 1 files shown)


erisu:
I'm not sure whether this class should be deprecated and would like to hear others' opinions.

GitHub search isn't always thorough, but it appears that 846 files are importing this class. Also, I believe the 846 files include forked repos, but there are some unique plugin repositories in the list.

https://github.com/search?q=import+org.apache.cordova.PermissionHelper&type=code

Deprecation would only warn, not break builds, but I'm concerned about how many of these plugins are unmaintained and whether removing the class entirely in the future could cause issues.

GitToTheHub:
I looked through the search results and `cordova-plugin-camera` is one of the first search results which uses the `PermissionHelper`. The next one is `cordova-plugin-audioinput` which is actively maintained. The fix would be straightforward. The plugin calls instead of `PermissionHelper.hasPermission` just `cordova.hasPermission`. The same for `PermissionHelper.requestPermission` it would call `cordova.requestPermission` and for `PermissionHelper.requestPermissions` it would call `cordova.requestPermissions`. The right way would be to call the methods directly on `CordovaInterface` of the `CordovaPlugin` instead calling `PermissionHelper` which forwards the calls to it. How would you do the transitioning? Would it be the right approach to leave everything like it is, though it is not necessary? The minimum would be to inform plugin authors to change the way of calling it. A deprecation notice would be a first good way.

breautek (approved):
I can agree with properly adding the deprecation annotation but I don't think this is something we can realistically ever remove.

But we can at least annotate properly so that future plugin development will hopefully steer away from using the PermissionHelper API.

erisu (approved):
These changes look good to me. They remove the build warnings about the missing deprecated annotation.

Since there maybe plugins that rely on it, all we can do is hope that plugin developers naturally migrate away from this helper class over time. 
To avoid breaking unmaintained or private plugins that we don't have visibility into, I think this class should remain.

PR #1939, opened 2026-05-21
fix(SystemBarPlugin): remove sentinel value `INVALID_COLOR`, use `null` instead 

### LINKED ISSUE #1938: Status bar white color (#FFFFFFFF) is treated as invalid color
# Bug Report

When passing a valid color such as #FFFFFFFF to the status bar color configuration, the color should be parsed correctly and applied as white.

#FFFFFFFF is being treated as an invalid color.
The issue appears to come from the following implementation:

Image

`static final int INVALID_COLOR = -1;`

Because Android's Color.parseColor("#FFFFFFFF") returns -1 (0xFFFFFFFF as signed int), a valid white color collides with the sentinel value used for invalid colors.

As a result, fully opaque white cannot be used.

**This becomes especially noticeable when a custom color is applied first and then changed back to white, since the white color is ignored.**

example: 

Image

I would like to propose a small fix to make this case work correctly.

## Information

Android color integers are signed 32-bit integers.
Color.parseColor("#FFFFFFFF") correctly returns -1, which is also equal to Color.WHITE.

The issue is not in Color.parseColor(), but in using -1 as INVALID_COLOR.

### Command or Code
window.statusbar.setBackgroundColor('#FFFFFFFF');

Platform: Android
Device: Any
I tested this on both a physical Android 16 device and an emulator.

### Version information
cordova 13.0.0
cordova-android 15.0.0
ionic 4

## Checklist

gabriel2mm:
### Platforms affected
- Android

### Motivation and Context

This change fixes an issue where the color #FFFFFFFF (fully opaque white) is treated as an invalid color when used for the status bar background.

The problem happens because Color.parseColor("#FFFFFFFF") correctly returns -1, which collides with the current sentinel value used for INVALID_COLOR.

As a result, valid white colors cannot be applied to the status bar.

fixes: https://github.com/apache/cordova-android/issues/1938

### Description

Updated the invalid color handling to avoid collisions with valid ARGB color values returned by Color.parseColor().

This allows colors such as #FFFFFFFF to be parsed and applied correctly.

{A58497E7-1000-435C-AB29-5667A8697CF0}

Example of an invalid sentinel collision:
window.statusbar.setBackgroundColor('#80000000');

This returns a value equal to Integer.MIN_VALUE.

### Testing

- npm t
- Build testing

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/SystemBarPlugin.java
+++ b/framework/src/org/apache/cordova/SystemBarPlugin.java
@@ -42,7 +42,7 @@ import org.json.JSONException;
 public class SystemBarPlugin extends CordovaPlugin {
     static final String PLUGIN_NAME = "SystemBarPlugin";
 
-    static final int INVALID_COLOR = -1;
+    static final int INVALID_COLOR = Integer.MIN_VALUE;
 
     // Internal variables
     private Context context;
--- a/framework/src/org/apache/cordova/SystemBarPlugin.java
+++ b/framework/src/org/apache/cordova/SystemBarPlugin.java
@@ -39,15 +39,15 @@ import androidx.core.view.WindowInsetsControllerCompat;
 import org.json.JSONArray;
 import org.json.JSONException;
 
+import java.util.Objects;
+
 public class SystemBarPlugin extends CordovaPlugin {
     static final String PLUGIN_NAME = "SystemBarPlugin";
-
-    static final int INVALID_COLOR = Integer.MIN_VALUE;
 
     // Internal variables
     private Context context;
     private Resources resources;
-    private int overrideStatusBarBackgroundColor = INVALID_COLOR;
+    private Integer overrideStatusBarBackgroundColor = null;
 
     private boolean canEdgeToEdge = false;
 
@@ -129,13 +129,11 @@ public class SystemBarPlugin extends CordovaPlugin {
             int r = argbVals.getInt(1);
             int g = argbVals.getInt(2);
             int b = argbVals.getInt(3);
-            String hexColor = String.format("#%02X%02X%02X%02X", a, r, g, b);
+            overrideStatusBarBackgroundColor = parseColorFromString(String.format("#%02X%02X%02X%02X", a, r, g, b));
 
-            int parsedColor = parseColorFromString(hexColor);
-            if (parsedColor == INVALID_COLOR) return;
-
-            overrideStatusBarBackgroundColor = parsedColor;
-            updateStatusBar(overrideStatusBarBackgroundColor);
+            if (overrideStatusBarBackgroundColor != null) {
+                updateStatusBar(overrideStatusBarBackgroundColor);
+            }
         } catch (JSONException e) {
             // Silently skip
         }
@@ -145,26 +143,26 @@ public class SystemBarPlugin extends CordovaPlugin {
      * Attempt to update all system bars (status, navigation and gesture bars) in various points
      * of the apps life cycle.
      * For example:
-     *  1. Device configurations between (E.g. between dark and light mode)
-     *  2. User resumes the app
-     *  3. App transitions from SplashScreen Theme to App's Theme
+     * 1. Device configurations between (E.g. between dark and light mode)
+     * 2. User resumes the app
+     * 3. App transitions from SplashScreen Theme to App's Theme
      */
     private void updateSystemBars() {
         // Update Root View Background Color
-        int rootViewBackgroundColor = getPreferenceBackgroundColor();
-        if (rootViewBackgroundColor == INVALID_COLOR) {
+        Integer rootViewBackgroundColor = getPreferenceBackgroundColor();
+        if (rootViewBackgroundColor == null) {
             rootViewBackgroundColor = canEdgeToEdge ? Color.TRANSPARENT : getUiModeColor();
         }
         updateRootView(rootViewBackgroundColor);
 
         // Update StatusBar Background Color
-        int statusBarBackgroundColor;
-        if (overrideStatusBarBackgroundColor != INVALID_COLOR) {
+        Integer statusBarBackgroundColor;
+        if (overrideStatusBarBackgroundColor != null) {
             statusBarBackgroundColor = overrideStatusBarBackgroundColor;
         } else if (preferences.contains("StatusBarBackgroundColor")) {
             statusBarBackgroundColor = getPreferenceStatusBarBackgroundColor();
-        } else if(preferences.contains("BackgroundColor")){
-            statusBarBackgroundColor =  rootViewBackgroundColor;
+        } else if (preferences.contains("BackgroundColor")) {
+            statusBarBackgroundColor = rootViewBackgroundColor;
         } else {
             statusBarBackgroundColor = canEdgeToEdge ? Color.TRANSPARENT : getUiModeColor();
         }
@@ -261,34 +259,31 @@ public class SystemBarPlugin extends CordovaPlugin {
     }
 
     /**
-     * Returns the StatusBarBackgroundColor preference value.
-     * If the value is missing or fails to parse, it will attempt to try to guess the background
-     * color by extracting from the apps R.color.cdv_background_color or determine from the uiModes.
-     * If all fails, the color normally used in light mode is returned.
+     * Returns the StatusBarBackgroundColor preference value or {@link #getUiModeColor()} as fallback.
      *
-     * @return int
+     * @return Integer
      */
-    private int getPreferenceStatusBarBackgroundColor() {
+    private Integer getPreferenceStatusBarBackgroundColor() {
         String colorString = preferences.getString("StatusBarBackgroundColor", null);
-
-        int parsedColor = parseColorFromString(colorString);
-        if (parsedColor != INVALID_COLOR) return parsedColor;
-
-        return getUiModeColor(); // fallback
+        return Objects.requireNonNullElse(parseColorFromString(colorString), getUiModeColor());
     }
 
     /**
      * Returns the BackgroundColor preference value.
-     * If missing or fails to decode, it will return INVALID_COLOR (-1).
+     * If the value is missing or fails to decode, null is returned.
      *
-     * @return int
+     * @return Integer|null
      */
-    private int getPreferenceBackgroundColor() {
+    private Integer getPreferenceBackgroundColor() {
+        if (!preferences.contains("BackgroundColor")) {
+            return null;
+        }
+
         try {
-            return preferences.getInteger("BackgroundColor", INVALID_COLOR);
+            return preferences.getInteger("BackgroundColor", 0);
         } catch (NumberFormatException e) {
             LOG.e(PLUGIN_NAME, "Invalid background color argument. Example valid string: '0x00000000'");
-            return INVALID_COLOR;
+            return null;
         }
     }
 
@@ -354,20 +349,20 @@ public class SystemBarPlugin extends CordovaPlugin {
     }
 
     /**
-     * Parse color string that would be provided by app developers.
-     * If the color string is empty or unable to parse, it will return INVALID_COLOR (-1).
+     * Parses a color string provided by app developers.
+     * If the color string is empty or unable to parse, null is returned.
      *
      * @param colorPref hex string value, #AARRGGBB or #RRGGBB
-     * @return int
+     * @return Integer|null
      */
-    private int parseColorFromString(final String colorPref) {
-        if (colorPref.isEmpty()) return INVALID_COLOR;
+    private Integer parseColorFromString(final String colorPref) {
+        if (colorPref == null || colorPref.isEmpty()) return null;
 
         try {
             return Color.parseColor(colorPref);
         } catch (IllegalArgumentException ignore) {
             LOG.e(PLUGIN_NAME, "Invalid color hex code. Valid format: #RRGGBB or #AARRGGBB");
-            return INVALID_COLOR;
+            return null;
         }
     }
 }


GitToTheHub:
Hi good catch. Since you already noted that `#80000000` resolves to `Integer.MIN_VALUE` the change would be to not use `INVALID_COLOR` at all. `INVALID_COLOR` in most cases is set, when a string could not parsed by `Color.parseColor`, which then throws an `IllegalArgumentException`. So the right approach would be to react on the exception directly instead of setting `INVALID_COLOR` and use this.

These methods set `INVALID_COLOR`:
- `parseColorFromString` will return `INVALID_COLOR` if the parameter string is empty or `Color.parse` throws `IllegalArgumentException`. `parseColorFromString` is called by:
  - `setStatusBarBackgroundColor` which is called from the corresponding JavaScript function
  - `getPreferenceStatusBarBackgroundColor` for reading the preference `StatusBarBackgroundColor`
- `getPreferenceBackgroundColor` returns `INVALID_COLOR` if the preference `BackgroundColor` could not be parsed as int

The variable `overrideStatusBarBackgroundColor` is an int and can also contain `INVALID_COLOR`. It is set by `setStatusBarBackgroundColor` which is called by the corresponding JavaScript function. It is only used in `updateSystemBars`. Instead of an `int`, this could contain the string given by `setStatusBarBackgroundColor` and parse it every time in `updateSystemBars` to catch the exception.

gabriel2mm:
I agree with the points raised. I’ve pushed a new commit for review where I removed the INVALID_COLOR. I believe INVALID_COLOR no longer makes sense in this functionality and could be handled differently. @GitToTheHub could you please take another look?

dpogue:
It might be safe to use 0 (`0x00000000`, or `#00000000` which is transparent: `rgba(0, 0, 0, 0)`) as a sentinel value. That's what's being proposed as a "null" value for the Android WebView implementation of theme-color handling.

gabriel2mm:
> It might be safe to use 0 (`0x00000000`, or `#00000000` which is transparent: `rgba(0, 0, 0, 0)`) as a sentinel value. That's what's being proposed as a "null" value for the Android WebView implementation of theme-color handling.

I had written this to you before, but I changed my mind afterward, and I’ll try to explain why:

Using 0 does not work simply because, on some Android devices, 0x00000000 represents the initial state of the status bar. See:

{20738F37-5FB1-4E83-978E-1CA23B20DA16}

When attempting to change the color and then revert it back to the initial state, we get blocked due to the sentinel value being 0.

{B802E94D-6B12-4BCE-ABA5-8350F19BE332}

I still have the feeling that handling it the way I proposed would be a better approach.

gabriel2mm:
If possible, I’d appreciate it if someone from the community could review this PR so we can determine whether it’s viable to merge it. 
I’m committed to supporting this version of Cordova with fixes and improvements.

breautek (approved):
Tests passes and I think this is definitely an improvement.

We can't really reserve an `int` for INVALID_COLOR because any value within the `int` range is a valid colour, assuming you have a 4x 8bit channels (that's you're RGBA).

If wanted to reserve a number value for invalid color and not have it conflict, we'd have to pick a value outside of an int32 range, which means using `Long` value.

Or we can do what this PR does and use `String` / `null` which I think is perfectly reasonable as well provided there is null guarding. Looks like all APIs being modified here are private methods, so it should be relatively safe change.

Assuming there is no other objections, ping me in a couple days and I'll merge.

framework/src/org/apache/cordova/SystemBarPlugin.java
```diff
@@ -129,13 +127,10 @@ private void setStatusBarBackgroundColor(JSONArray argbVals) {
             int r = argbVals.getInt(1);
             int g = argbVals.getInt(2);
             int b = argbVals.getInt(3);
-            String hexColor = String.format("#%02X%02X%02X%02X", a, r, g, b);
 
-            int parsedColor = parseColorFromString(hexColor);
-            if (parsedColor == INVALID_COLOR) return;
+            overrideStatusBarBackgroundColor = String.format("#%02X%02X%02X%02X", a, r, g, b);
```
GitToTheHub: Previously `overrideStatusBarBackgroundColor` did get the parsed color here.
I would change `overrideStatusBarBackgroundColor` to Integer and set the parsed color directly here, e.g.:

```java
overrideStatusBarBackgroundColor = parseColorFromString(String.format("#%02X%02X%02X%02X", a, r, g, b));
```

So it's the same behaviour as before and is null if the parsing failed

framework/src/org/apache/cordova/SystemBarPlugin.java
```diff
@@ -129,13 +127,10 @@ private void setStatusBarBackgroundColor(JSONArray argbVals) {
...
 
-            overrideStatusBarBackgroundColor = parsedColor;
-            updateStatusBar(overrideStatusBarBackgroundColor);
+            updateSystemBars();
```
GitToTheHub: When you change `overrideStatusBarBackgroundColor` to an `Integer`, you can call `updateStatusBar(overrideStatusBarBackgroundColor)` again instead of `updateSystemBars()`. Or is there any other reason to call `updateSystemBars()` instead of `updateStatusBar()`?

framework/src/org/apache/cordova/SystemBarPlugin.java
```diff
@@ -145,30 +140,34 @@ private void setStatusBarBackgroundColor(JSONArray argbVals) {
...
+        } else if (preferences.contains("BackgroundColor")) {
+            statusBarBackgroundColor = rootViewBackgroundColor;
         } else {
             statusBarBackgroundColor = canEdgeToEdge ? Color.TRANSPARENT : getUiModeColor();
         }
 
+        if (statusBarBackgroundColor == null) {
+            statusBarBackgroundColor = canEdgeToEdge ? Color.TRANSPARENT : getUiModeColor();
+        }
+
```
GitToTheHub: When `overrideStatusBarBackgroundColor` is an `Integer` this can be removed.

framework/src/org/apache/cordova/SystemBarPlugin.java
```diff
@@ -262,33 +261,32 @@ private static boolean isColorLight(int color) {
...
+     * @return Integer|null
      */
-    private int getPreferenceStatusBarBackgroundColor() {
+    private Integer getPreferenceStatusBarBackgroundColor() {
         String colorString = preferences.getString("StatusBarBackgroundColor", null);
 
-        int parsedColor = parseColorFromString(colorString);
-        if (parsedColor != INVALID_COLOR) return parsedColor;
-
-        return getUiModeColor(); // fallback
```
GitToTheHub, line 277: Why did you removed the fallback `getUiModeColor()`?

GitToTheHub (changes requested):

framework/src/org/apache/cordova/SystemBarPlugin.java
gabriel2mm: Thanks a lot for the suggestions. They were really helpful — I updated and retested the code, and everything seems to be working fine from what I can see.

Updated as suggested.

framework/src/org/apache/cordova/SystemBarPlugin.java
gabriel2mm: I think it was a misunderstanding — I reviewed it and added the method back to the code.

framework/src/org/apache/cordova/SystemBarPlugin.java
gabriel2mm: Updated as suggested.

framework/src/org/apache/cordova/SystemBarPlugin.java
gabriel2mm: Updated as suggested.

framework/src/org/apache/cordova/SystemBarPlugin.java
```diff
@@ -129,12 +127,9 @@ private void setStatusBarBackgroundColor(JSONArray argbVals) {
...
             int g = argbVals.getInt(2);
             int b = argbVals.getInt(3);
-            String hexColor = String.format("#%02X%02X%02X%02X", a, r, g, b);
 
-            int parsedColor = parseColorFromString(hexColor);
-            if (parsedColor == INVALID_COLOR) return;
+            overrideStatusBarBackgroundColor = parseColorFromString(String.format("#%02X%02X%02X%02X", a, r, g, b));
 
-            overrideStatusBarBackgroundColor = parsedColor;
             updateStatusBar(overrideStatusBarBackgroundColor);
```
GitToTheHub: We must check `overrideStatusBarBackgroundColor` if it's not null before calling `updateStatusBar`, otherwise this would throw a `NullPointerException` because `updateStatusBar` takes a primitive `int` which cannot be null.

framework/src/org/apache/cordova/SystemBarPlugin.java
```diff
@@ -262,33 +257,33 @@ private static boolean isColorLight(int color) {
...
-    private int getPreferenceStatusBarBackgroundColor() {
+    private Integer getPreferenceStatusBarBackgroundColor() {
         String colorString = preferences.getString("StatusBarBackgroundColor", null);
 
-        int parsedColor = parseColorFromString(colorString);
-        if (parsedColor != INVALID_COLOR) return parsedColor;
-
-        return getUiModeColor(); // fallback
+        Integer parsedColor = parseColorFromString(colorString);
+        return parsedColor != null ? parsedColor : getUiModeColor();//getUiModeColor is used as a fallback.
```
GitToTheHub: This can be simplified to:

```java
return Objects.requireNonNullElse(parseColorFromString(colorString), getUiModeColor());
```

but needs to import `java.util.Objects`:

```java
import java.util.Objects;
```

framework/src/org/apache/cordova/SystemBarPlugin.java
```diff
@@ -262,33 +257,33 @@ private static boolean isColorLight(int color) {
...
-     * color by extracting from the apps R.color.cdv_background_color or determine from the uiModes.
-     * If all fails, the color normally used in light mode is returned.
+     * If the value is missing or fails to parse, null is returned.
      *
-     * @return int
+     * @return Integer|null
      */
-    private int getPreferenceStatusBarBackgroundColor() {
+    private Integer getPreferenceStatusBarBackgroundColor() {
         String colorString = preferences.getString("StatusBarBackgroundColor", null);
```
GitToTheHub: The whitespace can be removed

framework/src/org/apache/cordova/SystemBarPlugin.java
```diff
@@ -262,33 +257,33 @@ private static boolean isColorLight(int color) {
 
     /**
      * Returns the StatusBarBackgroundColor preference value.
-     * If the value is missing or fails to parse, it will attempt to try to guess the background
-     * color by extracting from the apps R.color.cdv_background_color or determine from the uiModes.
-     * If all fails, the color normally used in light mode is returned.
+     * If the value is missing or fails to parse, null is returned.
```
GitToTheHub: This comment has to be updated:

```javadoc
/**
 * Returns the StatusBarBackgroundColor preference value or {@link #getUiModeColor()} as fallback.
 *
 * @return Integer
 */
```

framework/src/org/apache/cordova/SystemBarPlugin.java
```diff
@@ -129,12 +127,9 @@ private void setStatusBarBackgroundColor(JSONArray argbVals) {
             int r = argbVals.getInt(1);
             int g = argbVals.getInt(2);
             int b = argbVals.getInt(3);
-            String hexColor = String.format("#%02X%02X%02X%02X", a, r, g, b);
```
GitToTheHub: You can remove the whitespace

framework/src/org/apache/cordova/SystemBarPlugin.java
```diff
@@ -129,12 +127,9 @@ private void setStatusBarBackgroundColor(JSONArray argbVals) {
...
 
-            int parsedColor = parseColorFromString(hexColor);
-            if (parsedColor == INVALID_COLOR) return;
+            overrideStatusBarBackgroundColor = parseColorFromString(String.format("#%02X%02X%02X%02X", a, r, g, b));
```
GitToTheHub, line 133: You can remove the whitespace

GitToTheHub (changes requested):

framework/src/org/apache/cordova/SystemBarPlugin.java
gabriel2mm: Updated as suggested.

framework/src/org/apache/cordova/SystemBarPlugin.java
gabriel2mm: Updated as suggested.

framework/src/org/apache/cordova/SystemBarPlugin.java
gabriel2mm: Updated as suggested.

framework/src/org/apache/cordova/SystemBarPlugin.java
gabriel2mm: Updated as suggested.

framework/src/org/apache/cordova/SystemBarPlugin.java
gabriel2mm: Updated as suggested.

framework/src/org/apache/cordova/SystemBarPlugin.java
gabriel2mm: Updated as suggested.

GitToTheHub:
Hi, I have to verify some things, before I can approve this PR and merge. Thanks for the changes.

GitToTheHub (approved):
LGTM

PR #1950, opened 2026-06-04
fix(android): add missing imports in SystemWebChromeClient

denusklo:
### Platforms affected
Android

### Motivation and Context
`cordova-android` 12.0.1 introduced permission-handling code in `SystemWebChromeClient.java` that uses `List`, `ArrayList`, and `PackageManager`, but the corresponding import statements were not included. This causes a compilation failure for any project using `cordova-android@12.0.1`.

### Description
Added 3 missing import statements to `SystemWebChromeClient.java`:
```java
import java.util.ArrayList;
import java.util.List;
import android.content.pm.PackageManager;
```

### Testing
Built an Ionic 3 Android release with `cordova-android@12.0.1` — build previously failed with `cannot find symbol` errors for `List`, `ArrayList`, and `PackageManager`. After applying this fix, the build completes successfully.

### Checklist
- [x] Commit is prefixed with `(platform)` if this change only applies to one platform
- [ ] If this Pull Request resolves an issue, I linked to the issue in the text above

### DIFF
--- a/framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
+++ b/framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
@@ -19,7 +19,10 @@
 package org.apache.cordova.engine;
 
 import java.util.Arrays;
+import java.util.ArrayList;
+import java.util.List;
 import android.annotation.TargetApi;
+import android.content.pm.PackageManager;
 import android.app.Activity;
 import android.content.Context;
 import android.content.ActivityNotFoundException;


breautek:
v12 is completely EOL so we won't be making any patches or releases -- for that reason alone I'm going to close this PR. Google Play requires targeting relatively recent SDK targets that these older platform versions isn't tested or usually does not support without breaking changes. We would advise you to upgrade as soon as it's feasible.

But also, a simple search in the file shows that none of those imports are actually found or used in 12.x version, although they are used (with the necessary imports) in master. So I don't understand how `SystemWebChromeClient` could be causing import failures on those classes if they don't use them.

I see (all 3) of the added imports are in fact used in current master and they have the appropriate imports already.

If you don't distribute to Google Play or are otherwise not affected by Google's terms then you may maintain a copy of v12 with your own patches within the terms of the Apache license. Cordova CLI supports installing platforms directly from github: `cordova platform add https://github.com/denusklo/cordova-android.git#12.0.x` for example would install this PR branch.

PR #1951, opened 2026-06-07
refactor: address warnings in SystemWebChromeClient.java

jcesarmobile:
Android Studio was presenting several warnings about:
- variables that could be final
- variables that could be local
- redundant local variables
- use lambdas instead of anonymous functions
- missing descriptions
- unused variables

and a few others

I addressed them all

### DIFF
--- a/framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
+++ b/framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
@@ -73,21 +73,16 @@ public class SystemWebChromeClient extends WebChromeClient {
     private PermissionListener permissionListener;
     private static final int FILECHOOSER_RESULTCODE = 5173;
     private static final String LOG_TAG = "SystemWebChromeClient";
-    private long MAX_QUOTA = 100 * 1024 * 1024;
     protected final SystemWebViewEngine parentEngine;
 
     // the video progress view
     private View mVideoProgressView;
 
-    private CordovaDialogsHelper dialogsHelper;
-    private Context appContext;
-
-    private WebChromeClient.CustomViewCallback mCustomViewCallback;
-    private View mCustomView;
+    private final CordovaDialogsHelper dialogsHelper;
 
     public SystemWebChromeClient(SystemWebViewEngine parentEngine) {
         this.parentEngine = parentEngine;
-        appContext = parentEngine.webView.getContext();
+        Context appContext = parentEngine.webView.getContext();
         dialogsHelper = new CordovaDialogsHelper(appContext);
         permissionLauncher = parentEngine.cordova.getActivity().registerForActivityResult(new ActivityResultContracts.RequestMultiplePermissions(), isGranted -> {
             if (permissionListener != null) {
@@ -105,13 +100,11 @@ public class SystemWebChromeClient extends WebChromeClient {
      */
     @Override
     public boolean onJsAlert(WebView view, String url, String message, final JsResult result) {
-        dialogsHelper.showAlert(message, new CordovaDialogsHelper.Result() {
-            @Override public void gotResult(boolean success, String value) {
-                if (success) {
-                    result.confirm();
-                } else {
-                    result.cancel();
-                }
+        dialogsHelper.showAlert(message, (success, value) -> {
+            if (success) {
+                result.confirm();
+            } else {
+                result.cancel();
             }
         });
         return true;
@@ -122,14 +115,11 @@ public class SystemWebChromeClient extends WebChromeClient {
      */
     @Override
     public boolean onJsConfirm(WebView view, String url, String message, final JsResult result) {
-        dialogsHelper.showConfirm(message, new CordovaDialogsHelper.Result() {
-            @Override
-            public void gotResult(boolean success, String value) {
-                if (success) {
-                    result.confirm();
-                } else {
-                    result.cancel();
-                }
+        dialogsHelper.showConfirm(message, (success, value) -> {
+            if (success) {
+                result.confirm();
+            } else {
+                result.cancel();
             }
         });
         return true;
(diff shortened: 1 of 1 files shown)


framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
```diff
@@ -73,21 +73,16 @@ private interface PermissionListener {
...
-
-    private WebChromeClient.CustomViewCallback mCustomViewCallback;
-    private View mCustomView;
+    private final CordovaDialogsHelper dialogsHelper;
 
     public SystemWebChromeClient(SystemWebViewEngine parentEngine) {
         this.parentEngine = parentEngine;
-        appContext = parentEngine.webView.getContext();
+        Context appContext = parentEngine.webView.getContext();
         dialogsHelper = new CordovaDialogsHelper(appContext);
```
GitToTheHub: I think the temporary variable `appContect` is not needed here and `new CordovaDialogsHelper` could be directly called with `parentEngine.webView.getContext()`:

```java
dialogsHelper = new CordovaDialogsHelper(parentEngine.webView.getContext());
```

framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
```diff
@@ -173,6 +160,7 @@ public void onExceededDatabaseQuota(String url, String databaseIdentifier, long
             long totalUsedQuota, WebStorage.QuotaUpdater quotaUpdater)
     {
         LOG.d(LOG_TAG, "onExceededDatabaseQuota estimatedSize: %d  currentQuota: %d  totalUsedQuota: %d", estimatedSize, currentQuota, totalUsedQuota);
+        long MAX_QUOTA = 100 * 1024 * 1024;
         quotaUpdater.updateQuota(MAX_QUOTA);
```
GitToTheHub, line 163: I think the variable assignment `MAX_QUOTA` is redundant here. I would suggest this code:

```java
quotaUpdater.updateQuota(100 * 1024 * 1024);
```

GitToTheHub (changes requested):

framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
breautek: I disagree -- having a named variable can help describe intent since it avoids [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)).

So having `MAX_QUOTA` here is more preferable in my opinion.

GitToTheHub: Honestly MAX_QUOTA does not explain using 100 MB here. Why is 100 MB the max quota? What will happen if this function is called again, after 100 MB was set? I think a comment would better suit here an explain the reason.

breautek: > Why is 100 MB the max quota?

That part is obviously subjective.

> What will happen if this function is called again, after 100 MB was set?

This on the on exceed hook, which means the user has already exceeded the previously set quota, whatever what that might have been. I think the default is ~5 megabytes, but that's a webview implementation detail that's not documented.

If the user exceeds the quota and the quota is not expandable then they'll hit the standard [QuotaExceededError](https://developer.mozilla.org/en-US/docs/Web/API/QuotaExceededError).

Also worth noting that this whole system is deprecated and apparently no longer used since API 19 (according to https://developer.android.com/reference/android/webkit/WebChromeClient#onExceededDatabaseQuota(java.lang.String,%20java.lang.String,%20long,%20long,%20long,%20android.webkit.WebStorage.QuotaUpdater)) and the HTML5 quota system is used instead. So I probably wouldn't fret over this -- perhaps we can remove the hook and leave the implementation as default in our next major.

breautek (approved):

framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
GitToTheHub: > Also worth noting that this whole system is deprecated and apparently no longer used since API 19 (according to https://developer.android.com/reference/android/webkit/WebChromeClient#onExceededDatabaseQuota(java.lang.String,%20java.lang.String,%20long,%20long,%20long,%20android.webkit.WebStorage.QuotaUpdater)) and the HTML5 quota system is used instead. So I probably wouldn't fret over this -- perhaps we can remove the hook and leave the implementation as default in our next major.

Good point. So this code is not used anymore. This could be documented so it's not forget?

framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
GitToTheHub, line 158: Since we found out, that this method is not used anymore since API 19 (Android 4.4) and the HTML5 quota system is used instead, this could be added to the documentation of the method.

Link to Android documentation of this method: https://developer.android.com/reference/android/webkit/WebChromeClient#onExceededDatabaseQuota(java.lang.String,%20java.lang.String,%20long,%20long,%20long,%20android.webkit.WebStorage.QuotaUpdater)

Example:

```javadoc
/**
 * Handle database quota exceeded notification.
 * This method is no longer called since API 19 (Androd 4.4).
 * The HTML5 quota system is used instead.
 * The method can be removed in a future release.
 * @see https://developer.android.com/reference/android/webkit/WebChromeClient#onExceededDatabaseQuota(java.lang.String,%20java.lang.String,%20long,%20long,%20long,%20android.webkit.WebStorage.QuotaUpdater)
 */
```

GitToTheHub (changes requested):

framework/src/org/apache/cordova/engine/SystemWebChromeClient.java
jcesarmobile: I would prefer to remove it, in a separate PR

GitToTheHub: ok

GitToTheHub (approved):

erisu (approved):
LGTM

PR #1955, opened 2026-06-11
fix(statusbar): Properly handle CSS colours with alpha

dpogue:
### Platforms affected
Android

### Motivation and Context

`window.statusbar.setBackgroundColor()` is intended to take a CSS colour, get its computed RGB(A) value, and send that over to the platform code. However, the code here was assuming that we were dealing with an Android-style `#AARRGGBB` hex colour (which is wrong).

### Description

This matches the implementation from https://github.com/apache/cordova-ios/pull/1661 and turns the RGB ints and A float into an Android Color before converting it back into a `AARRGGBB` integer (because apparently none of the Android View APIs actually support Color in the places where you set colour values).

### Testing

I was planning to write E2E tests for this, but it seems that we don't actually run our Android tests in CI...

For now it's untested

### Checklist

### DIFF
--- a/cordova-js-src/plugin/android/statusbar.js
+++ b/cordova-js-src/plugin/android/statusbar.js
@@ -60,28 +60,29 @@ Object.defineProperty(statusBar, 'setBackgroundColor', {
         statusBarScript.style.color = value;
         var rgbStr = window.getComputedStyle(statusBarScript).getPropertyValue('color');
 
-        if (!rgbStr.match(/^rgb/)) { return; }
-
-        var rgbVals = rgbStr.match(/\d+/g).map(function (v) { return parseInt(v, 10); });
-
-        if (rgbVals.length < 3) {
+        if (!rgbStr.match(/^rgb/)) {
             return;
-        } else if (rgbVals.length === 3) {
-            rgbVals = [255].concat(rgbVals);
         }
 
-        // TODO: Use `padStart(2, '0')` once SDK 24 is dropped.
-        const padRgb = (val) => {
-            const hex = val.toString(16);
-            return hex.length === 1 ? '0' + hex : hex;
-        };
-        const a = padRgb(rgbVals[0]);
-        const r = padRgb(rgbVals[1]);
-        const g = padRgb(rgbVals[2]);
-        const b = padRgb(rgbVals[3]);
-        const hexStr = '#' + a + r + g + b;
+        var rgbVals = rgbStr.match(/[\d.]+/g).map(function (v, i) { return (i < 3) ? parseInt(v, 10) : parseFloat(v); });
+        if (rgbVals.length < 3) {
+            return;
+        }
 
         if (window.StatusBar) {
+            // try to let the StatusBar plugin handle it
+            // TODO: Use `padStart(2, '0')` once SDK 24 is dropped.
+            const padRgb = (val) => {
+                const hex = val.toString(16);
+                return hex.length === 1 ? '0' + hex : hex;
+            };
+
+            const r = padRgb(rgbVals[0]);
+            const g = padRgb(rgbVals[1]);
+            const b = padRgb(rgbVals[2]);
+            const a = padRgb(255 * (rgbVals[3] !== undefined ? rgbVals[3] : 1.0));
+
+            const hexStr = '#' + a + r + g + b;
             window.StatusBar.backgroundColorByHexString(hexStr);
         } else {
             exec(null, null, 'SystemBarPlugin', 'setStatusBarBackgroundColor', rgbVals);
--- a/framework/src/org/apache/cordova/SystemBarPlugin.java
+++ b/framework/src/org/apache/cordova/SystemBarPlugin.java
@@ -121,19 +121,17 @@ public class SystemBarPlugin extends CordovaPlugin {
      * If the supplied ARGB is invalid or fails to parse, it will silently ignore
      * the change request.
      *
-     * @param argbVals {A, R, G, B}
+     * @param argbVals {R, G, B, A}
      */
     private void setStatusBarBackgroundColor(JSONArray argbVals) {
         try {
-            int a = argbVals.getInt(0);
-            int r = argbVals.getInt(1);
-            int g = argbVals.getInt(2);
-            int b = argbVals.getInt(3);
-            overrideStatusBarBackgroundColor = parseColorFromString(String.format("#%02X%02X%02X%02X", a, r, g, b));
+            int r = argbVals.getInt(0);
+            int g = argbVals.getInt(1);
+            int b = argbVals.getInt(2);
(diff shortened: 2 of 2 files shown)


framework/src/org/apache/cordova/SystemBarPlugin.java
```diff
@@ -121,19 +121,17 @@ private void setStatusBarVisible(final boolean visible) {
...
-
-            if (overrideStatusBarBackgroundColor != null) {
-                updateStatusBar(overrideStatusBarBackgroundColor);
-            }
+            float r = argbVals.getInt(0) / 255.f;
+            float g = argbVals.getInt(1) / 255.f;
+            float b = argbVals.getInt(2) / 255.f;
+            float a = (float)argbVals.optDouble(3, 1.0);
+
+            overrideStatusBarBackgroundColor = Color.valueOf(r, g, b, a).toArgb();
```
GitToTheHub: Sadly [Color.valueOf](https://developer.android.com/reference/android/graphics/Color#valueOf(float,%20float,%20float,%20float)) only exists since API 26 (Android 8) and cannot be used on Android 7. So I think we have to stay on using [Color.parseColor](https://developer.android.com/reference/android/graphics/Color#parseColor(java.lang.String))

GitToTheHub (changes requested):

framework/src/org/apache/cordova/SystemBarPlugin.java
dpogue: oh, good catch!

framework/src/org/apache/cordova/SystemBarPlugin.java
```diff
@@ -121,19 +121,17 @@ private void setStatusBarVisible(final boolean visible) {
      * If the supplied ARGB is invalid or fails to parse, it will silently ignore
      * the change request.
      *
-     * @param argbVals {A, R, G, B}
+     * @param argbVals {R, G, B, A}
      */
     private void setStatusBarBackgroundColor(JSONArray argbVals) {
         try {
```
GitToTheHub, line 127: The try-catch can be removed and is not needed. A `JSONException` will not be thrown here.

cordova-js-src/plugin/android/statusbar.js
```diff
@@ -60,25 +60,24 @@ Object.defineProperty(statusBar, 'setBackgroundColor', {
         statusBarScript.style.color = value;
         var rgbStr = window.getComputedStyle(statusBarScript).getPropertyValue('color');
 
-        if (!rgbStr.match(/^rgb/)) { return; }
-
-        var rgbVals = rgbStr.match(/\d+/g).map(function (v) { return parseInt(v, 10); });
+        if (!rgbStr.match(/^rgb/)) {
+            return;
+        }
```
GitToTheHub, line 65: Could this be written as a one-liner?

```js
if (!rgbStr.match(/^rgb/)) return;
```

Or should this always be encapsulated?

dpogue:
I've updated this to use `Color.argb` which is available from API level 1

cordova-js-src/plugin/android/statusbar.js
```diff
@@ -60,25 +60,24 @@ Object.defineProperty(statusBar, 'setBackgroundColor', {
...
             return hex.length === 1 ? '0' + hex : hex;
         };
-        const a = padRgb(rgbVals[0]);
-        const r = padRgb(rgbVals[1]);
-        const g = padRgb(rgbVals[2]);
-        const b = padRgb(rgbVals[3]);
+        const r = padRgb(rgbVals[0]);
+        const g = padRgb(rgbVals[1]);
+        const b = padRgb(rgbVals[2]);
+        const a = padRgb(255 * (rgbVals[3] !== undefined ? rgbVals[3] : 1.0));
```
GitToTheHub: Since this code is only used for `if (window.StatusBar)` this could be moved in the code block there

GitToTheHub (changes requested):

framework/src/org/apache/cordova/SystemBarPlugin.java
dpogue: `argbVals.getInt(1)` could throw `JSONException` if this ends up getting called with bad data. This is technically callable by anyone from the webview, so we can't guarantee we'll only receive the expected data and we need to handle that safely without crashing.

cordova-js-src/plugin/android/statusbar.js
dpogue: I find the one-line `if` and `return` much harder and slower to read at a glance because it doesn't visually look like a conditional block due to the lack of indenting

framework/src/org/apache/cordova/SystemBarPlugin.java
GitToTheHub: Oh you are right, sorry. I forgot that this can throw a `JSONException`. Good point :)

cordova-js-src/plugin/android/statusbar.js
GitToTheHub: Ok than it stays like it is

cordova-js-src/plugin/android/statusbar.js
```diff
@@ -60,28 +60,28 @@ Object.defineProperty(statusBar, 'setBackgroundColor', {
...
-            const hex = val.toString(16);
-            return hex.length === 1 ? '0' + hex : hex;
-        };
-        const a = padRgb(rgbVals[0]);
-        const r = padRgb(rgbVals[1]);
-        const g = padRgb(rgbVals[2]);
-        const b = padRgb(rgbVals[3]);
-        const hexStr = '#' + a + r + g + b;
-
         if (window.StatusBar) {
```
GitToTheHub, line 72: Just a thought. This line is only executed if the cordova-plugin-statusbar is used or? Because I didn't find a `window.StatusBar.backgroundColorByHexString` method online. This seems plugin related. Maybe this can be added as a comment

```jsdoc
// cordova-plugin-statusbar used
```

GitToTheHub:
Looks good. The only thing is, that the status bar don't overlays the web view as on iOS. So alpha values don't impact the web view content.

GitToTheHub:
Is there a reason why this is still a draft? When we release cordova-android, should we sync it with a release of cordova-ios, so on both platform the format `#RRGGBBAA` can be used? Otherwise the users have to mix the format.

dpogue:
> Is there a reason why this is still a draft?

I had been hoping to write test cases, and then left this as a draft so it could be merged at the same time as iOS, but it should be good to go implementation-wise

GitToTheHub:
Ok. If the iOS PR is good to go, this could be merged together

GitToTheHub (approved):

erisu (approved):

Issue #1946, opened 2026-05-27
Visual glitch occurs during keyboard (IME) animations


erisu:
# Bug Report

## Problem

A brief visual glitch occurs during keyboard (IME) animations where the app background becomes visible temporarily.

This appears to have been introduced in #1924, which was resolving an issue where the WebView was not fully scrollable when the IME keyboard was open.

Based on initial investigation, the behavior may be related to the WebView being resized while the IME is still animating, which can temporarily expose the underlying window background.

The issue is more noticeable when there is a strong contrast between the app's background and the web content's (for example, a light web page background over a dark application background defined via configuration preferences).

### What is expected to happen?

During the keyboard animations, the WebView should remain fully visible.

### What does actually happen?

During the keyboard animations, the app's background is briefly visible, which appears as a visual glitch.

## Information

- Create a long page with multiple input elements (so the page is scrollable).
  - This step may not be necessary to reproduce the visual glitch, but it is important when addressing the issue as we need to continue to ensure the WebView remains scrollable and can reach the bottom of the page (rather than being hidden by the IME).
- Set the web page's background to a light color (e.g. white).
- Set the application background color to a dark color using `config.xml`:

```xml
<preference name="BackgroundColor" value="#000000" />
```

- Build and run the application on an Android device.
- Focus on an input field to trigger opening of the keyboard.

### Environment, Platform, Device

I used the following emulator:

- Android SDK 24

### Version information

```
Cordova Packages:

    cli: 13.0.0
        common: 6.0.0
        create: 6.0.0
        lib: 13.0.0
            common: 6.0.0
            fetch: 5.0.0

Project Installed Platforms:

    android: 15.0.1-dev.0

Project Installed Plugins:


Environment:

    OS: Tahoe 26.4.1 (25E253) (darwin 25.4.0) arm64
    Node: v24.13.0
    npm: 11.6.2
```

> [!NOTE]
> `android: 15.0.1-dev.0` is current master branch.

The app was launched from **Android Studio Panda 4 | 2025.3.4**.

## Checklist

- [x] I searched for existing GitHub issues
- [x] I updated all Cordova tooling to most recent version
- [x] I included all the necessary information above

gabriel2mm:
The behavior is similar to what already existed in Cordova 14. However, in the previous version, the background shown during keyboard animations matched the device navigation bar color. In this version, we noticed a change due to the edge-to-edge behavior introduced by Android.

[...]

PR #1983, opened 2026-07-27
feat(SystemBarPlugin): add setNavigationBarBackgroundColor to change the navigation bar

ollm:
### Platforms affected

Android

### Motivation and Context

Since status bar handling has been added directly to cordova, navigation bar control should also be included.

### Description

The changes add `window.statusbar.setNavigationBarBackgroundColor('#ffffff00')` to change the navigation bar color on older devices or when gesture navigation is disabled (3-button navigation).

The navigation buttons color is detected automatically, just like with `setBackgroundColor`. `transparentNavigationBar` is also applied automatically if the provided color is fully transparent.

The color is also applied as the background color of the view. This should mitigate/resolve the following issue https://github.com/apache/cordova-android/issues/1946

Example usage:

``` js
window.statusbar.setNavigationBarBackgroundColor('#1f2933');
```

Android 11:

Captura desde 2026-07-27 11-59-16

Android 17:

Captura desde 2026-07-27 11-59-06

``` js
window.statusbar.setNavigationBarBackgroundColor('#fafafa')
```

Android 11:

Captura desde 2026-07-27 11-59-23

Android 17:

Captura desde 2026-07-27 11-58-54

### Testing

I tested it on the Android 11 and Android 17 emulators.

### Note

 I'm not sure if this is intentional, but I wanted to mention it. If `#00000000` is passed to `setBackgroundColor`, the following code treats it as fully transparent, so the navigation bar color is not changed to the expected one (White icons). For now, I'm using `#01010100` as a workaround.

https://github.com/apache/cordova-android/blob/ed4e629e9da5721fdaa10dd6da5c85049723686e/framework/src/org/apache/cordova/SystemBarPlugin.java#L189-L193

### Checklist

### DIFF
--- a/cordova-js-src/platform.js
+++ b/cordova-js-src/platform.js
@@ -41,6 +41,10 @@ module.exports = {
         // see the file under plugin/android/statusbar.js
         modulemapper.clobbers('cordova/plugin/android/statusbar', 'window.statusbar');
 
+        // Attach the internal navigationBar utility to window.navigationbar
+        // see the file under plugin/android/navigationbar.js
+        modulemapper.clobbers('cordova/plugin/android/navigationbar', 'window.navigationbar');
+
         var APP_PLUGIN_NAME = Number(cordova.platformVersion.split('.')[0]) >= 4 ? 'CoreAndroid' : 'App';
 
         // Inject a listener for the backbutton on the document.
--- a/cordova-js-src/plugin/android/navigationbar.js
+++ b/cordova-js-src/plugin/android/navigationbar.js
@@ -1,0 +1,53 @@
+/*
+    Licensed to the Apache Software Foundation (ASF) under one
+    or more contributor license agreements.  See the NOTICE file
+    distributed with this work for additional information
+    regarding copyright ownership.  The ASF licenses this file
+    to you under the Apache License, Version 2.0 (the
+    "License"); you may not use this file except in compliance
+    with the License.  You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing,
+    software distributed under the License is distributed on an
+    "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+    KIND, either express or implied.  See the License for the
+    specific language governing permissions and limitations
+    under the License.
+*/
+
+var exec = require('cordova/exec');
+
+var navigationBar = {};
+
+// This <script> element is explicitly used by Cordova's navigationbar for computing color. (Do not use this element)
+const navigationBarScript = document.createElement('script');
+document.head.appendChild(navigationBarScript);
+
+/**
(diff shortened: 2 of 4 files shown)


GitToTheHub:
@erisu What do you think? I think it's a good idea to let the user also control the appearance of the navigation bar. I think it should be used over its own API. So it should exists a `NavigationBarPlugin`.

GitToTheHub:
I saw the native code is not in a `StatusBarPlugin`, it's in a `SystemBarPlugin`. So the navigation bar code could be handled there also. Maybe `statusbar.js` could be changed to `systembar.js`, also. The call for changing the navigation bar could be done over `windows.navigationbar`?

ollm:
> I saw the native code is not in a `StatusBarPlugin`, it's in a `SystemBarPlugin`. So the navigation bar code could be handled there also. Maybe `statusbar.js` could be changed to `systembar.js`, also. The call for changing the navigation bar could be done over `windows.navigationbar`?

I think that would be the best approach, since I'm already reusing some functions from `SystemBarPlugin`.

I've also made the passed color be applied as the background color of the view https://github.com/ollm/cordova-android/commit/61e2295c39cc87a37bc6b29cf8071fec11b4d40c. This should mitigate/resolve the following issue https://github.com/apache/cordova-android/issues/1946

erisu:
> @erisu What do you think?

I'll try to find some time to test the current behavior and the changes in this PR as well.

> I saw the native code is not in a `StatusBarPlugin`, it's in a `SystemBarPlugin`. So the navigation bar code could be handled there also. Maybe `statusbar.js` could be changed to `systembar.js`, also. The call for changing the navigation bar could be done over `windows.navigationbar`?

My suggestion would be to avoid adding this to `statusbar.js` and to keep the file name as it is. Since the module is clobbered onto the `window.statusbar` object, it makes sense for it to contain status bar-related functionality only and the filename match its purpose. Additionally, the iOS implementation follows the same approach, so keeping this structure would also maintain consistency across platforms.

Instead, one possible approach would be to introduce a new `navigationbar.js` file and clobber it onto the `window.navigationbar` object. The API could then be something like `window.navigationbar.setBackgroundColor()`.
I also did a quick search and couldn't find any web standard that defines `window.navigationbar`, so it seems like it would be available for us to use.

I don't see any issue with both `statusbar.js` and `navigationbar.js` pointing to the same native-side class since the class is dedicated to all System Bars.

Again, I will try to find some time to look more in-depth.

GitToTheHub:
Ok sounds good. I think the solution with the navigationbar.js is a good one.

ollm:
If you'd like, I can already move the js code to a `navigationbar.js` file.

GitToTheHub:
I think you can already move the code to a navigationbar.js

ollm:
I've moved the code to `navigationbar.js` d263a689157357db385422f29035185cb98f4eef

GitToTheHub:
Looks good 🙂

PR #1990, opened 2026-08-05
chore(dependabot): update config

erisu:
### Motivation and Context

Update dependabot configs

### Description

- GH Action Interval changes from daily to weekly
- NPM Interval remains as weekly
- NPM PRs
  - Group patch releases
    - Increase cooldown from 4 to 5 days
  - Group minor releases
    - Increase cooldown from 4 to 5 days
  - Major releases will remain as single PR for each package
    - Increase cooldown from 4 to 14 days
- GH Actions PR
  - Group patch and minors
    - Increase cooldown from 4 to 5 days
  - Major releases will remain as single PR for each package
    - Increase cooldown from 4 to 14 days

Overall, the cooldown was increased to allow more time for packages to be vetted before PRs are being opened. By default, Dependabot has an open PR limit of five that would avoid spamming. Instead of increasing this limit, I believe grouping some updates is the better approach.

Major updates would still be opened as individual PRs so they can be reviewed more carefully. They also have a significantly longer cooldown (two weeks), since patch and minor updates are more likely to be our next targeted release.

One drawback of the 5 PR limit is that major update PRs could consume all available slots and prevent patch and minor update PRs from being opened. In this situation, I recommend closing major update PRs if a major release is not being targeted. We can manually revist and apply major updates when ready.

### Testing

n/a

### Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -16,17 +16,38 @@
 # under the License.
 
 version: 2
+
 updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     cooldown:
-      default-days: 4
+      default-days: 5
+      semver-major-days: 14
     schedule:
-      interval: "daily" # Every weekday, Monday to Friday
+      interval: "weekly" # Every Monday
+    groups:
+      gh-actions-patch-minor:
+        patterns:
+          - "*"
+        update-types:
+          - "patch"
+          - "minor"
 
   - package-ecosystem: "npm"
     directory: "/"
     cooldown:
-      default-days: 4
+      default-days: 5
+      semver-major-days: 14
     schedule:
-      interval: "weekly" # By default on a Monday
+      interval: "weekly" # Every Monday
+    groups:
+      npm-patch:
+        patterns:
+          - "*"
+        update-types:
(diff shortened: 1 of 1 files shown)


GitToTheHub:
Is it possible to merge patches automatically? I mean this should bring no problems if doing so.

erisu:
> Is it possible to merge patches automatically? I mean this should bring no problems if doing so.

I don't think there's a policy that prevents it. But, IMO, it would still be valuable to have human involvement in the process. We might be able to use the merge queue feature to automatically merge a PR after it has been reviewed and approved. But as I never used this feature before, I would need to look into it more. I don't know if it has granular configuration options to define which PRs it applies to.

GitToTheHub:
> But, IMO, it would still be valuable to have human involvement in the process.

I had just the thought to simplify the dependabot updates, but maybe it's better a human give them free.

> We might be able to use the merge queue feature to automatically merge a PR after it has been reviewed and approved.

I don't see a big benefit that the merge is done automatically after it is approved. If you review and approve it, the merge step doesn't take much effort.

GitToTheHub (approved):

PR #1991, opened 2026-08-07
fix(statusbar): restore `StatusBarOverlaysWebView` support in non-edge-to-edge mode

GitToTheHub:
### Motivation and Context

- When using the plugin `cordova-plugin-statusbar` with `cordova-android@14.0.1`, the status bar overlays the web view by default and a translucent color can be set. Also the setting can be controlled by the preference `StatusBarOverlaysWebView` in `config.xml`. This PR restores the behaviour. `StatusBarOverlaysWebView ` will be only applied, when the status bar plugin is added or the app is not in edge-to-edge mode

### Notes

[@dpogue mentioned](https://github.com/apache/cordova-android/pull/1991#issuecomment-5220804122), that this should be done by `<meta name="viewport" content="viewport-fit=cover|contain">` by a MutationObserver in JavaScript. On iOS this behaviour can already be obtained by using `<meta name="viewport" content="viewport-fit=cover“>`. @dpogue opened PR #2010 on cordova-android which would resolve this.

## Sample

Configuration:

- Add cordova-plugin-statusbar
- Set `<preference name="StatusBarBackgroundColor" value="#22ff0000" />` in `config.xml`
 - Test on Android 10 Emulator

| cordova-android 14.0.1 | cordova-android 15.1.0 |
| --- | --- |
|  |   |

Restored behaviour in this PR:

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -107,6 +107,7 @@ public class CordovaActivity extends AppCompatActivity {
 
     private boolean canEdgeToEdge = false;
     private boolean isFullScreen = false;
+    private boolean statusBarOverlaysWebView = false;
 
     /**
      * Called when the activity is first created.
@@ -122,6 +123,7 @@ public class CordovaActivity extends AppCompatActivity {
         loadConfig();
 
         canEdgeToEdge = preferences.getBoolean("AndroidEdgeToEdge", false);
+        statusBarOverlaysWebView = preferences.getBoolean("StatusBarOverlaysWebView", true);
 
         String logLevel = preferences.getString("loglevel", "ERROR");
         LOG.setLogLevel(logLevel);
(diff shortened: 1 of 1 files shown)


dpogue:
As previously mentioned, this should really be controlled by `<meta name="viewport" content="viewport-fit=contain">` or `<meta name="viewport" content="viewport-fit=cover">` in the HTML content rather than through config.xml, to match behaviour on iOS.

Currently, the Android WebView doesn't expose the viewport-fit value to us, so what we'd need to do is set up a MutationObserver in JavaScript to observe the value and toggle things in platform code using an internal API.

GitToTheHub:
I try to understand something about the status bar.

When I set `StatusBarOverlaysWebView` to `true` on iOS and set a translucent background color to the status bar e.g. `<preference name="StatusBarBackgroundColor" value="#22ff0000" />`, the status bar background color is not applied when plugin cordova-plugin-statusbar is installed. Is this a bug? The readme says:

> If StatusBarOverlaysWebView is set to true, then a 8 digit hex (#AARRGGBB) string can optionally be used to define the transparency.

This is the same behaviour without the plugin when `viewport-fit=cover` is set. When I remove `viewport-fit=cover` the status bar overlays the web view, the translucent background color is applied and the web content is visible behind the status bar.

GitToTheHub:
> As previously mentioned, this should really be controlled by <meta name="viewport" content="viewport-fit=contain"> or <meta name="viewport" content="viewport-fit=cover"> in the HTML content rather than through config.xml, to match behaviour on iOS.

I understand. But since we support for iOS the config.xml preference `StatusBarBackgroundColor` and the html meta `theme-color`, we could support `StatusBarOverlaysWebView` as a workaround for Android since `viewport-fit=cover` and `viewport-fit=cover` is not supported currently by Android. For iOS the plugin documentation of cordova-plugin-statusbar already states that [`viewport-fit=cover` is needed](https://github.com/apache/cordova-plugin-statusbar#ios-quirks) to let the status bar overlay the web view, so there is no code change needed for cordova-ios.

PR #2000, opened 2026-08-11
fix(statusbar): hide status bar correctly

### LINKED ISSUE #1999: Cordova Android 15 - Hiding the statusbar: difference between StatusBar plugin and integrated functionality
# Bug Report

According to the blog post of the Cordova Android 15.0.0 release, the StatusBar plugin has become redundant since its functionality is integrated in Cordova Android. However, I'm getting different results between using the StatusBar plugin or the integrated functionality.

## Problem

Scenario A:
- Add StatusBar plugin
- Inside the device ready event, call: `StatusBar.hide();`
- Result: the statusbar completely disappears and the webapp content fills the freed-up space. When the user slides down from the top, the status bar appears (over the content) with a translucent background and after a short while, disappears again.

Scenario B:
- Remove StatusBar plugin
- Inside the device ready event, call: `window.statusbar.visible = false;`
- Result: the background of the statusbar becomes transparent, but the content of the statusbar is still visible (clock, connectivity, battery and such). The webapp content fills the freed-up space, but now overlaps with the remaining statusbar content. When the user slides down from the top, nothing happens. There's no way to make the statusbar content disappear.

The result of scenario A is what I would expect and prefer to happen.

The result of scenario B.

## Information
Both scenarios have AndroidEdgeToEdge not set in the config.xml (defaults to false).

### Command
cordova run android

I've tested this on a Pixel 3a running Android 12 and a Pixel 8 running Android 16.
Btw, the latter has a camera notch and I was happy to see the content of the webapp staying within the safe zone.

### Version information
- Cordova CLI 13.0.0
- Cordova Android 15.0.0
- cordova-plugin-statusbar 4.0.0

## Checklist

GitToTheHub:
### Motivation and Context

Calling `window.statusbar.visible = false` sets the background of the statusbar transparent, but the content of the statusbar is still visible (clock, connectivity, battery and such). The webapp content fills the freed-up space, but now overlaps with the remaining statusbar content. When the user slides down from the top, nothing happens. There's no way to make the statusbar content disappear.

### Description

- Hide the real Android status bar through the built-in API and allow it to be temporarily revealed by swipe. Respect the system status-bar visibility when calculating Cordova's faux status-bar inset, preserving compatibility with cordova-plugin-statusbar.
- Fixes #1999

Generated-By: GPT-5.6 Terra

### Checklist

### DIFF
--- a/framework/src/org/apache/cordova/CordovaActivity.java
+++ b/framework/src/org/apache/cordova/CordovaActivity.java
@@ -223,7 +223,8 @@ public class CordovaActivity extends AppCompatActivity {
                     WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout()
             );
 
-            boolean isStatusBarVisible = statusBarView.getVisibility() != View.GONE;
+            boolean isStatusBarVisible = statusBarView.getVisibility() != View.GONE
+                    && insets.isVisible(WindowInsetsCompat.Type.statusBars());
             int top = isStatusBarVisible && !canEdgeToEdge && !isFullScreen ? bars.top : 0;
             int left = !canEdgeToEdge && !isFullScreen ? bars.left : 0;
             int right = !canEdgeToEdge && !isFullScreen ? bars.right : 0;
--- a/framework/src/org/apache/cordova/SystemBarPlugin.java
+++ b/framework/src/org/apache/cordova/SystemBarPlugin.java
@@ -34,6 +34,7 @@ import android.widget.FrameLayout;
 import androidx.core.content.ContextCompat;
 import androidx.core.view.ViewCompat;
 import androidx.core.view.WindowCompat;
+import androidx.core.view.WindowInsetsCompat;
 import androidx.core.view.WindowInsetsControllerCompat;
 
 import org.json.JSONArray;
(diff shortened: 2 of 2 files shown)


dpogue:
Confirm that this matches the behaviour on iOS, because I'm pretty sure it only hides the background there.

The intention for apps that want to do more complex things with the status bar (such as hiding it entirely) was to use either `Fullscreen` preference or to use the StatusBar plugin. The built-in status bar support was just meant to cover the most common simple cases of setting a background colour and having it push down the webview or not.

GitToTheHub:
> Confirm that this matches the behaviour on iOS, because I'm pretty sure it only hides the background there.

I checked this with the cordova-ios master, not using `cordova-plugin-statusbar` and calling `window.statusbar.visible = false`. The status bar is completely hidden:



It would be a little bit confusing calling `window.statusbar.visible = false` and it would only hide the status bar background. If this was meant to just hide the background it should be called `window.statusbar.backgroundVisible`.

Also calling `StatusBar.hide()` on `cordova-plugin-statusbar` by using cordova-android 15.1.0 will leave a gray space at the top:

GitToTheHub:
Since it's clear now that on iOS the system status bar is hidden when `window.statusbar.visible = false` is called, as it also now documented on [cordova-ios/cordova-js-src/plugin/ios/statusbar.js`](https://github.com/apache/cordova-ios/blob/04fccb9827a77d909cead366a11de5944431fbe2/cordova-js-src/plugin/ios/statusbar.js#L29-L32), this change here makes it equal to Android.

dpogue (approved):

Issue #1967, opened 2026-07-03
Option to completely disable the built-in window.statusbar plugin


ollm:
# Feature Request

## Motivation Behind Feature

I'm using my own status bar plugin because it provides additional functionality that is not available in the built-in implementation, such as proper `edge-to-edge` support and the ability to retrieve the status bar height.

Since the built-in status bar plugin was introduced in Cordova Android, I've been experiencing conflicts between the two implementations. For example, the status bar icon color changes unexpectedly a few moments after the app starts, most likely when the built-in plugin updates the status bar.

The simplest solution would be to provide an option to completely disable the built-in plugin so developers can use their own implementation without interference.

## Feature Description

Add a configuration option that completely disables the built-in `window.statusbar` plugin, preventing it from initializing or modifying the status bar.

This would allow custom plugins, such as my own implementation, to work without conflicts:
https://github.com/ollm/cordova-plugin-statusbar/tree/feature/addGetStatusBarHeightMethod

## Alternatives or Workarounds

The current workaround is to remain on Cordova Android `14.0.1`, where the built-in status bar plugin is not present.

breautek:
Notwithstanding the bug present 15.0.0 (you can test against github's master branch for the fix) -- using cordova-android@15 with `AndroidEdge2Edge` preference enabled is effectively what you're already asking I think.

While in this mode:
- The `window.statusbar` API is still present but it is a no-op.
- Cordova makes no status bar view -- there is simply the system bar.
- The webview is rendered full screen, and will overlay underneath system bars.
- Safe area insets can be used to pad or offset your content to clear system bars and other screen obstacles. It's my understanding this will require a fairly recent webview version installed on the device. Alternatively a plugin that reports the safe area insets sourced from the native views can be used, which is similar to what your status bar fork is already doing.

If you're still using cordova-plugin-statusbar -- the plugin will still initialize and control the status bar to some degree. The plugin still offers features that is not available in the "core" statusbar implementation.

> Notwithstanding the bug present 15.0.0

on 15.0.0 there is a bug where the webview is not positioned properly resulting in a white bar appearing in the status bar area, which is fixed but pending release at the time of writing. So for testing purposes, it's worth to install the platform directly from github.

Can you let me know where the gaps are?

ollm:
Thanks for the reply!

I tested the latest `master` branch, and the issue still persists (I have `AndroidEdgeToEdge` enabled).

[...]

PR #2017, opened 2026-09-14
fix(android): status bar icon appearance is inverted

MaximBelov:
### Platforms affected

Android

### Motivation and Context

`SystemBarPlugin.updateRootView` sets `APPEARANCE_LIGHT_STATUS_BARS` the wrong way round. The flag means "the status bar background is light, so draw dark icons", but the code sets it when the background is dark:

```java
if (isBackgroundColorLight) {
    controller.setSystemBarsAppearance(0, appearance);
} else {
    controller.setSystemBarsAppearance(appearance, appearance);
}
```

`updateStatusBar`, a few lines below, does the same job correctly through `WindowInsetsControllerCompat.setAppearanceLightStatusBars(...)` and runs after it — but the value written here is the one that sticks, because the two go through different controllers.

Since `updateRootView` takes its colour from the root view, falling back to the ui-mode colour, the practical result is that the status bar text follows the system's day/night setting rather than the colour of the status bar itself.

I ran into this with the status bar at `#007db5`. Same build, same bar colour, only the system theme changed, sampling the screenshot rather than judging by eye:

| system theme | clock drawn |
| ------------ | ----------- |
| day          | `#edf6fa`   |
| night        | `#002536`   |

The second one is dark ink on a dark bar. With a black status bar in night mode the clock and icons vanish altogether. An app cannot correct it from the JS side: calling `setStatusBarBackgroundColor` re-runs `updateStatusBar`, whose appearance write loses to this one.

This looks like the same thing reported in #1967.

### Description

Drop the platform-controller branch and use the compat controller for both bars, which is what the following line already does for the navigation bar. That fixes the inversion and removes the mixed use of two controllers writing the same flag. `android.view.WindowInsetsController` is no longer referenced, so the import goes with it.

### Testing

Built and installed on a physical device (Android 14, API 34) with the status bar colour set to `#007db5`, and compared both system themes. Before the change the clock was `#edf6fa` in day mode and `#002536` in night mode; after it, `#edf6fa` in both. The navigation bar appearance is unchanged, as that line was already using the compat controller.

I have not added automated coverage — this is view/window state that the existing test setup does not reach.

### Checklist


MaximBelov:
Closing this — I measured it properly and it turns out to be a net loss.

I ran the same build with and without this change on two emulators, status bar at `#b5c426` (a light colour), two runs each, reading the icon colour out of a screenshot rather than judging by eye:

| | without the change | with it |
| --- | --- | --- |
| API 36, day | dark icons ✔ | dark icons ✔ |
| API 36, night | dark icons ✔ | dark icons ✔ |
| API 34, day | white icons ✘ | white icons ✘ |
| API 34, night | dark icons ✔ | white icons ✘ |

API 36 is already correct without it, so the change buys nothing there, and on API 34 it takes the one mode that was right and breaks it.

The reason is the part I had wrong. Below API 35 the `window.getInsetsController()` write in `updateRootView` is the one that actually reaches the window; the `WindowInsetsControllerCompat` write does not, at least not once `cordova-plugin-statusbar` has been through its `setSystemUiVisibility` path at init. Removing the platform write leaves the appearance at whatever that plugin set — `StatusBarStyle` defaults to `lightcontent`, so white, permanently. Confirmed directly on a physical Android 14: painting the bar `#b5c426` through `SystemBarPlugin` moved the bar from luminance 94 to 174 and left the ink at 243, and `StatusBar.styleDefault()` afterwards moved nothing at all.

The inversion itself is real — on API 34 the icon colour tracks the system day/night setting rather than the bar it sits on, so it is wrong in exactly one of the two modes whatever colour the bar is. But un-inverting the branch in place is not the fix either: `updateRootView` derives its value from the root view colour, not the status bar's, so where the two differ it only swaps which mode is wrong. What it needs is for `updateStatusBar` — which owns the bar and knows its actual colour — to be the one that writes last, through whichever controller works on that API level.

Sorry for the noise.

